PR #6, opened 2018-02-14
Fix e-mail address in docs

ttheodor:


ftynse:
@ttheodor I don't think this will be replicated to the docs webpage automatically.  One also needs to update the gh-pages branch with the generated content.

Issue #16, opened 2018-02-14
Comparison operators


abadams:
The front-end doesn't understand ==, <, >, !=, etc. Hilariously, we *do* have a ternary operator, but I'm not sure what you would use for the first argument :)

Tagging @zdevito to add them to the lexer/parser

abadams:
I got lost somewhere inside semantic checking and type inference with this. I think this will be trivial for Zach, so I'll leave it to him.

prigoyal:
done in #28

PR #17, opened 2018-02-14
Add test case for concat operation

prigoyal:
for repro, we should support comparison operators

### DIFF
--- a/test/test_execution_engine.cc
+++ b/test/test_execution_engine.cc
@@ -45,6 +45,24 @@ struct ATenCompilationUnitTest : public ::testing::Test {
   }
 };
 
+TEST_F(ATenCompilationUnitTest, DISABLED_Concat) {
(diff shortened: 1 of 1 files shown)


abadams:
See #16

prigoyal:
oops, I pushed again to disable the test, need stamp again @abadams

prigoyal:
merging, CI is botched due to resources allocation

test/test_execution_engine.cc
```diff
@@ -45,6 +45,24 @@ struct ATenCompilationUnitTest : public ::testing::Test {
   }
 };
 
+TEST_F(ATenCompilationUnitTest, DISABLED_Concat) {
```
nicolasvasilache: This is dead code, I see this has been activated in master, still in the future for such cases, please make a test that does something and:
1.  expect the throw / failure that is expected
2. open a proper issue
3. tag the issue in the test as TODO(#number)

Thanks!

PR #19, opened 2018-02-15
Update docs and write about temporaties explicitly

prigoyal:
we treat everything as input or output in TC. Since TC doesn't do allocation, make it explicit with an example in the doc.

also small cosmetics fixes on the installation doc and added my name so people can contact easily about build issues on slack, github. if this doesn't feel right, let me know and I can revert.

can't tag Albert Cohen for review for some reason.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -1,6 +1,8 @@
 Installation Guide
 ==================
 
+**Author**: `Priya Goyal <https://github.com/prigoyal>`_
+
 The following instructions are provided for developers who would like to
 experiment with the library.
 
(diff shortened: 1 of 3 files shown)


prigoyal:
also testing CI via this PR (checking if the resource allocation is correct now)

docs/source/installation.rst
```diff
@@ -1,6 +1,8 @@
 Installation Guide
 ==================
 
+**Author**: `Priya Goyal <https://github.com/prigoyal>`_
```
prigoyal: I added this for support purposes, when people read docs, they can contact on github/slack immediately.

prigoyal:
thanks @ftynse I'll make sure to do that separately next time onwards :)

PR #25, opened 2018-02-15
I guess this is x(j) instead of b(j)?

vmiheer:


prigoyal:
Hi @vmiheer, thanks for the PR. Yes, that is correct. Can you take a look at the facebook-github-bot message above once before we can take a look at the PR further? Thanks for contributing.

ftynse:
Fixed in #31

Thanks for spotting this!

PR #30, opened 2018-02-16
Permit parallel CPU backend compilation (though currently disabled)

wsmoses:
Parallel compilation on CPU now available, though needs information from ISL to enable

### DIFF
--- a/src/core/polyhedral/codegen_llvm.cc
+++ b/src/core/polyhedral/codegen_llvm.cc
@@ -21,6 +21,7 @@
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Analysis/TargetTransformInfo.h"
+#include "llvm/Config/llvm-config.h"
 #include "llvm/ExecutionEngine/ExecutionEngine.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/LegacyPassManager.h"
@@ -41,6 +42,10 @@
 #include "tc/core/polyhedral/schedule_isl_conversion.h"
 #include "tc/core/polyhedral/scop.h"
 #include "tc/core/scope_guard.h"
+
+#ifndef LLVM_VERSION_MAJOR
+#error LLVM_VERSION_MAJOR not set
+#endif
 
 using namespace Halide;
 
@@ -303,8 +308,8 @@ public:
     Halide::Internal::debug(3) << "Optimizing module\n";
 
     if (Halide::Internal::debug::debug_level() >= 3) {
-        #if LLVM_VERSION >= 50
-        module->print(dbgs(), nullptr, false, true);
+        #if LLVM_VERSION_MAJOR >= 5
+        module->print(llvm::dbgs(), nullptr, false, true);
         #else
         module->dump();
         #endif
@@ -338,8 +343,8 @@ public:
 
     llvm::PassManagerBuilder b;
     b.OptLevel = 3;
-    b.tapirTarget = new llvm::tapir::CilkABI();
-#if LLVM_VERSION >= 50
+    b.tapirTarget = new llvm::CilkABI();
+#if LLVM_VERSION_MAJOR >= 5
     b.Inliner = llvm::createFunctionInliningPass(b.OptLevel, 0, false);
 #else
     b.Inliner = llvm::createFunctionInliningPass(b.OptLevel, 0);
@@ -347,7 +352,7 @@ public:
     b.LoopVectorize = true;
     b.SLPVectorize = true;
 
-#if LLVM_VERSION >= 50
+#if LLVM_VERSION_MAJOR >= 5
     if (TM) {
         TM->adjustPassManager(b);
     }
@@ -366,8 +371,8 @@ public:
 
     Halide::Internal::debug(3) << "After LLVM optimizations:\n";
     if (Halide::Internal::debug::debug_level() >= 2) {
-        #if LLVM_VERSION >= 50
-        module->print(dbgs(), nullptr, false, true);
+        #if LLVM_VERSION_MAJOR >= 5
+        module->print(llvm::dbgs(), nullptr, false, true);
         #else
         module->dump();
         #endif
@@ -530,7 +535,8 @@ class LLVMCodegen {
         llvm::BasicBlock::Create(llvmCtx, "loop_latch", function);
     auto* loopExitBB = llvm::BasicBlock::Create(llvmCtx, "loop_exit", function);
 
-    bool parallel = true;
+    //TODO: integrate query ISL as to whether the relevant loop ought be parallelized
+    bool parallel = false;
 
     llvm::Value* SyncRegion = nullptr;
     if (parallel) {
--- a/src/core/polyhedral/codegen_llvm.cc
+++ b/src/core/polyhedral/codegen_llvm.cc
@@ -28,8 +28,8 @@
 #include "llvm/IR/Verifier.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Support/raw_ostream.h"
-#include "llvm/Transforms/IPO/PassManagerBuilder.h"
 #include "llvm/Transforms/IPO.h"
+#include "llvm/Transforms/IPO/PassManagerBuilder.h"
 #include "llvm/Transforms/Tapir/CilkABI.h"
 
 #include "Halide/Halide.h"
@@ -205,6 +205,10 @@ class IslAstExprInterpeter {
   }
 };
 
+DEFINE_bool(llvm_dump_before_opt, false, "Print IR before optimization");
+DEFINE_bool(llvm_dump_after_opt, false, "Print IR after optimization");
+static constexpr int kOptLevel = 3;
+
 class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
  public:
   const isl::pw_multi_aff* iteratorMap_;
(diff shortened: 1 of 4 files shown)


wsmoses:
This requires bumping halide and also the Tapir in the docker file as mentioned to @prigoyal

test/test_mapper_llvm.cc
```diff
@@ -68,7 +68,7 @@ TEST(LLVMCodegen, MultiStmt) {
  {
    O1(i, j) +=! A(i, j, rk, rl) * B(i, j)
    O2(i, j) = C(i, j) * D(i, j)
-   O3(i, j) = O1(i, j) + O2(i, j)
+   O3(i, j) = O1(i, j) + O2(i, j)
```
ftynse: Trailing whitespace

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
 
     value = sym_get(name);
   }
+
+ public:
+  void optimize_module() {
+    Halide::Internal::debug(3) << "Optimizing module\n";
```
ftynse: Why use Halide's internal debug printer when we have our glog + flags?

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -451,6 +542,19 @@ class LLVMCodegen {
         llvm::BasicBlock::Create(llvmCtx, "loop_latch", function);
     auto* loopExitBB = llvm::BasicBlock::Create(llvmCtx, "loop_exit", function);
 
+    // TODO: integrate query ISL as to whether the relevant loop ought be
+    // parallelized
```
ftynse: Open an issue?  :)

src/core/polyhedral/llvm_jit.cc
```diff
@@ -35,11 +35,10 @@ Jit::Jit()
...
       DL_(TM_->createDataLayout()),
       objectLayer_([]() { return std::make_shared<SectionMemoryManager>(); }),
-      compileLayer_(objectLayer_, orc::SimpleCompiler(*TM_)),
-      optimizeLayer_(compileLayer_, [this](std::shared_ptr<Module> M) {
-        return optimizeModule(std::move(M));
-      }) {
+      compileLayer_(objectLayer_, orc::SimpleCompiler(*TM_)) {
   sys::DynamicLibrary::LoadLibraryPermanently(nullptr);
+  std::string err;
+  sys::DynamicLibrary::LoadLibraryPermanently("cilkrts", &err);
```
ftynse: Error message ignored?

src/core/polyhedral/llvm_jit.cc
```diff
@@ -35,11 +35,10 @@ Jit::Jit()
     : TM_(EngineBuilder().selectTarget()),
       DL_(TM_->createDataLayout()),
       objectLayer_([]() { return std::make_shared<SectionMemoryManager>(); }),
-      compileLayer_(objectLayer_, orc::SimpleCompiler(*TM_)),
-      optimizeLayer_(compileLayer_, [this](std::shared_ptr<Module> M) {
-        return optimizeModule(std::move(M));
-      }) {
+      compileLayer_(objectLayer_, orc::SimpleCompiler(*TM_)) {
   sys::DynamicLibrary::LoadLibraryPermanently(nullptr);
```
ftynse: What's the point of keeping this?

src/core/polyhedral/codegen_llvm.cc
wsmoses: This function was taken verbatim from Halide [with the addition of one line]. Unfortunately, right now we need to do this but hopefully at one point Halide proper will support Tapir-like optimization [or permit me to just add the one line].

If you think it better, I can change to use our logging.

src/core/polyhedral/codegen_llvm.cc
wsmoses: Done, see #34

src/core/polyhedral/codegen_llvm.cc
prigoyal: I think it's better to use our own logging system. One of many reasons is also to control how we log things to terminal for python/C++ backend which right now needs to be solidified.

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+    }
+
+    // We override PassManager::add so that we have an opportunity to
+    // blacklist problematic LLVM passes.
+    class MyFunctionPassManager : public llvm::legacy::FunctionPassManager {
+     public:
+      MyFunctionPassManager(llvm::Module* m)
+          : llvm::legacy::FunctionPassManager(m) {}
+      virtual void add(llvm::Pass* p) override {
+        Halide::Internal::debug(2)
```
prigoyal: same comment as above :)

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+#if LLVM_VERSION_MAJOR >= 5
+      module->print(llvm::dbgs(), nullptr, false, true);
+#else
+      module->dump();
+#endif
+    }
+
+    // We override PassManager::add so that we have an opportunity to
+    // blacklist problematic LLVM passes.
+    class MyFunctionPassManager : public llvm::legacy::FunctionPassManager {
```
prigoyal: can you rename it to `LLVMPassManager`

Looking at `MyFunctionPassManager` seems like it is an unimplemented code :)

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+    std::unique_ptr<llvm::TargetMachine> TM =
+        Halide::Internal::make_target_machine(*module);
+    module_pass_manager.add(llvm::createTargetTransformInfoWrapperPass(
+        TM ? TM->getTargetIRAnalysis() : llvm::TargetIRAnalysis()));
+    function_pass_manager.add(llvm::createTargetTransformInfoWrapperPass(
+        TM ? TM->getTargetIRAnalysis() : llvm::TargetIRAnalysis()));
+
+    llvm::PassManagerBuilder b;
+    b.OptLevel = 3;
```
prigoyal: can you create constant variables for them in the header file and use that? we shouldn't hardcode magical numbers in the code :)  For example, in the header file:

```
static constexpr int kOptLevel = 3;
```

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+    b.tapirTarget = new llvm::CilkABI();
+#if LLVM_VERSION_MAJOR >= 5
+    b.Inliner = llvm::createFunctionInliningPass(b.OptLevel, 0, false);
+#else
+    b.Inliner = llvm::createFunctionInliningPass(b.OptLevel, 0);
+#endif
+    b.LoopVectorize = true;
```
prigoyal: why are they hardcoded like this? doesn't `PassManagerBuilder` have a proper constructor?

same comment for `b.SLPVectorize` below

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+    // Run optimization passes
+    function_pass_manager.doInitialization();
+    for (llvm::Module::iterator i = module->begin(); i != module->end(); i++) {
+      function_pass_manager.run(*i);
+    }
+    function_pass_manager.doFinalization();
+    module_pass_manager.run(*module);
+
+    Halide::Internal::debug(3) << "After LLVM optimizations:\n";
```
prigoyal: let's switch to our logging, the number 3 feels magical :)

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+    class MyModulePassManager : public llvm::legacy::PassManager {
+     public:
+      virtual void add(llvm::Pass* p) override {
+        Halide::Internal::debug(2)
+            << "Adding module pass: " << p->getPassName().str() << "\n";
+        llvm::legacy::PassManager::add(p);
+      }
+    };
+
+    MyFunctionPassManager function_pass_manager(module.get());
```
prigoyal: `functionPassManager`

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+    MyModulePassManager module_pass_manager;
```
prigoyal: `modulePassManager`

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -294,6 +303,88 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+    std::unique_ptr<llvm::TargetMachine> TM =
```
prigoyal: change `TM` -> `targetMachine`

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -451,6 +542,19 @@ class LLVMCodegen {
         llvm::BasicBlock::Create(llvmCtx, "loop_latch", function);
     auto* loopExitBB = llvm::BasicBlock::Create(llvmCtx, "loop_exit", function);
 
+    // TODO: integrate query ISL as to whether the relevant loop ought be
+    // parallelized
+    bool parallel = false;
+
+    llvm::Value* SyncRegion = nullptr;
```
prigoyal: how does this raw pointer gets freed?

prigoyal:
@wsmoses I'll update the docker images for you today. The current tapir master should be fine?

src/core/polyhedral/codegen_llvm.cc
wsmoses: It's handled by LLVM

src/core/polyhedral/codegen_llvm.cc
wsmoses: Nope, constructor sets defaults, with no arguments

src/core/polyhedral/llvm_jit.cc
```diff
@@ -29,17 +31,55 @@
...
 using namespace llvm;
 
+// Parse through ldconfig to find the path of a particular
+// shared library. This is an unfortunate way to have to
+// find it, but I couldn't immediately find something in
+// imported libraries that would resolve this for us.
+std::string find_library_path(std::string library) {
+  std::string command = "ldconfig -p | grep " + library;
+
+  FILE* fpipe = popen(command.c_str(), "r");
```
nicolasvasilache: @wsmoses  oh wow that is really painful.. can't we just set some macro at compile time like I've been doing for CUB?
https://github.com/facebookresearch/TensorComprehensions/blob/master/CMakeLists.txt#L113

PR #33, opened 2018-02-16
.gitsubmodule paths and instructions for cloning should be https not ssh

doodlesbykumbi:
Running `git clone git@github.com:facebookresearch/TensorComprehensions.git --recursive` outputs

```
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Same output for submodules.

Using `https` to clone and link submodules circumvents this annoyance.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,11 @@
 # Development branches, we modify those heavily
 [submodule "third-party/islpp"]
 	path = third-party/islpp
-	url = git@github.com:nicolasvasilache/isl.git
+	url = https://github.com/nicolasvasilache/isl.git
 	branch = ntv_dev
 [submodule "third-party/cub"]
 	path = third-party/cub
-        url = git@github.com:nicolasvasilache/cub.git
+        url = https://github.com/nicolasvasilache/cub.git
 	branch = nvrtc-cub
 [submodule "third-party/halide"]
 	path = third-party/halide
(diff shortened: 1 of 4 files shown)


ftynse:
Having ssh public key associated with github works too...  Using https for submodules will bring annoyance to contributors when one of the submodules needs to be updated.

doodlesbykumbi:
@ftynse that's a fair point. well, i guess there are tradeoffs. 

i thought it was easy enough to change the remote origin of any submodule on the fly so the contributor experience wouldn't be damaged too much. i assume contributors would be competent git users. i assume less for the casual user of the package.

my use case was trying to build from source in docker. if i'm just interested to clone the repo and make use of it, which i suspect will be a common occurrence, it's a bit of a pain to have to sign up for an account, install my ssh key etc. etc. before you're able to build from source.

prigoyal:
@doodlesbykumbi 
>> my use case was trying to build from source in docker.

We already provide docker images here https://facebookresearch.github.io/TensorComprehensions/docker_image.html have you looked at that? the nvidia-docker images are also available with full instructions.

If you have used docker, can you give us the exact set of commands you ran so we can repro and help figure out the cloning issue?

doodlesbykumbi:
@prigoyal yup. here you go:

```
docker run -v $PWD:/meow-i -t tensorcomprehensions/linux-xenial-gcc5-tapir5.0-cuda9-cudnn7-py3-conda:x86
```

running the following in `bash` within the docker container
```
git clone git@github.com:facebookresearch/TensorComprehensions.git --recursive
```

outputs
```
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

prigoyal:
cool, I am addressing this. thanks for the report :)

PR #36, opened 2018-02-16
Regenerate docs with fixes

prigoyal:
chose better variable names and explicitly mention `+=!` in the docs

### DIFF
--- a/docs/source/introduction.rst
+++ b/docs/source/introduction.rst
@@ -23,13 +23,13 @@ Let's start with a simple example is a matrix vector product:
 
 .. code::
 
-    def mv(float(R,C) A, float(C) x) -> (o) {
-      o(i) += A(i,j) * x(j)
+    def mv(float(R,C) A, float(C) B) -> (o) {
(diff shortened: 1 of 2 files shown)


prigoyal:
thanks @wsmoses

docs/source/introduction.rst
```diff
@@ -23,13 +23,13 @@ Let's start with a simple example is a matrix vector product:
 
 .. code::
 
-    def mv(float(R,C) A, float(C) x) -> (o) {
-      o(i) += A(i,j) * x(j)
+    def mv(float(R,C) A, float(C) B) -> (o) {
+      o(i) += A(i,j) * B(j)
```
nicolasvasilache: @prigoyal I fail to see how this better than it used to be .. in mathematical expressions, people use lower case for vectors and upper case for higher order objects like matrices.

This change breaks uniformity of input/output object types.

If you don't like the names, this makes more sense to me:
```
def mv(float(R,C) M, float(C) x) -> (y) {
      y(i) += M(i,j) * x(j)
}
```

PR #39, opened 2018-02-16
Abstract away ExecutionEngine from CUDA

ftynse:
Separate `ExecutionEngine` and `MappedScop` into CUDA-dependent and -independent parts.  Update the build system to support building without CUDA even present.


ftynse:
There's a second branch https://github.com/facebookresearch/TensorComprehensions/tree/uncuda-dirty  that I did not manage to make compile.

These changes are mostly orthogonal to decoupling MappedScop from CUDA specifics and factoring RTCFunction stuff. You may want to also look there.

prigoyal:
@ftynse  one thing i am wondering about: are you thinking of creating separate cpu_execution_engine, etc?

That means a lot of code duplication. Do we really need this separation? to me it seems like execution_engine, tc_executor logic should be reused and rather we should make calls from there to cuda codegen or llvm codegen

ftynse:
@prigoyal yes, there will be a separate `CpuExecutionEngine` (`TapirExecutionEngine`?).  We will abstract away the common parts of the `TcExecutor` and `ExecutionEngine` back into care.  Given the amount of CUDA-specific stuff, it was easier this way.

Having the same execution engine calling either CUDA or LLVM, or OpenCL means it has to link with _all_ of them, which is exactly what we want to avoid.  Ideally, it should be possible to build a version of TC without having CUDA even present in the system.

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    void clear() {
+      static_cast<CudaTcExecutor&>(*exec).clearRTC();
+    }
+  };
+
+  CudaExecutionEngine() = default;
+
+  void addTC(const std::string& tc);
```
prigoyal: is this `define`? what does this do?

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -0,0 +1,134 @@
...
+
+  lang::TreeRef treeForFunction(const std::string& name) {
```
prigoyal: can we name it `treeForDef`? I find `def` to directly related to def in the TC and it becomes clear what's going on here.

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -0,0 +1,118 @@
...
+    }
+    rtcFun->clear();
+  }
+
+  std::string kernelName() const {
+    return execInfo_.kernelName;
+  }
+
+ private:
+  void compileWithPetPpcg();
```
prigoyal: why does this exist? didn't we remove petppcg? this header is looking a lot like back in October

src/core/halide_utils.cc
```diff
@@ -0,0 +1,180 @@
...
+  // fill the ParameterValMap
+  for (size_t i = 0; i < inputsDLT.size(); i++) {
+    const ImageParam& in = halide.inputs[i];
+    const DLTensor* tensor = inputsDLT[i];
+    // for error messages
+    auto param_type = halide.getDef().params()[i].tensorType();
+    for (int d = 0; d < in.dimensions(); d++) {
+      Expr extent = in.parameter().extent_constraint(d);
+      int64_t current_size = tensor->shape[d];
+      auto dim_exp_tree = param_type.dims()[d];
```
prigoyal: `paramType`

src/core/halide_utils.cc
```diff
@@ -0,0 +1,180 @@
...
+
+  // fill the ParameterValMap
+  for (size_t i = 0; i < inputsDLT.size(); i++) {
+    const ImageParam& in = halide.inputs[i];
+    const DLTensor* tensor = inputsDLT[i];
+    // for error messages
+    auto param_type = halide.getDef().params()[i].tensorType();
+    for (int d = 0; d < in.dimensions(); d++) {
+      Expr extent = in.parameter().extent_constraint(d);
+      int64_t current_size = tensor->shape[d];
```
prigoyal: camelcase please :)

include/tc/core/cuda/cuda_execution_engine.h
ftynse: it was in the original code :)

include/tc/core/cuda/cuda_execution_engine.h
ftynse: yes, in a separate PR

include/tc/core/cuda/cuda_tc_executor.h
ftynse: I don't know why this was not removed.  I removed it in a separate commit later.

src/core/halide_utils.cc
ftynse: yes, it's inherited code but I'll fix it, thanks!

prigoyal:
@ftynse okay from my side for python bindings. mostly moving code around and aten/c2 part is not changed much except imports. Please feel free to cleanup and we can do another round of review and merge when you are ready.

ftynse:
We probably want to add a CI configuration without CUDA at all so that future changes don't break compilation of "core" TC by polluting it with CUDA headers.

ftynse:
There are more cleanups to do, but this builds and works so I'd like it merged before it grows to 5k lines.

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -31,64 +32,54 @@ namespace tc {
...
+class CudaExecutionEngine : public ExecutionEngine {
  public:
-  struct ExecutorInfo {
-    ExecutorInfo(
+  struct CudaExecutorInfo : public ExecutionEngine::ExecutorInfo {
+    CudaExecutorInfo(
         std::string id,
         std::vector<const DLTensor*> inputsInfo,
-        std::unique_ptr<MappingOptions> options,
+        const MappingOptions& options,
```
skimo-openhub: independent change?

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -128,18 +129,6 @@ class CudaExecutionEngine {
...
       const std::vector<const DLTensor*>& inputsInfo,
       const MappingOptions& options);
-  size_t emplaceExecutor(std::unique_ptr<ExecutorInfo> p);
-
-  /// For thread-safety perform all cheap operations under lock
-  std::mutex executorInfoMutex;
-
-  // XXX:if ExecutorInfo is moved/copied (even when the vector's underlying
-  // storage is extended) something inside isl segfaults,  unique_ptr is used as
-  // a workaround
```
skimo-openhub: This comment got dropped.
If it is no longer relevant, then it should be dropped separately.

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -31,64 +32,54 @@ namespace tc {
...
-          exec(tc, inputsInfo),
-          objectLocalHandle(handle) {}
+        : ExecutionEngine::ExecutorInfo(
+              id,
+              inputsInfo,
+              options.toProtobufSerializedString(),
+              tc,
+              handle) {
+      exec =
+          std::unique_ptr<CudaTcExecutor>(new CudaTcExecutor(tc, inputsInfo));
```
skimo-openhub: Why are you overwriting this field?
Why not pass it to the constructor of `ExecutorInfo`?

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -31,64 +32,54 @@ namespace tc {
...
+    }
 
     void clear() {
-      exec.clearRTC();
+      static_cast<CudaTcExecutor&>(*exec).clearRTC();
```
skimo-openhub: How about calling it `clear` and making it a virtual function
of `TcExecutor`?

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -31,64 +32,54 @@ namespace tc {
...
 
   CudaExecutionEngine() = default;
 
-  /// Create the CudaExecutionEngine::tcNameMap_ using the language passed
-  /// to it - should support many TC.
-  void define(const std::string& language);
-
-  /// Create the CudaExecutionEngine::tcNameMap_ from the parsed TC
-  /// string - supports many TC.
-  void define(const std::vector<lang::TreeRef>& treeRefs);
```
skimo-openhub: Hmm... it seems you changed the comments on these functions
when you moved them out.  Do they reflect a change in behavior
introduced by this commit?

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -31,64 +32,54 @@ namespace tc {
...
-
   lang::TreeRef treeForFunction(const std::string& name) {
     return tcNameMap_.at(name);
   }
 
   // TODO: Pass autotuning info (none by default, otherwise some struct with
   //       maxtime and other things)
 
-  /// Returns a handle for the compiled kernel
+  /// returns a handle for the compiled kernel
```
skimo-openhub: Why change `Returns` to `returns`?

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -31,64 +32,54 @@ namespace tc {
...
+      const std::string& name,
+      const std::vector<const DLTensor*>& inputs,
+      const std::string& mappingOptions) override {
+    return compile(name, inputs, MappingOptions(mappingOptions));
+  }
   size_t compile(
       const std::string& name,
       const std::vector<const DLTensor*>& inputs,
       const MappingOptions& options);
+  /// @}
```
skimo-openhub: Why do you need two versions of this method to be able
to extract the abstract parts?
Same for `run`.

src/autotuner/genetic_tuning_harness.cc
```diff
@@ -234,18 +234,18 @@ bool GeneticTunerHarness::warmupOrPrune(
             CHECK(info);
             USING_MAPPING_SHORT_NAMES(BX, BY, BZ, TX, TY, TZ);
             auto& exec = info->exec;
-            auto block = exec.block();
+            auto block = static_cast<CudaTcExecutor&>(*exec).block();
             auto nThreads = TX.mappingSize(block) * TY.mappingSize(block) *
                 TZ.mappingSize(block);
-            auto grid = exec.grid();
+            auto grid = static_cast<CudaTcExecutor&>(*exec).grid();
```
skimo-openhub: I know Nicolas doesn't like getters, but surely
it's better to perform this `static_cast` in at most one place.

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -26,11 +26,6 @@
 
 namespace tc {
 
-using PpcgppCompilationCommand = std::function<std::string(
-    std::unique_ptr<tc::polyhedral::Scop>&& scop,
-    const std::vector<const DLTensor*>& inputs,
-    const MappingOptions& options)>;
```
skimo-openhub: I missed this one.  I was only grepping for `PPCG` and `ppcg`.
Surely, this can go in independently.

include/tc/core/cuda.h
```diff
@@ -84,7 +84,31 @@ struct WithDevice {
...
+      const std::vector<std::string>& gpuNames,
+      const std::vector<size_t>& sharedMemSizes)
+      : gpuNames_(gpuNames), sharedMemSizes_(sharedMemSizes) {}
+
+ public:
+  static CudaGPUInfo& GPUInfo();
+
+  // These functions require init to have been run, they are thus members of
+  // the singleton object and not static functions.
+  int NumberGPUs() const;
```
nicolasvasilache: 2 non-blocking remarks here:
1. querying the cuda runtime API should generally be cached as much as possible because calls like ```cudaGetDeviceProperties``` may add up to non-trivial overheads. We should measure it first but we need to be aware of it, see e.g. [here](https://github.com/facebook/fbcuda/blob/8ef2af9b579b8610c59ab0ccf6e9075350de4320/util/CachedDeviceProperties.cpp) 
2.  nit ```lowerCamelCase```

nicolasvasilache:
Also @bwasti FYI

include/tc/core/cuda.h
ftynse: this is merely moving the existing code around, will fix later

PR #42, opened 2018-02-19
Add cilk runtime to docker images

prigoyal:
1404 images didn't have cilk runtime and this was blocking @wsmoses PR. building and uploading new images now.

### DIFF
--- a/docker/linux-xenial-gcc5-tapir5.0/Dockerfile
+++ b/docker/linux-xenial-gcc5-tapir5.0/Dockerfile
@@ -4,6 +4,7 @@ FROM tensorcomprehensions/linux-xenial:${BUILD_ID}
 ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update
+RUN apt-get install -y --no-install-recommends vim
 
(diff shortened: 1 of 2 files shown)


docker/linux-xenial-gcc5-tapir5.0/Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM tensorcomprehensions/linux-xenial:${BUILD_ID}
 ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update
+RUN apt-get install -y --no-install-recommends vim
```
wsmoses: Any reason why we're adding vim?

prigoyal: yep, I normally have to build conda packages and sometimes have to edit the build scripts locally. Plus, users are running docker images as well so it's better to provide one editor at least :)

prigoyal:
merging this now, all checks pass, I'll update the images x86_1 and will ping you @wsmoses when they are ready. they are building atm

PR #43, opened 2018-02-19
Update docs install instructions, docker images and github issue template

prigoyal:
based on user feedback, updating these instructions
1. update docker image in docs to x86_1
2. add instructions for disabling ocaml llvm bindings if ocaml related failure happens in build
3. remove sudo and add  a not to use it where needed
4. update github issue template file for how to find the clang version


docs/source/installation_conda.rst
```diff
@@ -136,9 +136,9 @@ Now, Install cuDNN v6.0 (skip if you have it already):
     $ CUDNN_TAR_FILE="cudnn-8.0-linux-x64-v6.0.tgz"
     $ wget http://developer.download.nvidia.com/compute/redist/cudnn/v6.0/${CUDNN_TAR_FILE}
     $ tar -xzvf ${CUDNN_TAR_FILE}
-    $ sudo cp -P cuda/include/cudnn.h /usr/local/cuda/include
-    $ sudo cp -P cuda/lib64/libcudnn* /usr/local/cuda/lib64/
-    $ sudo chmod a+r /usr/local/cuda/lib64/libcudnn*
+    $ cp -P cuda/include/cudnn.h /usr/local/cuda/include
```
ftynse: This won't work unless you work under root.  If you do run under root, sudo (if installed) is a noop.

docs/source/installation_non_conda.rst
```diff
@@ -13,11 +13,11 @@ Step 1: Install some build dependencies
 .. code-block:: bash
 
     $ apt-get update
-    $ apt-get install -y libgoogle-glog-dev curl build-essential cmake git automake libgmp3-dev libtool ssh libyaml-dev realpath wget valgrind software-properties-common unzip
+    $ apt-get install -y libgoogle-glog-dev curl build-essential cmake git automake libgmp3-dev libtool ssh libyaml-dev realpath wget valgrind software-properties-common unzip libz-dev
 
 .. note::
 
-    You might additionally need to install :code:`cmake3` and :code:`libz-dev`, if these packages are not found by apt-get, you can proceed otherwise install them by running :code:`apt-get install libz-dev`
+    You might additionally need to install :code:`cmake3`, if this package is not found by apt-get, you can proceed.
```
ftynse: When you _might_ need it?  Some ubuntu versions have it while others don't?

docs/source/installation_conda.rst
prigoyal: in the docker env, sudo doesn't work. so I provided general instructions, if something doesn't work, then people will add sudo themselves.

docs/source/installation_non_conda.rst
prigoyal: yes

prigoyal:
I can add a note that if some command don't work, people should add sudo

ftynse:
> I can add a note that if some command don't work, people should add sudo

Please do, and only for commands that failed due to lack of permissions.  Putting sudo in front of a failing compilation step won't help :)

docs/source/installation_conda.rst
ftynse: why doesn't it work?  do you need to install `sudo` package?

prigoyal: nope, in docker everything is already root, and sudo throws errors `sudo is unknown`

ftynse: which probably means that the `sudo` command is just not installed :)
it's usually possible to run `sudo` or `su` under root on a full installation

prigoyal: yeah docker requires specifically installing sudo https://stackoverflow.com/a/25908200 but I didn't feel that was necessary to do. I will add a note for asking people to use sudo if a command fails to run

prigoyal:
simple docs changes, merging this now, thanks @ftynse :)

PR #44, opened 2018-02-20
[RFC] Always emit the same function signature in codegen_llvm

ttheodor:


src/core/polyhedral/codegen_llvm.cc
```diff
@@ -355,16 +355,34 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
   }
 };
 
+template <typename CI, typename CO, typename Pred>
+bool all(const CI& inputs, const CO& outputs, Pred p) {
+  for (const auto& v : inputs) {
+    if (not p(v)) {
```
prigoyal: this feels like python. is it possible to make a more explicit check here? like you are checking for nullptr etc.?

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -355,16 +355,34 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+      return false;
+    }
+  }
+  for (const auto& v : outputs) {
+    if (not p(v)) {
```
prigoyal: same comment as above

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -355,16 +355,34 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+      return false;
+    }
+  }
+  return true;
+}
+
 class LLVMCodegen {
-  void collectTensor(const Halide::OutputImageParam& t) {
+  llvm::Type* convertTensorToType(const Halide::OutputImageParam& t) {
```
prigoyal: why return a raw pointer? how do we ensure it doesn't leak later?

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -355,16 +355,34 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
     auto sizes =
         getTensorSizesWithoutLeadingDim(t, scop_.globalParameterContext);
     if (not sizes.empty()) {
-      args_.emplace_back(
-          makePtrToArrayType(halide_cg.llvm_type_of(t.type()), sizes));
+      return makePtrToArrayType(halide_cg.llvm_type_of(t.type()), sizes);
     } else {
-      args_.emplace_back(halide_cg.llvm_type_of(t.type())->getPointerTo());
+      return halide_cg.llvm_type_of(t.type())->getPointerTo();
```
prigoyal: just curious: what is `halide_cg`? also use `camelCase` :)

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -449,6 +467,57 @@ class LLVMCodegen {
     }
   }
 
+  void createWrapper(
+      const std::vector<Halide::ImageParam>& inputs,
+      const std::vector<Halide::OutputImageParam>& outputs,
+      const std::string& fname) {
+    CHECK(not inputs.empty());
```
prigoyal: can you use something like `CHECK_NE`? this is looking like python atm

src/core/polyhedral/codegen_llvm.cc
ttheodor: There is no explicit check here. Each instantiation can potentially check for a different thing. 

Also I don't know whether it looks like python but a template predicates exist in every corner of the STL. I'd argue that, by definition, STL's style is idiomatic C++ :stuck_out_tongue:. And actually there's `std::all_of` which I should have used.

src/core/polyhedral/codegen_llvm.cc
ttheodor: In LLVM almost everything is passed around as a pointer. The associated `LLVMContext` keeps track of everything.

src/core/polyhedral/codegen_llvm.cc
ttheodor: @wsmoses  introduced it

skimo-openhub:
On Tue, Feb 20, 2018 at 05:49:58AM -0800, Priya Goyal wrote:
> +  void createWrapper(
> +      const std::vector<Halide::ImageParam>& inputs,
> +      const std::vector<Halide::OutputImageParam>& outputs,
> +      const std::string& fname) {
> +    CHECK(not inputs.empty());
> 
> can you use something like `CHECK_NE`? this is looking like python atm

CHECK_NE doesn't provide any additional information on booleans.
If the boolean value isn't the one you were expecting,
then obviously, it was the other one.
In fact, I find the output of CHECK_NE confusing on booleans.

skimo

src/core/polyhedral/codegen_llvm.cc
wsmoses: It translates from Halide to LLVM

src/core/polyhedral/codegen_llvm.cc
prigoyal: yep, that's why it looks like python because of `not`. With your explanation, I understand better now the choice for `not`. Thanks :)

PR #48, opened 2018-02-20
Bump c2, add c2-python example and c2 conda package

prigoyal:
using conda package will allow writing caffe2 - TC examples from python. added a python example as well

adding a special doc section for installing caffe2 and running python example for it

### DIFF
--- a/include/tc/c2/tc_op.h
+++ b/include/tc/c2/tc_op.h
@@ -37,13 +37,15 @@ class TcOp : public Operator<Context> {
  public:
   TcOp(const OperatorDef& operator_def, Workspace* ws)
       : caffe2::Operator<Context>(operator_def, ws),
+        tc_(OperatorBase::GetSingleArgument<std::string>("tcDef", "ERROR")),
+        tcName_(
+            OperatorBase::GetSingleArgument<std::string>("tcName", "ERROR")),
         mappingOptions_(tc::MappingOptions::makeNaiveMappingOptions()),
         gradMappingOptions_(tc::MappingOptions::makeNaiveMappingOptions()) {
-    tc_ = OperatorBase::GetSingleArgument<std::string>("def", "ERROR");
-    gradTc_ = OperatorBase::GetSingleArgument<std::string>("grad_def", "ERROR");
-    tcName_ = OperatorBase::GetSingleArgument<std::string>("name", "ERROR");
+    gradTc_ =
+        OperatorBase::GetSingleArgument<std::string>("tcGradDef", "ERROR");
     gradTcName_ =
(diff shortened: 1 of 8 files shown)


include/tc/c2/tc_op.h
```diff
@@ -37,11 +37,12 @@ class TcOp : public Operator<Context> {
  public:
   TcOp(const OperatorDef& operator_def, Workspace* ws)
       : caffe2::Operator<Context>(operator_def, ws),
+        tc_(OperatorBase::GetSingleArgument<std::string>("lang", "ERROR")),
```
ftynse: here the `lang` argument is read

test/test_caffe2.cc
```diff
@@ -638,7 +638,7 @@ def fun(float(B, N, M) X, float(B, M, K) Y) -> (Z)
   Workspace w_test;
   init_ws(w_test);
   Argument tcArg = MakeArgument<string>("def", tc);
```
ftynse: but here the `def` argument is written

include/tc/c2/tc_op.h
prigoyal: @ftynse def is a keyword in python, that can't be used. I am happy to use "tc" though if you like

prigoyal: yes, I am still making changes, thanks for catching this :)

include/tc/c2/tc_op.h
```diff
@@ -37,11 +37,12 @@ class TcOp : public Operator<Context> {
...
+        tcName_(
+            OperatorBase::GetSingleArgument<std::string>("tcName", "ERROR")),
```
prigoyal: for clarity, this change is needed because `name` is an arg that python bindings in caffe2 core code uses already. so it conflicts with the name here if I also use tc -caffe2 from python

test/test_caffe2.cc
prigoyal: yes, thanks for catching, I am still making changes

include/tc/c2/tc_op.h
ftynse: `tcDef` ?

prigoyal: it's really a name so we should make it explicit that it's the *name* that user needs to set the value to. When I read `tcDef` it feels to me that I need to pass the tc language which is misleading

ftynse: sorry, I entered the comment in the wrong field.  I meant `tcDef` instead of `def`, not here

prigoyal: got it, I'll make the change

prigoyal:
@ftynse all the comments have been addressed, feel free to stamp/review more as you find time. Thanks :)

PR #51, opened 2018-02-21
Fix extractAccess to process each IRNode once

ftynse:
extractAccess introduces reference identifiers with sequential numbers
in the domain of the access relation it constructs.  The reference
idetifiers are stored in a map halide::IRNode->isl::id.  Assuming that
each IRNode is visited once by extractAccess, the sequential numbers are
obtained by querying the size of the map.  However, some IRNodes may be
visited more than once, for example, with let statement substitution.
Make FindAccesses visitor class, which class extractAccess on IRNodes,
inherit IRGraphVisitor instead of IRVisitor to ensure each IRNode is
visited once.

This fixes the problem that appeared in memory promotion when the same
reference identifier was attached to different tensor names.

### DIFF
--- a/src/core/halide2isl.cc
+++ b/src/core/halide2isl.cc
@@ -185,11 +185,11 @@ isl::map extractAccess(
 
 std::pair<isl::union_map, isl::union_map>
 extractAccesses(isl::set domain, const Stmt& s, AccessMap* accesses) {
-  class FindAccesses : public IRVisitor {
-    using IRVisitor::visit;
+  class FindAccesses : public IRGraphVisitor {
+    using IRGraphVisitor::visit;
 
     void visit(const Call* op) override {
-      IRVisitor::visit(op);
+      IRGraphVisitor::visit(op);
       if (op->call_type == Call::Halide || op->call_type == Call::Image) {
         reads = reads.unite(
             extractAccess(domain, op, op->name, op->args, accesses));
@@ -197,7 +197,7 @@ extractAccesses(isl::set domain, const Stmt& s, AccessMap* accesses) {
     }
 
     void visit(const Provide* op) override {
-      IRVisitor::visit(op);
+      IRGraphVisitor::visit(op);
       writes =
           writes.unite(extractAccess(domain, op, op->name, op->args, accesses));
     }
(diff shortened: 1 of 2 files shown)


test/test_tc_mapper_bugs.cc
```diff
@@ -658,6 +658,44 @@ TEST_F(TMM_128_1024_1024, Tightening) {
...
+                     .tileImperfectlyNested(false)
+                     .useSharedMemory(true)
+                     .usePrivateMemory(true)
+                     .unrollCopyShared(false)
+                     .matchLibraryCalls(false);
+
+  tc::ATenCompilationUnit atCompl;
+  atCompl.define(TC);
+  // Expecting this to compile without dying.
+  atCompl.compile("layernorm", inputs, options);
```
prigoyal: maybe run as well? :)

abadams:
There are uses of IRVisitor still in that class - should be IRGraphVisitor.

src/core/halide2isl.cc
```diff
@@ -185,7 +185,7 @@ isl::map extractAccess(
 
 std::pair<isl::union_map, isl::union_map>
 extractAccesses(isl::set domain, const Stmt& s, AccessMap* accesses) {
-  class FindAccesses : public IRVisitor {
+  class FindAccesses : public IRGraphVisitor {
     using IRVisitor::visit;
```
abadams: Should be IRGraphVisitor here, and in the calls to IRVisitor::visit below.

test/test_tc_mapper_bugs.cc
ftynse: What would be the point of that?  Originally we had an assertion during the compilation.

src/core/halide2isl.cc
ftynse: The function is virtual, so it should not matter :)

skimo-openhub:
Just some typos that also appear in the commit message.

On Wed, Feb 21, 2018 at 10:18:01PM +0000, ftynse wrote:
> extractAccess introduces reference identifiers with sequential numbers
> in the domain of the access relation it constructs.  The reference
> idetifiers are stored in a map halide::IRNode->isl::id.  Assuming that

identifiers

> each IRNode is visited once by extractAccess, the sequential numbers are
> obtained by querying the size of the map.  However, some IRNodes may be
> visited more than once, for example, with let statement substitution.
> Make FindAccesses visitor class, which class extractAccess on IRNodes,

There is a verb missing in the which-clause, maybe "uses"
I would also drop "class".

> inherit IRGraphVisitor instead of IRVisitor to ensure each IRNode is

inherit from

skimo

Issue #49, opened 2018-02-21
test_mapper_llvm fails because wrong cilkrts is found


skimo-openhub:
With ac1f3c34aed50c7abb984420eceb1e01fd39ccc2, I get
```
Running build/test/test_mapper_llvm
[==========] Running 4 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 4 tests from LLVMCodegen
[ RUN      ] LLVMCodegen.Basic
unknown file: Failure
C++ exception with description "Failed to find cilkrts: /usr/lib32/libcilkrts.so.5: wrong ELF class: ELFCLASS32" thrown in the test body.
[  FAILED  ] LLVMCodegen.Basic (12 ms)
```
The output of `ldconfig -p | grep libcilkrts` is
```
        libcilkrts.so.5 (libc6,x32) => /usr/libx32/libcilkrts.so.5
        libcilkrts.so.5 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libcilkrts.so.5
        libcilkrts.so.5 (libc6) => /usr/lib32/libcilkrts.so.5
```

ttheodor:
Here's the offending code: https://github.com/facebookresearch/TensorComprehensions/blob/master/src/core/polyhedral/llvm_jit.cc#L38

prigoyal:
@wsmoses we can make change to the build system to only look for 64-bit cilk runtime.

skimo-openhub:
On Thu, Feb 22, 2018 at 10:21:47AM +0000, ftynse wrote:
> I have a different variation of this
> ```
> sh: 1: ldconfig: not found      
> unknown file: Failure                                      
> C++ exception with description "Failed locate library: libcilkrts.so" thrown in the test body.
> ```
> 
> On Debian, `ldconfig` is located in `/sbin` and not in the path for a regular user.  It looks like we need a better process of loading the library...

I think Priya was proposing to look for the library at configure/build
time instead of at run time.

skimo

prigoyal:
yes, I was proposing to do it at configure/build time. If this test if blocking then feel free to disable the test for now. Billy is traveling I think for some talk so might not be available for this.

wsmoses:
@skimo-openhub 
I made a quick fix to only search for 64bit libraries here: https://github.com/facebookresearch/TensorComprehensions/tree/tfix

This doesn't stop the fact that we should fix this, but perhaps this will unblock you.

PR #54, opened 2018-02-22
Temporary fix for library path to unblock folks

wsmoses:
This _should_ resolve the immediate issue presented by #49 

We should do a better fix for finding the library, but we ought get this in to make sure people aren't blocked.

### DIFF
--- a/src/core/polyhedral/llvm_jit.cc
+++ b/src/core/polyhedral/llvm_jit.cc
@@ -36,7 +36,7 @@ using namespace llvm;
 // find it, but I couldn't immediately find something in
 // imported libraries that would resolve this for us.
 std::string find_library_path(std::string library) {
-  std::string command = "ldconfig -p | grep " + library;
+  std::string command = "ldconfig -p | grep " + library + " | grep x86-64";
 
   FILE* fpipe = popen(command.c_str(), "r");
 


src/core/polyhedral/llvm_jit.cc
```diff
@@ -36,7 +36,7 @@ using namespace llvm;
 // find it, but I couldn't immediately find something in
 // imported libraries that would resolve this for us.
 std::string find_library_path(std::string library) {
-  std::string command = "ldconfig -p | grep " + library;
+  std::string command = "ldconfig -p | grep " + library + " | grep x86-64";
```
prigoyal: is this always guaranteed on all machines?

prigoyal:
can @skimo-openhub or someone else try out this fix as well? CI was anyways fine earlier as well

skimo-openhub:
Sorry, I seem to have missed this one.
Your patch works for me, but I share @prigoyal's concern that this may not work on every system,
so I feel reluctant to just accept this patch as it might break other people's setup.
However, if you can tell me that TC is only supposed to work on x86-64 anyway, I'll accept it.

wsmoses:
As far as I'm aware, we presently only support 64bit systems. And agreed it doesn't work on every system (it was an immediate fix to get you unblocked back in the day :P).

Issue #45, opened 2018-02-20
Semantic analysis fails to catch invalid TC involving temporaries


prigoyal:
we don't support temporaries very well right now, i.e. everything has to be either an input/output. However, I myself have fallen into trap of missing marking the temporary variables as outputs. The semantic analysis doesn't catch this TC and propagates to Halide further to get the IR. this should be corrected

prigoyal:
closed by #57

PR #57, opened 2018-02-23
Add a check for use of temporary tensors

zdevito:
Addresses #45

### DIFF
--- a/include/tc/lang/sema.h
+++ b/include/tc/lang/sema.h
@@ -350,9 +350,9 @@ struct Sema {
     }
 
     for (auto p : func.params())
-      non_temporaries.insert(p.ident().name());
+      nonTemporaries.insert(p.ident().name());
     for (auto r : func.returns())
-      non_temporaries.insert(r.ident().name());
+      nonTemporaries.insert(r.ident().name());
 
     auto statements_ =
         checkList(func.statements(), [&](TreeRef r) { return checkStmt(r); });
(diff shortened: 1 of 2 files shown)


include/tc/lang/sema.h
```diff
@@ -601,5 +613,7 @@ struct Sema {
   // if you write to an input, using it in a range expression is no longer
   // allowed
   std::unordered_set<std::string> live_input_names;
+
+  std::unordered_set<std::string> non_temporaries;
```
prigoyal: can you make it cameCase i.e. `nonTemporaries` ? this is as per our coding conventions so we should follow them as well :)

zdevito:
Are we working off a different convention than what is is CodingConvensions.md? Because it says either is ok? I changed it to lowerCamelCase as requested.

```
### Class data members ###

As with variables, use `lowerCamelCase` or `lower_case_with_underscores`.
```

prigoyal:
we are using `lowerCamelCase`, you can look at other code reviews and most of our code except lang/ , we are trying to strictly follow this. I'll clean up the underscores part , that shouldn't be there

prigoyal:
looks good, thanks much!

skimo-openhub:
On Fri, Feb 23, 2018 at 05:13:44AM -0800, Priya Goyal wrote:
> we are using `lowerCamelCase`, you can look at other code reviews and most of our code except lang/ , we are trying to strictly follow this. I'll clean up the underscores part , that shouldn't be there

If this is a requirement now for data members,
then please update the relevant section in CodingConvensions.md:

On Thu, Feb 22, 2018 at 09:45:34PM -0800, Zachary DeVito wrote:
> Are we working off a different convention than what is is CodingConvensions.md? Because it says either is ok? I changed it to lowerCamelCase as requested.
> 
> ```
> ### Class data members ###
> 
> As with variables, use `lowerCamelCase` or `lower_case_with_underscores`.
> ```

skimo

PR #58, opened 2018-02-23
remove PPCG remnants

skimo-openhub:
Some parts of the code/documentation still referenced PPCG,
which has been removed.  Remove those references as well.


nicolasvasilache:
great, thanks for cleaning up our technical debt @skimo-openhub

PR #59, opened 2018-02-23
Caffe2Test.TcConvolutionOp_Gradient (disabled): drop check for error

skimo-openhub:
In prehistory, this check was disabled at some point, then got enabled
with a check for a specific error message and then got disabled again
for other reasons.  Since then, many things have changed and
the specific error message cannot possibly be generated anymore.
Remove it.


skimo-openhub:
It seems I missed some references to Ppcg.
The other remaining once are currently covered by #39.

prigoyal:
merging this now, thanks @skimo-openhub :)

PR #60, opened 2018-02-23
cmake: use HINTS instead of PATHS to prefer custom gflags and glog

skimo-openhub:
Root CMakeLists.txt was using PATHS in find_library call for gflags and
glog.  In this case, cmake looks in system paths _first_, before looking
at whatever is passed in PATHS, so it selects system versions of glog
and gflags.  HINTS are inspected _before_ system paths and our
installations of glog and gflags will be found, if installed, before the
system ones.


skimo-openhub:
Cherry-picked from #39.

PR #61, opened 2018-02-23
Add a unit test for softmax

prigoyal:
we have a weird bug where we don't allow reusing the same variable twice

```
FAILS:
def softmax(float(N, D) I) -> (O, tmp) {
        tmp(n) max=! I(n, d)
        O(n, d) = exp(I(n, d) - tmp(n))
        tmp(n) +=! O(n, d)
        O(n, d) = O(n, d) / tmp(n)
}
ERROR: Func tmp cannot be given a new update definition, because it has already been realized or used in the definition of another Func. Aborted (core dump)

FAILS:
def softmax(float(N, D) I) -> (O, tmp, tmp1) {
        tmp(n) max=! I(n, d)
        O(n, d) = exp(I(n, d) - tmp(n))
        tmp1(n) +=! O(n, d)
        O(n, d) = O(n, d) / tmp1(n)
}
ERROR: Func O cannot be given a new update definition, because it has already been realized or used in the definition of another Func

PASSES:
def softmax(float(N, D) I) -> (O, expsum, maxVal) {
        maxVal(n) max= I(n, d)
        expsum(n) +=! exp(I(n, d) - maxVal(n))
        O(n, d) = exp(I(n, d) - maxVal(n)) / expsum(n)
 }
```

It throws error:
```
Func tmp cannot be given a new update definition, because it has already been realized or used in the definition of another Func. Aborted (core dump)
```

cc @jekbradbury

### DIFF
--- a/test/test_execution_engine.cc
+++ b/test/test_execution_engine.cc
@@ -45,6 +45,87 @@ struct ATenCompilationUnitTest : public ::testing::Test {
   }
 };
 
+TEST_F(ATenCompilationUnitTest, DISABLED_SoftmaxA) {
+  at::Tensor a = at::CUDA(at::kFloat).rand({32, 16});
+  std::vector<at::Tensor> inputs = {a};
+  std::vector<at::Tensor> outputs;
+
+  // Tensor dependencies should strictly be DAG
+  Check(
+      R"(
+      def softmax(float(N, D) I) -> (O, tmp) {
+        tmp(n) max= I(n, d)
+        O(n, d) = exp(I(n, d) - tmp(n))
+        tmp(n) +=! O(n, d)
+        O(n, d) = O(n, d) / tmp(n)
+      }
+    )",
+      "softmax",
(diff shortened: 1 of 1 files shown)


abadams:
It's not necessary to go down to three lines of code, you just can't reuse names on the LHS once they've been used in another tensor. Otherwise the sizes of the tensors is ill-defined (in the general case - there's a solution in this case).

```
def softmax(float(N, D) I) -> (O, maxVal, expDistance, expSum) {
        maxVal(n) max=! I(n, d)
        expDistance(n, d) = exp(I(n, d) - maxVal(n))
        expSum(n) +=! expDistance(n, d)
        O(n, d) = expDistance(n, d) / expSum(n)
}

```

prigoyal:
yep, thanks for the clarification @abadams that makes sense, however, I am still wondering about the `max=!` issue and why that fails compilation 

@zdevito can you take a look at that case, there's issue #63 filed as well

prigoyal:
CI is green, merging, simple test cases added

PR #68, opened 2018-02-24
Update halide build to use exceptions and new docker images

prigoyal:
1. when building docker images, the paths need to be hardcoded and hence $HOME is null but when we actually run the docker image, then $HOME is /root .

2. Turn on WITH_EXCEPTIONS=1 for the halide build

3. Update the images to x86 now, x86_1 was confusing to people

4. add special note for choosing CLANG_PREFIX when running TC via docker image

5. shorten docker file names


prigoyal:
CI is green, merged

PR #70, opened 2018-02-26
TcExecutor: remove useless declarations

skimo-openhub:


skimo-openhub:
Cherry-picked from #39.

PR #72, opened 2018-02-26
PyTorch Integration with Tensor Comprehensions

prigoyal:
This PR adds PyTorch integration with TC. Here is what the PyTorch-TC package provides:

- inputs and outputs to functions are `torch.*Tensor`s
- Integration with PyTorch **autograd**: if you specify forward and backward functions, you get an autograd function that takes Variable as input and returns Variable as output.
- autotuner results can be cached to a file (for reuse)
- Various test for ML layers are provided in `test_python/layers/`

To make it easy to use TC, we provide conda packages for it. We will provide instructions on how to install the **conda** package soon. Documentation will also be made available in few hours.

Edited the CodeOwners.md file to document the python bindings, conda recipes properly. Also rearranged the bullet points so that build related sections are together, docs are together, framework sections are together etc. If anyone has suggestions for improvements, please let me know :)


prigoyal:
cc @soumith , @hmansell

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+
+- name: copy
+  lang: |
+    def copy(float({dimParams}) I) -> (O) {{
+      O({dimIndices}) = I({dimIndices})
+    }}
+
+- name: cosine
+  lang: |
+    def cosine(float(M) I) -> (O) {
```
nicolasvasilache: this is one of the few that didn't exist in https://github.com/facebookresearch/TensorComprehensions/tree/master/include/tc/library
please consider opening a task so that we have parity between C++ and Python

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+      O(i) = cos(I(i))
+    }
+
+- name: cosine_similarity
+  lang: |
+    def cosine_similarity(float(M, N) I1, float(M, N) I2) -> (O, sumI1, sumI2) {{
```
nicolasvasilache: this is one of the few that didn't exist in https://github.com/facebookresearch/TensorComprehensions/tree/master/include/tc/library
please consider opening a task so that we have parity between C++ and Python

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+      sumI1(m) +=! I1(m, n) * I1(m, n)
+      sumI2(m) +=! I2(m, n) * I2(m, n)
+      O(m) +=! (I1(m, n) * I2(m, n)) / fmax(rsqrt(sumI1(m)) * sqrt(sumI2(m)), {eps})
+    }}
+
+- name: add
+  lang: |
+    def add(float(N) A, float(N) B) -> (output) {
+      output(i) = A(i) + B(i) + 1
```
nicolasvasilache: this seems a bit misleading? why add an extra 1?

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+
+- name: abs
+  lang: |
+    def abs(float(M, N) A) -> (O1) {
+      O1(m, n) = fabs(A(m, n))
+    }
+
+- name: layernorm
+  lang: |
+    def layernorm(float(T, B, C) I) -> (O, mean, centered, var)
```
nicolasvasilache: this is one of the few that didn't exist in https://github.com/facebookresearch/TensorComprehensions/tree/master/include/tc/library
please consider opening a task so that we have parity between C++ and Python

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+       mean(t, b) +=! I(t, b, c) / C
+       centered(t, b, c) = I(t, b, c) - mean(t, b)
+       var(t, b) +=! centered(t, b, c) * centered(t, b, c)
+       var(t, b) = (var(t, b) + {eps}) / C
+       O(t, b, c) = centered(t, b, c) / rsqrt(var(t, b))
+    }}
+
+- name: batchnorm
+  lang: |
+    def batchnorm(float(N,C,H,W) I, float(C) rMeanIn, float(C) rVarIn)
```
nicolasvasilache: this is one that we only had in unit tests but not in https://github.com/facebookresearch/TensorComprehensions/tree/master/include/tc/library
please consider opening a task so that we have parity between C++ and Python

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+       expectedVariance(c) +=! (variance(n, c, h, w) + {eps}) / (N * H * W)
+       rVarOut(c) = rsqrt(
+         (1 - {momentum}) * rVarIn(c) + {momentum} * expectedVariance(c))
+       O(n, c, h, w) = centered(n, c, h, w) * rVarOut(c)
+       normalizedOut(n, c, h, w) = O(n, c, h, w)
+    }}
+
+- name: small_mobilenet
+  lang: |
+    def small_mobilenet(float(C1, H, W) I, float(C1, KH1, KW1) W1, float(C1) B1, float(C2, C1) W2, float(C2) B2)
```
nicolasvasilache: this is one that we only had in unit tests but not in https://github.com/facebookresearch/TensorComprehensions/tree/master/include/tc/library
please consider opening a task so that we have parity between C++ and Python

tensor_comprehensions/pybinds/pybind_engine.cc
```diff
@@ -0,0 +1,133 @@
...
+          args.insert(args.begin(), "tc");
+        }
+        int numArgs = args.size();
+        // now we construct a char** argv type from args
+        std::vector<char*> vargs; // char* vector args
+        for (auto& arg : args) {
+          vargs.push_back(const_cast<char*>(arg.data()));
+        }
+        char** argv = vargs.data();
+        tc::python::globalDebugGflagsGlogInit(&numArgs, &argv);
```
nicolasvasilache: this function returns an unused ```bool```, what is the rationale here?

tensor_comprehensions/library/layers.yaml
prigoyal: I think there are many TCs here in python side that don't have test cases in C++. I have tried adding them in C++ side as much as possible. I'll double check and add all the TC in C++ as well but that'll happen in separate PR with proper task tracking. Thanks.

tensor_comprehensions/library/layers.yaml
prigoyal: just a simple example, can be removed or templated.

tensor_comprehensions/pybinds/pybind_engine.cc
prigoyal: this is how gflags/glog work. if the init fails at any point, we get the exception otherwise, it's important to catch True everywhere for successful init. There is InitGoogleTest and InitGoogleLogging and ParseCommandLineFlags that all should pass. If any of them fails, the init was problematic either due to bad flags, etc. 

For reference, this is also how things are in main in C++ tests.

tensor_comprehensions/library/layers.yaml
prigoyal: https://github.com/facebookresearch/TensorComprehensions/issues/74 issue filed for it

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -0,0 +1,120 @@
...
+            }
+            if (type == "conv") {
+              return tc::MappingOptions::makeConvolutionMappingOptions();
+            }
+            if (type == "group_conv") {
+              return tc::MappingOptions::makeGroupConvolutionMappingOptions();
+            }
+            throw std::runtime_error("Invalid option passed");
+          }),
+          "Initiailze the mapping options from one of the following:\n 1. naive\n 2. pointwise\n 3. mlp\n 4. conv\n 5. group_conv\n 6. single_thread")
```
nicolasvasilache: ```Initialize```

tensor_comprehensions/library/layers.yaml
prigoyal: I think the library referenced in above link is actually out-of-date. It has lots of fcrelu, LUT and matmul/convolution. Nothing else. 
Also, the library is not used anywhere except some C2 tests, we might consider cleaning up this library or start using is more properly and hence maintainig it.

tensor_comprehensions/tc_unit.py
```diff
@@ -0,0 +1,674 @@
...
+        options = get_options_from_cache_file(name, *inputs, **kwargs)
+    elif "options_cache" in kwargs and kwargs["options_cache"]:
+        options_cache = kwargs["options_cache"]
+        assert kwargs["type"] is not None, "layer type not specified: forward/backward"
+        options = options_cache[kwargs["type"]]
+        logger.info("Tuned kernel options found, using those options")
+
+    if options is None:
+        options = Options("naive")
+        logger.warning("No mapping options passed, 'naive' type mapping options will be used and will likely have bad performance. See help(your_layer.__call__) for setting mapping options.")
```
nicolasvasilache: As discussed, builtin options are not meant to be used by users but as a starting point for autotuning because this will give a false sense of the performance of what TC can achieve while only poking at the tip of the iceberg.

In any case, you also want to emit similar warnings for everything that wasn't explicitly autotuned (for now).
None of the builtin options are meant to be used to obtain "good" perf, they only arised from the core devs having played a bit by hand with the mapper and as relevant strategies for testing.

prigoyal: yes, for that reason, we explicitly mention "bad performance" in the warning message and then users can adapt. The warning message also gives instruction for how to set the options. The warning is thrown everywhere when the explicit options are not passed. The documentation will also provide clear instructions for passing options.

The reason for providing defaults and using defaults is that it allows quick experimentation without a learning curve of "Options". The current "naive" options work fine and don't fail compilation, at least for all the test cases provided here.

test_python/layers/test_external_cuda_injection.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+import torch.cuda
+import unittest
+
+
+class TestManualCudaInjection(unittest.TestCase):
+    def test_simple_cuda_injection(self):
+        lang = """
+        def add(float(N) A, float(N) B) -> (output) {
+            output(i) = A(i) + B(i) + 1
```
nicolasvasilache: I have the same comment as above, adding an extra ```1``` is surprising  and upon further inspection ```my_add``` below does not do it.
So you essentially have 2 implementations that don't do the same thing in the same test.
You inject the code in an unchecked way.

This test only passes because the output is not checked for correctness which is another thing I would like to see fixed.

Thanks!

prigoyal: @nicolasvasilache adding such checks for all the tests is not trivial, but there are checks in place, please look at `test_tc.py`

If you want to make contribution, please send a PR.

setup.py
```diff
@@ -0,0 +1,127 @@
...
+        tc_version += '.post' + str(tc_build_number)
+
+print('git_version: {} tc_version: {} tc_build_number: {}'.format(
+    git_version, tc_version, tc_build_number
+))
+
+################################################################################
+# Copy the proto files only
+################################################################################
+copyfile("src/proto/compcache.proto", "tensor_comprehensions/compilation_cache.proto")
```
nicolasvasilache: Could you please explain why this and the ```.gitignore``` change above are needed versus say just using the particular file?
If it is because we don't want to ship stuff from ```src``` then maybe it is appropriate to have a ```proto/*.proto``` outside of src and use that without copying? 
What do you think @ftynse @ttheodor ?

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,247 @@
...
+
+- name: tensordot
+  lang: |
+    def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
+      O(n, c1, c3, h, w) +=! I0(n, c1, c2, h, w) * I1(n, c2, c3, h, w)
+    }
+
+- name: fully_connected
+  lang: |
+    def fully_connected(float(B,M) I, float(N,M) W1, float(N) B1) -> (O1) {
```
nicolasvasilache: nit: spacings are inconsistent in the parameters in this def and a few others

conda_recipes/README.md
```diff
@@ -0,0 +1,145 @@
...
+**NOTE**: If you want to exit the conda environment, you can run
+
+```Shell
+source deactivate
+```
+
+5. Now, we have conda setup, we are ready to build conda package for TC. For this,
+since TC has its dependencies that are linked dynamically, we will build conda
+packages for them separately. In short as of January 31, 2018 we need to build
+packages for `clang+llvm-tapir5.0`, `gflags`, `glog`, `Halide`, TC version of `isl`, `PET`, TC version of `ppcg`, `protobuf`, and finally `Tensor Comprehensions`.
```
nicolasvasilache: PET / PPCG are outdated, please also update the date here.

tensor_comprehensions/library/layers.yaml
prigoyal: thanks for the nits, I'll correct them

setup.py
prigoyal: as part of the conda package, we don't ship the `src/proto` codepath or any of `src/`

hence, we just pick whatever is relevant and we only need the proto files.

Usage:

look at `tc_unit/decode` function. This is a functionality people wanted that once they have autotuned their options, the want to see what they look like. so now they can do

`tc.decode(cache_file)`

Reason for chosing name `decode` is because `protoc --decode` is the actual command and it was intuitive to use this name and user mentioned also "how do I decode the binary options file"?

conda_recipes/Dockerfile
```diff
@@ -0,0 +1,68 @@
+FROM nvidia/cuda:8.0-cudnn6-devel-ubuntu14.04
```
nicolasvasilache: FMI, why do we rewrite a full new Dockerfile rather than build on top of what we have been using in CI?

prigoyal: the one in CI doesn't have GPU but for building conda packages, I need to run python test which basically need GPU in all cases. Hence building the docker image which has GPUs. You can read how nvidia-docker works. I provided the readme instructions here for using and building this image.

tensor_comprehensions/library/layers.yaml
prigoyal: #74 issue filed for it

tensor_comprehensions/library/layers.yaml
prigoyal: #74 issue filed for it

conda_recipes/Dockerfile
prigoyal: #resolved

tensor_comprehensions/tc_unit.py
prigoyal: #resolved

tensor_comprehensions/pybinds/pybind_engine.cc
prigoyal: #resolved (please comment if some other clarity is needed). Thanks.

tensor_comprehensions/library/layers.yaml
prigoyal: #resolved

tensor_comprehensions/library/layers.yaml
prigoyal: #resolved

tensor_comprehensions/library/layers.yaml
prigoyal: #resolved

conda_recipes/Dockerfile
nicolasvasilache: The docker image can't have GPUs, you can't download them :)

All our trusty docker images used to inherit from this ```FROM nvidia/cuda:8.0-cudnn6-devel-ubuntu14.04``` and all our xenial docker images used to inherit from this ```FROM nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04``` but I now realize all that has changed drastically in the last stretch before the release and that we now use this [file](https://github.com/facebookresearch/TensorComprehensions/blob/master/docker/common/install_cuda.sh).

Still, when providing the Docker for conda_recipes you are using a very similar Docker base than what I wrote a few months back which got me confused. 

Does this mean we should uniformize what is in [here](https://github.com/facebookresearch/TensorComprehensions/tree/master/docker) and revert back to the old docker files like you did here?

I'll open a task specifically for this.

conda_recipes/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+RUN apt-get install -y --no-install-recommends libgtest-dev libz-dev libgmp3-dev
+RUN apt-get install -y --no-install-recommends automake libtool valgrind subversion
+RUN apt-get install -y --no-install-recommends ca-certificates software-properties-common
+
+RUN cmake --version
+
+# GCC 4.8.*
+RUN add-apt-repository ppa:ubuntu-toolchain-r/test
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends libcilkrts5 gcc-4.8 g++-4.8
```
nicolasvasilache: could you explain why it is required to install ```gcc-4.8``` manually? 
IIRC ubuntu 14.04 (trusty) should ship with ```gcc-4.8```.

prigoyal: It should but this is done as a safety measure.

conda_recipes/Dockerfile
prigoyal: The docker files will be moved from here to a separate repo https://github.com/prigoyal/tensorcomp-dockerfiles that I have setup for Jenkins setup. The question about whether the images should be tiered or not is something that I am discussing with @pietern and @ezyang already. Please open the issue in the above repo to centralize discussions properly.

>> Still, when providing the Docker for conda_recipes you are using a very similar Docker base than what I wrote a few months back which got me confused.

Not quite sure, the docker images follow quite different structure now and provide more safeguarding. 

>> FROM nvidia/cuda:8.0-cudnn6-devel-ubuntu14.04 

why did we inherit from those earlier? did you use nvidia-smi?

>> The docker image can't have GPUs, you can't download them :)

yeah, that's why we use `nvidia-docker` https://github.com/NVIDIA/nvidia-docker , looking at the diagram here will make things clear.

prigoyal: #resolved

conda_recipes/Dockerfile
prigoyal: #resolved

conda_recipes/Dockerfile
nicolasvasilache: > why did we inherit from those earlier? did you use nvidia-smi?

No since there is no cuda ***driver*** on the CPU-only CI machines.
However we have always had cross-compilation tests on CPU-only CI that were testing simple parts of the CUDA ***toolkit***. 

In particular you don't need a physical GPU to generate PTX and see whether compilation passes (e.g. NVRTC + CUB)...

prigoyal: you can continue doing that, no need to start from NVIDIA images

conda_recipes/README.md
```diff
@@ -0,0 +1,144 @@
+# How to build conda package for Tensor Comprehensions and its dependencies
```
nicolasvasilache: Cool, nice doc :)

conda_recipes/caffe2/build.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
nicolasvasilache: So now we have the main ```build.sh``` and these specific ```build.sh``` for different conda targets.

In the same spirit of avoiding copy-pasta, let's please restructure things such that the main ```build.sh``` uses those. This is clearly a outside of the scope of this PR and we can do it in a later.

prigoyal: conda packaging env variables are different. Each package needs to be independent enough to build. I can explain to you in-person how conda packaging works, that might be higher bandwidth and will help answer this question better. 

also, I believe you wanted to remove the main build.sh as well. is that still your goal?

nicolasvasilache: opened #78

prigoyal: closed and responded! thanks!

tensor_comprehensions/tc_unit.py
```diff
@@ -0,0 +1,674 @@
...
+        if "options" in kwargs:
+            orig_options = kwargs["options"]
+            kwargs["options"] = forward_best_options
+            outputs = cu.compile_and_run(name, input_tensors, **kwargs)
+            kwargs["options"] = orig_options
+        else:
+            outputs = cu.compile_and_run(name, input_tensors, options=forward_best_options, **kwargs)
+        # now that we have the outputs of the forward pass, we have the inputs
+        # for the backward layer and we can now tune the backward layer
+        reorder_function = kwargs["reorder_function"] if "reorder_function" in kwargs else None
```
nicolasvasilache: Unless I am missing it in this large PR I do not see a test that uses a custom reorder_function.
Please add one such example. 
Anything simple that produces a temporary as part of the outputs is fine.

tensor_comprehensions/tc_unit.py
```diff
@@ -0,0 +1,674 @@
...
+
+    def __init__(self, lang, **kwargs_define):
+        self.cu = TcCompilationUnit()
+        self.cu.define(lang)
+        self.kwargs_define = kwargs_define
+        self.lang = lang
+        self.tuner = None   # this tuner maintains a cache for kernels/input sizes tuned so far
+
+    def __call__(self, *inputs, **kwargs):
+        r"""Runs the define TC language on given inputs.
```
nicolasvasilache: ```defined```

tensor_comprehensions/tc_unit.py
```diff
@@ -0,0 +1,674 @@
...
+            name, input_tensors, mapping_options=options, cache_file=cache_file
+        )
+        # update the cache with the options
+        options_cache["forward"] = forward_best_options
+        if not backward:
+            self.tuner_cache[hash_key] = options_cache
+            return forward_best_options
+
+        # now, we have to tune the backward layer, for that, we need to run
+        # the forward layer first, get it's output,
```
nicolasvasilache: ```its```

tensor_comprehensions/tc_unit.py
```diff
@@ -0,0 +1,674 @@
...
+        except RuntimeError:
+            return options
+
+    def autotune(self, *inputs, **kwargs):
+        input_tensors = get_tensors(list(inputs))
+        kwargs.update(self.kwargs)
+        name, backward_name = get_tc_names_from_kwargs(**kwargs)
+        kwargs.pop("name", None)
+        backward = True if backward_name is not None else False
+        hash_key = get_tc_hash_key(name, *input_tensors)
```
nicolasvasilache: So with this you are actually signing up for trouble as there is no guarantee that what you cache is correct in the long run. 

This should work fine for what you are currently doing with it but you should plan to pybind export functions (that we will need to expose) from [here](https://github.com/facebookresearch/TensorComprehensions/blob/master/include/tc/core/compilation_cache.h).

The reason is that the caching logic will depend on output tensor sizes, input/output tensor striding, allocation alignment many other fun things as we start digging deeper into optimizations.

There is nothing to change for now but it is just a friendly warning that you should not rely too much on your own framework-specific caching mechanism since this seems quite central to your autotuner wrapper logic.

tensor_comprehensions/tc_unit.py
```diff
@@ -0,0 +1,674 @@
...
+        self.cu = TcCompilationUnit()
+        self.cu.define(lang)
+        self.kwargs_define = kwargs_define
+        self.lang = lang
+        self.tuner = None   # this tuner maintains a cache for kernels/input sizes tuned so far
+
+    def __call__(self, *inputs, **kwargs):
+        r"""Runs the define TC language on given inputs.
+
+        Args:
```
nicolasvasilache: nit: there seems to be a lot of duplicate between the ```__call__``` and the ```autotune``` comments, is there a way to compress a bit?

nicolasvasilache:
Nice and sweet, python API is a very lightweight wrapper around the core C++ API so that will be as future-proof as it can. I like it.

There is too much python side handling of caching logic to my taste but we haven't provided you with the proper C++ functions to do better for now; so we should iterate to avoid you building more logic that some day will break.

There are still concerns about:
1. silent default options, however loud the warning is
2. python TC string definitions that are different from the C++ string definitions 
3. python test that do not test correctness

3. is a blocker
2. should be a blocker but I can make my peace with it
1. should disappear completely in a subsequent PR, only risk is that we will disseminate non-best-practice code for a bit, not a blocker in my opinion but it may raise some unhappiness on the python side in the future

prigoyal:
>> 1. silent default options, however loud the warning is

**NOT BLOCKER** User are happy and desire this behavior, they don't want the learning curve of "Options" so it's good if we all make our peace with it. 

>> 2. python TC string definitions that are different from the C++ string definitions

 **NOT BLOCKER** We already discussed it in much details. The correct fix is to allow handling scalars for bounds inference (task open for this). In them meantime, people can adopt one of the two ways of formatting below 

**OPTION 1**
```
import tensor_comprehensions as tc
import torch
lang = """
def avgpool(float(B, C, H, W) input) -> (output) {{
    output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
}}
"""
avgpool = tc.define(lang, name="avgpool", constants={"sH":1, "sW":1, "kH":2, "kW":2})
inp = torch.ones(32, 3, 10, 10).cuda()
out = avgpool(inp)
```

**OPTION 2**
```
import tensor_comprehensions as tc
import torch
import re
LANG="""
def avgpool(float(B, C, H, W) input) -> (output) {
    output(b, c, h, w) += input(b, c, h * <sh> + kh, w * <sw> + kw) where kh in 0:<kH>, kw in 0:<kW>
}
"""
sH, sW, kH, kW = 1, 1, 2, 2
LANG = re.sub('<sh>', str(sH), LANG)
LANG = re.sub('<sw>', str(sW), LANG)
LANG = re.sub('<kH>', str(kH), LANG)
LANG = re.sub('<kW>', str(kW), LANG)
avgpool = tc.define(LANG, name="avgpool")
inp = torch.ones(1, 1, 4, 4).cuda()
out = avgpool(inp)
```

OR any other option people can come up with

>> 3. python test that do not test correctness

**NOT BLOCKER** most of the python tests are backed by C++ backend. I will be completing the checks on python front over next days since a proper python test harness is needed. However, this is not a release blocker: I have myself verifieed running torch functions and comparing tc output on my command line + some users have also verified their outputs are correct.


NONE of the above concerns should be a blocker. THey rather seem to be enhancements to me. The python packages are ready. Things will never be perfect but we are properly documenting what works and what doesn't work.

tensor_comprehensions/tc_unit.py
prigoyal: there is no support for strided tensors at the moment. When that is added and the task is closed for that, the python API will be modified accordingly.

tensor_comprehensions/README.md
```diff
@@ -0,0 +1,63 @@
...
+}
+"""
+matmul = tc.define(lang, name="matmul")
+mat1, mat2 = torch.randn(3, 4).cuda(), torch.randn(4, 5).cuda()
+out = matmul(mat1, mat2)
+```
+
+## Communication
+
+* **Email**: prigoyal@fb.com
```
nicolasvasilache: ```tensorcomp@fb.com```

prigoyal: this section is just not needed, I'll remove it, the communication will go on the README page

prigoyal:
@ftynse I had a quick sync with @nicolasvasilache just now about all the points above. We have agreed that the the 3rd point is a nice to have thing and I'll work on it after the release or as I find time over this week. We have also agreed that I'll wait until midnight for your review (midnight = EST midnight), since you are at conference until Wed and the release has been planned for Wed for few weeks, so it's fine that I merge this PR at midnight. You can still leave your review comments and I'll try to address them in separate PRs.

tensor_comprehensions/tc_unit.py
prigoyal: added the example `test_python/layers/test_convolution_reorder.py`

prigoyal:
sure, that's just a matter of whether we want to make the extra call for substitute or not. TC define call can do that substitution and takes the dictionary of arguments. From the feedback from @soumith about using the constants dictionary. I think this is largely empirical and I'll add an option to also do the `substitute` as well and expose a test case, users can chose whatever makes the most sense to them.

how about this:

```
lang = """
def avgpool(float(B, C, H, W) input) -> (output) {
    output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
}
"""
avgpool = tc.define(lang, substitute={sh=1, ...})
```

conda_recipes/isl-tc/meta.yaml
```diff
@@ -0,0 +1,35 @@
+package:
+  name: isl-tc
```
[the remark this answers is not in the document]
prigoyal: to make it specific to ISL, I think there was a comment long back when I was writing these scripts that we should make things as tc like ppcg-tc, etc so that we don't confuse with the other actual repos

src/core/flags.cc
```diff
@@ -110,6 +110,34 @@ DEFINE_int64(
     -1,
     "The number of best candidates to restore from the proto cache");
 
+namespace {
+bool parseCommandLineFlags(int* pargc, char*** pargv) {
+  if (*pargc == 0) {
+    return true;
+  }
+  // TODO: (prigoyal): we need to do some filtering on flags here,
```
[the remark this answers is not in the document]
prigoyal: will do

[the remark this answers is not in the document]
prigoyal: yeah, regarding `proto/` we only ship them in `conda` and not look at `src/`

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -0,0 +1,120 @@
...
+          [](tc::MappingOptions& instance, const std::string& type) {
+            instance.scheduleFusionStrategy(type);
+          },
+          "Set up outerScheduleFusionStrategy and intraTileFusionStrategy to the given value")
+      .def(
+          "outerScheduleFusionStrategy",
+          [](tc::MappingOptions& instance, const std::string& type) {
+            instance.outerScheduleFusionStrategy(type);
+          },
+          "Require TC to try and execute different TC expressions interleaved (Max), separately (Min)\nor interleaved as long as sufficient parallelism is exploited (Preserve3Coincident) by\nperforming loop fusion and fission. Applies before tiling")
```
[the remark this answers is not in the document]
prigoyal: will correct this

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -0,0 +1,120 @@
...
+          &tc::MappingOptions::unroll,
+          "Perform loop unrolling on the generated code and produce at most the given number of statements")
+      .def(
+          "scheduleFusionStrategy",
+          [](tc::MappingOptions& instance, const std::string& type) {
+            instance.scheduleFusionStrategy(type);
+          },
+          "Set up outerScheduleFusionStrategy and intraTileFusionStrategy to the given value")
+      .def(
+          "outerScheduleFusionStrategy",
```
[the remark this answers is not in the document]
prigoyal: will do

tensor_comprehensions/torch_tc/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2017-present, Facebook, Inc.
```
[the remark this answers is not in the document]
prigoyal: this is how python `packaging works`

CMakeLists.txt
```diff
@@ -174,8 +174,6 @@ endif()
 
 ################################################################################
 # ATen
-
-# ATen - if someone ships libATen.so,  we try to use that if available
```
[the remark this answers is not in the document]
prigoyal: we do but this has been more properly highlighted in next steps when we are searching for libaten in pytorch install

tensor_comprehensions/tc_unit.py
prigoyal: LOL, I'll add the section about polyhedral/halide to the python doc strings to the function. 

>> Copy-pasting code from stack overflow without a slightest clue of how it works is like half of software engineering.

yep but our users are also researchers and not engineers :)

setup.py
prigoyal: yeah, regarding `proto/` we only ship them in conda and not look at `src/`

prigoyal: I'll mark this as resolved, current shipping of proto seems fine and the decode naming is okay too

tensor_comprehensions/library/layers.yaml
prigoyal: agreed, I have a task open for bringing the test case parity between C++ and Python TC, I think what i would like to do is to have a properly library that is maintained and updated.

setup.py
```diff
@@ -0,0 +1,127 @@
...
+###############################################################################
+# Version and build number
+###############################################################################
+
+try:
+    git_version = subprocess.check_output(
+        ['git', 'rev-parse', 'HEAD'], cwd=TOP_DIR
+    ).decode('ascii').strip()
+except (OSError, subprocess.CalledProcessError):
+    git_version = None
```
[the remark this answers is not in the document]
prigoyal: yeah, I'll clean this up though, we might prefer to rather fail

ftynse:
I'm totally willing to pay the price of one extra explicit call to get separation of concerns: TC language is what is defined by the BNF in the documentation, substitution is done separately because TC language does not support it.

jekbradbury:
Why can't substitution rely on the native Python `template` module or `<string>.format`?

prigoyal:
@jekbradbury the right fix is really to be able to infer the bounds in `where` clauses when scalars are passed. Then we can write better TC like this:

```
def avgpool(float(B, C, H, W) input, float kH, float kW) -> (output) {
    output(b, c, h, w) += input(b, c, h + kh, w + kw) where kh in 0:kH, kw in 0:kW
}
```
but this isn't easy to fix atm due to how our computation model is right now, how we allocate tensors and scope we compile things in.

A couple of things we wanted to make sure in this API:
1. Keep the API minimal and low surface. Right now there are basically only three calls 
`tc.define`, `my_awesome_layer()`, `my_layer.autotune()`

2. We want to keep the parity between what an actual TC syntax is i.e. `def name () { code }` and that TC syntax should stick in Python as well. 

We can use any kind of substitution: the python `template` module will work just fine, user can do that, or you can also do `re.sub` or you can use what we provide i.e. pass `constants` to `tc.define` call. anything is fine and okay and we leave it to use to do whatever makes the most sense.

jekbradbury:
So this would work in Py3.6 right now?

lang = f"""
def avgpool(float(B, C, H, W) input) -> (output) {
    output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
}
"""
avgpool = tc.define(lang)

prigoyal:
@jekbradbury you would still have to do the following:

```
>>> sH = 1
>>> sW = 1
>>> kH = 2
>>> kW = 2
>>> lang = f"""
... def avgpool(float(B, C, H, W) input) -> (output) {{
... output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
... }}
... """
```
i.e. still use the `{{...}}` but remove the constants, or substitute call. The reason for still using `{{...}}` is that TC string parser doesn't interact well with the python string formatter cc @zdevito 

However, this is the thing that will stick around for ~1 month and is a temporary workaround for now. 

Although using the `f` is nice, I wouldn't want to want put any python3.6 specific features in API. we officially support py3.6 but py2 might as well just work and we haven't tested it and don't support it officially.

jekbradbury:
Ohhh of course, that clash of curly-brace semantics is annoying but seems kind of inevitable.

prigoyal:
yep, for now, it is :) but we are working on it and will put a giant WARNING / NOTE that this is temporary solution and will be away in 1 month or so :)

tensor_comprehensions/tc_unit.py
prigoyal: the python readme now has the polyhedral/halide part as well. That was purely an oversight on my part, apologies and correction made now :)

nicolasvasilache:
Re. fake templating by string replacement I strictly prefer the ```<...>``` syntax with proper factory functions exposed via pybind. The main reason is that this is what we have been using for a while in places like [this](https://github.com/facebookresearch/TensorComprehensions/blob/master/include/tc/library/group_convolution.h#L51).
However I understand the extra constraints on @prigoyal and we can always fix it later but as @ftynse  also expressed:  "If they are copied, they will get out of sync faster than you can imagine".
This is the general reason why copy-pasta represents such a big technical debt.

> Why can't substitution rely on the native Python template module or <string>.format

@jekbradbury reg. the above, any logic implemented outside of C++ that people risk starting to rely upon is a liability (do I really need Python when I start optimizing for older Android phones). See also @ftynse's comment about what TC is and isn't. Additionally, this is an opportunity to reintroduce all the  copy-pasta issues. 

This is likely a symptom not yet having found the proper abstraction to express such things, whether actual templates in the language, a proper TcBuilder API or something else that we would love other's feedback on. The immediate consequence is that whatever is implemented atm for this particular point is subject to change.

tensor_comprehensions/README.md
```diff
@@ -0,0 +1,63 @@
+# ![Tensor Comprehensions](docs/source/_static/img/tc-logo-full-color-with-text-2.png)
+
+# Using Tensor Comprehensions with PyTorch
+
+A blogpost on Tensor Comprehensions can be read [here](https://research.fb.com/announcing-tensor-comprehensions/).
+
+We provide integration of Tensor Comprehensions (TC) with PyTorch for both training
+and inference purposes. Using TC, you can express an operator using [Einstein
+notation](https://obilaniu6266h16.wordpress.com/2016/02/04/einstein-summation-in-numpy/)
```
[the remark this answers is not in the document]
prigoyal: good catch, removing it

tensor_comprehensions/library/layers.yaml
```diff
@@ -0,0 +1,245 @@
...
+# Purpose:
+# This module serves as a database for Tensor Comprehensions representation of layers
+# that are useful for the Machine Learning community. This database should be
+# extended with more layers including fusions, compositions etc.
+#
+# To extend this:
+# add a new entry which at least has the "name" and the "lang". You can chose to
+# add other entries like "grad".
+# For the gradient layers, name should be: name + "_grad"
+
```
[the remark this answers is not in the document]
prigoyal: this is a good question, I wanted to expose something like 

```
tc.database
```

to people so they can get the TC string for some operations defined here. Soon, I think this shall be moved to the header file

prigoyal:
>> I strictly prefer the <...> syntax with proper factory 

I agree and I would still say that the proper solution is to have bounds inference from scalar working and then we can avoid all sorts of custom bindings .

>> However I understand the extra constraints on @prigoyal and we can always fix it later but as @ftynse also expressed: "If they are copied, they will get out of sync faster than you can imagine".

There aren't any constraints to incur big technical debts. @ftynse is saying that we need to build the better header library in C++ backend which doesn't exist right now. Rather the TC have been defined in python side in the library file. 

>> This is the general reason why copy-pasta represents such a big technical debt.

@nicolasvasilache can you point out where is the copy-pasta in this? I would like to clean that up :)

>> (do I really need Python when I start optimizing for older Android phones)

no you don't, this will be addressed when the TC defs are moved to the library header (which TC codebase itself doesn't respect atm)

prigoyal:
landing this now, CI is green, all comments from @ftynse have been addressed as well.

Acknowledging two things at the time of landing this:
1. Issue 74 should become a higher priority. TC codebase doesn't respect the include/library itself and it's not a fair expectation to block the release/PR to make that nice in codebase now. That's an effort that requires proper thinking also thinking about long term pytorch JIT plans we have.

2. Loud and clear warnings about `{{...}}` syntax. For now, @ftynse has given consent that we can keep it as is. Issue #73  needs work to make this proper.

tensor_comprehensions/library/layers.yaml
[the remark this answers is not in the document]
ftynse: :+1:  we have some earlier attempts in `include/tc/library`, we need to consolidate that and make it simple to use

ftynse:
> Why can't substitution rely on the native Python template module or <string>.format?

Anything related to the TC language should be (eventually) done in the TC parser, and absolutely cannot rely on anything python-specific.

ftynse:
> i.e. still use the {{...}} but remove the constants, or substitute call. The reason for still using {{...}} is that TC string parser doesn't interact well with the python string formatter

>  that clash of curly-brace semantics is annoying but seems kind of inevitable.

I don't think TC parser should have anything special to support whatever upstream string manipulation.  We should think and choose a proper syntax for template-style substitutions, that's all.

prigoyal:
>> But I am positive that internally the substitution in define does rely on <string>.format but does so implicitly.

yes, that's correct :) it's one line for `.format()` if you look at the `define` call. that `define` basically does what you said: "you can apply whatever string operations to it before you call `define`"

>> I don't think TC parser should have anything special to support whatever upstream string manipulation.

I meant that if you do just one braces like this:
```
def avgpool(float(B, C, H, W) input) -> (output) {
   output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
}
```

and try to format it, it won't work and the reason is the `{..}` in the TC parser. It doesn't know that there are two `{{` now and parses things incorrectly.

Issue #74, opened 2018-02-26
Add C++ test for all TC from layers database


prigoyal:
there might be few missing TC defs in C++ land, they probably only exist in python side. I'll add C++ test case for all of them.

ftynse:
layers database is dead

PR #75, opened 2018-02-26
Documentation for PyTorch Integration with Tensor Comprehensions

prigoyal:
Adding the documentation for various components of PyTorch package.

Everything is covered from Getting Started to writing layers, autograd with TC, autotuning layers etc.

**NOTE**: PR is stacked on the #72 , please land the PRs sequentially or ask me to change the base of PR once #72 lands.


prigoyal:
cc @soumith

docs/source/framework/pytorch_integration/layers_database.rst
```diff
@@ -0,0 +1,364 @@
+ML Layers database
```
nicolasvasilache: Can we delete this file?

The main reason I am asking is because the list of layers mostly comes from C++ land with slight variations and risks making this the source of truth for PyTorch users. Making this the source of truth comes with all the issues mentioned previously by both @ftynse  and myself:
1. copy-pasta synchronization issues (we now have many of these layers in 3 places)
2. not the actual source of truth since the source of truth is always C++ and people's programs will break at some unexpected point in the future
3. the syntax is not TC but makes the ```{...}``` templating by string replace hack a first class citizen of the language

docs/source/framework/pytorch_integration/layers_database.rst
```diff
@@ -0,0 +1,364 @@
...
+
+    def cosine_similarity(float(M, N) I1, float(M, N) I2) -> (O, sumI1, sumI2) {{
+        sumI1(m) +=! I1(m, n) * I1(m, n)
+        sumI2(m) +=! I2(m, n) * I2(m, n)
+        O(m) +=! (I1(m, n) * I2(m, n)) / fmax(rsqrt(sumI1(m)) * sqrt(sumI2(m)), {eps})
+    }}
+
+What operations can not be expressed
+------------------------------------
+* **Reshaping** Tensors inside the language
```
nicolasvasilache: For reshaping there are most often workarounds  in pure TC that make the intent of the program much better, examples can be added in the future as I think this file should go for the reason exposed above.

docs/source/framework/pytorch_integration/writing_layers.rst
```diff
@@ -0,0 +1,280 @@
...
+need to pass the name of the kernel and the CUDA code string to the :code:`tc.define`
+call. For example:
+
+.. code-block:: python
+
+    import tensor_comprehensions as tc
+    import torch
+    lang = """
+    def add(float(N) A, float(N) B) -> (output) {
+        output(i) = A(i) + B(i) + 1
```
nicolasvasilache: same side effect of copy-pasta as in #72 : your add function adds an extra 1 that your my_add function does not

prigoyal:
thanks for the review @nicolasvasilache 

I don't think dropping below is a good idea: 
```
docs/source/framework/pytorch_integration/layers_database.rst
```

the intent for that is that when users read the documentation, they can easily see all the layers in one place and understand things.

docs/source/framework/pytorch_integration/writing_layers.rst
prigoyal: you are reviewing a stacked PR, this is part of base PR #72

docs/source/framework/pytorch_integration/layers_database.rst
prigoyal: IMO we should keep this file. As a user myself, I find it very convenient to be able to look at all the layers together. Rather than removing file, we need to do something better. THe issue I opened around creating a library for TC defs will help bringing structure to how we use and define and where we keep the TC defs. But that will happen later after the release.

docs/source/framework/pytorch_integration/layers_database.rst
prigoyal: nope, we can't delete this. It's pretty important and useful doc entry. 

Let's wait out for the Issue #74  to be resolved and then we can circulate at it.

prigoyal:
screen shot 2018-02-26 at 11 52 06 pm

this is how the layers doc looks like, it's pretty useful to look at.

prigoyal:
CI is all green, I'll generate the html docs for people to look at and iterate. Whatever feedback I got so far has been addressed.

docs/source/framework/pytorch_integration/layers_database.rst
ftynse: Ideally, this file should be auto-generated from headers with actual TC definitions, to avoid duplication and decoherence issues.

prigoyal: yep, exactly and that requires some thinking, discussion and work

PR #81, opened 2018-02-27
Correctly check the autotuner cache file exists

prigoyal:
1. I somehow missed the proper checking of file earlier, corrected now
2. do init when autotuner is called. The autotuner might be called with different parameter settings like generations=10 etc so the init call is needed to update and run autotuning with new settings :)


prigoyal:
thanks @zdevito :) merging this now as I need to work on conda packages asap, they take time to build

PR #90, opened 2018-03-01
Paths and build system cleanups

nicolasvasilache:
WIP with @salexspb towards build system cleanups to drastically simplify integration in other build systems.

This first changeset moves all the files required to build tc_core_cuda under core/cuda and updates include paths.
This is pure build paths, future refactoring of core behavior out of cuda files is outside the scope of this PR.


prigoyal:
we probably should modify title to avoid having facebook specific things :)

src/CMakeLists.txt
```diff
@@ -1,16 +1,16 @@
 include_directories(.)
 include_directories(../include)
 
+add_subdirectory(autotuner)
+add_subdirectory(core)
```
prigoyal: core depends on version, lang, proto so we might want to move this line after add_subdirectory(version). We should prioritize the order of build execution rather than naming :)

nicolasvasilache: So I'd claim that the ordering of dependencies is a lexicographic order that cmake solves so order of subdirs shouldn't matter.
The issue I had that triggered this is interleaving of WITH_CUDA and non WITH_CUDA behavior which made me hoist ```autotuner``` before ```aten```; I'm not sure I'd want to change that.

In the light of the above do you still feel autotuner should go back at then end?

I'm mostly worried about interleaving of these things when we also add opencl and more behaviors. 
It will then look like spaghetti interleaving but I can see that this is premature optimization at this point :)

prigoyal: oh I meant core actually :) autotuner is okay

nicolasvasilache: Ok so then I'll upate with this lexicographic rule:
1. no spaghetti options
2. order of dependencies rather than dictionary order

Thanks!

build.sh
```diff
@@ -411,8 +411,6 @@ function install_halide() {
       WITH_METAL= \
       WITH_EXCEPTIONS=1 \
       make -f ../Makefile -j $CORES install || exit 1
-      mkdir -p ${INSTALL_PREFIX}/include/Halide
-      mv ${INSTALL_PREFIX}/include/Halide*.h  ${INSTALL_PREFIX}/include/Halide/
```
skimo-openhub: You missed the `-d "${INSTALL_PREFIX}/include/Halide"` a couple of lines up.
This should check for the `Halide.h` _file_ now.

prigoyal:
@salexspb just out of curiosity, why can't fbcode import "Halide/Halide.h"? are you getting halide from tp2 and that's the reason? thanks! :)

build.sh
nicolasvasilache: thanks for catchnig it, fixed

salexspb:
@prigoyal , yes this the reason.

src/core/CMakeLists.txt
```diff
@@ -94,6 +53,36 @@ install(
...
+  find_library(CUDA_NVRTC_LIBRARIES nvrtc
+    PATHS ${CUDA_TOOLKIT_ROOT_DIR}
+    PATH_SUFFIXES lib lib64 targets/x86_64-linux/lib targets/x86_64-linux/lib/stubs)
+
+  add_library(
+    tc_core_cuda
+
+    SHARED
+
+    cuda/cuda.cc
```
salexspb: how about using a regexp ? :) Then we will make sure things work in a similar way with both buck and cmake

salexspb:
Merged this so I can already rebase my buck work. But still check out a comment :)

PR #91, opened 2018-03-01
Untiered docker images for easy Jenkins setup

### LINKED ISSUE #79: Docker images improvement - non tiered version
We need 4 docker files only:

gcc 4.8 - cuda8-cudnn6 - conda
gcc 4.8 - cuda8-cudnn6 - non conda

gcc 5 - cuda9-cudnn7 - conda
gcc 5 - cuda9-cudnn7 - non conda

as per discussion with @pietern and @ezyang, these images can be based off of nvidia docker images

prigoyal:
moving the docker images to base off of nvidia docker images and make them non-tiered

closes #79


nicolasvasilache:
So I am still having issues understanding why we need to explicitly install the compiler that IIRC comes by default with the particular distribution (```gcc-4.8``` for ```trusty``` and ```gcc-5``` for ```xenial```).

My current understanding is that there were libstdc++ version / linking issues when mixing this with conda. Is that accurate?

I think it's ok to punt on the investigation if it requires non-trivial extra work.
But if you have more details already, I'd still like to understand what the problem is.

prigoyal:
>> So I am still having issues understanding why we need to explicitly install the compiler that IIRC comes by default with the particular distribution (gcc-4.8 for trusty and gcc-5 for xenial).

this is a safety measure, if someone wants to play with the trusty image but use gcc5.4 let's say then they will know that the image corresponds to what gcc version **explicitly** and things won't silently fail. Conda also ships its own gcc and libstdc++ and given the light of this, I find it better to be explicit. it affects protobuf as well.

nothing more to investigate on this :)

nicolasvasilache:
got it, explicit trumps implicit, thanks!

prigoyal:
thanks @zdevito :)

PR #93, opened 2018-03-02
Bump isl for move of bindings

skimo-openhub:
The bindings have moved to a different location.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,7 +2,7 @@
 [submodule "third-party/islpp"]
 	path = third-party/islpp
 	url = https://github.com/nicolasvasilache/isl.git
-	branch = ntv_dev
+	branch = ntv_dev_cpp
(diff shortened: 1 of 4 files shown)


prigoyal:
1. thanks @skimo-openhub , can you add some description to the issue? it's not clear what move of bindings means form the title.

2. this change requires also changing the  isl-tc conda package since that is part of the installation instruction https://facebookresearch.github.io/TensorComprehensions/installation_conda_dep.html#step-5-install-tc. so let me bump up that for you. This might also require me to maintain a repo with packages for building from source. I;ll do so today and get this PR landed :)

ftynse:
@prigoyal  this goes to dev, not master, I don't think you should repackage conda stuff with every isl change

prigoyal:
whoops, didn't see that it was sent to dev, my apologies

PR #100, opened 2018-03-05
Make few things explicit - package distr and cuda() for tensors

prigoyal:
1. make it explicit that conda package is linux-64, some people tried installing on osx-64
2. add a note about .cuda() placement. This is jankiness and suprising thing in pytorch but the next pytorch releases will have this corrected and better supported

### DIFF
--- a/docs/source/framework/pytorch_integration/autograd_with_tc.rst
+++ b/docs/source/framework/pytorch_integration/autograd_with_tc.rst
@@ -41,6 +41,13 @@ Examples
      out = convolution(I, W)
      out[0].sum().backward()
 
+.. note::
+
+    Please note the usage of :code:`.cuda()` i.e. tensor data is declared as the CUDA
+    type. Applying :code:`Variable` on the tensor data essentially allows the layer to be
+    part of computations graph and if :code:`Variable(torch.rand(), requires_grad=True).cuda()`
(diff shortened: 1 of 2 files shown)


docs/source/framework/pytorch_integration/autograd_with_tc.rst
```diff
@@ -41,6 +41,13 @@ Examples
      out = convolution(I, W)
      out[0].sum().backward()
 
+.. note::
+
+    Please note the usage of :code:`.cuda()` i.e. tensor data is declared as the CUDA
+    type. Applying :code:`Variable` on the tensor data essentially is an extra no-op and
+    if :code:`Variable(torch.rand(), requires_grad=True).cuda()` is done, then the
+    grad will be available for the no-op and not the actual tensor data.
```
ftynse: I got lost here.  Why do we need that no-op in the first place?  And if it is a no-op, why does it matter if it receives grad or tensor data?

prigoyal: however, the usage of Variable is going away in next pytorch release. So we have to live with it only few more days until next pytorch release happens

ftynse: That part is clear, mentioning no-op is not...

prigoyal:
thanks @wsmoses :)

PR #102, opened 2018-03-06
Cleanups towards CPU mapper

nicolasvasilache:
This PR implements various cleanups which are part of the process of getting a proper CPU mapper:
1. bunch of cleanups
2. adds the ability to call the LLVM LLC and dump assembly when ```--llvm_dump_asm=1```
3. drops (Cuda)ExecutionInfo which is just pure technical debt


include/tc/core/execution_engine.h
```diff
@@ -106,7 +106,7 @@ class ExecutionEngine {
   size_t emplaceExecutor(std::unique_ptr<ExecutorInfo> p);
 
   /// For thread-safety perform all cheap operations under lock.
-  std::mutex executorInfoMutex;
+  std::mutex executorInfoMutex_;
```
ftynse: should we update the coding conventions to say "private _and protected_ members" should be suffixed with underscore https://github.com/facebookresearch/TensorComprehensions/blob/master/CodingConventions.md#class-data-members ?

include/tc/core/execution_engine.h
```diff
@@ -105,6 +106,11 @@ class ExecutionEngine {
  protected:
   size_t emplaceExecutor(std::unique_ptr<ExecutorInfo> p);
 
+  size_t getHandle(
+      const std::string& name,
+      const std::vector<const DLTensor*>& inputsInfo,
+      const MappingOptions& options);
```
ftynse: `MappingOptions` are CUDA-specific (have grid, block sizes and most configs are GPU-related).  I used a string for options in the parent classes...

src/core/flags.cc
```diff
@@ -40,6 +40,10 @@ DEFINE_bool(dump_cuda, false, "Print the generated cudaSource");
 // CPU codegen options
 DEFINE_bool(llvm_dump_before_opt, false, "Print IR before optimization");
 DEFINE_bool(llvm_dump_after_opt, false, "Print IR after optimization");
+DEFINE_bool(
+    llvm_dump_asm,
+    true,
+    "Launches the LLC in a separate process and prints the resulting asm");
```
skimo-openhub: s/the LLC/LLC/

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
           << "[LLVM-IR] After optimization:\n"
           << toString(module.get());
     }
+
+    // In a separate process call the LLC and dump the ASM
```
skimo-openhub: add comma before "call"

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -23,20 +28,30 @@
...
+      const MappingOptions& options,
+      lang::TreeRef tcDefinition,
+      size_t handle)
+      : TcExecutor(
+            id,
+            inputsInfo,
+            options.toProtobufSerializedString(),
+            tcDefinition,
+            handle) {}
+  // CudaTcExecutor(
```
ftynse: commented-out code

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+      char buffer[512];
+      std::stringstream outputss;
+      while (1) {
+        int charactersRead = fread(buffer, 1, sizeof(buffer), fpipe);
+        if (charactersRead == 0) {
+          break;
+        }
+        outputss << std::string(buffer, charactersRead);
+      }
```
skimo-openhub: The generation of asm string should be extracted out into a separate function,
with some helper functions, e.g., for slurping in a FILE.

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+      // Create a temporary file (do not use racy tmpname;
+      // do not use tmpnam that immediately destroys the file).
+      char fileCName[256] = "/tmp/XXXXXXXXXX";
+      auto fd = mkstemp(fileCName);
+      ScopeGuard sg([fileCName]() { unlink(fileCName); });
+      CHECK_GE(fd, 0) << "mkstemp(): " << fileCName << " returned fd: " << fd;
+      close(fd);
+
+      // Close and reopen as FILE*
+      auto tmp = fopen(fileCName, "wb+");
```
skimo-openhub: If you're worried about races, you should use `fdopen` instead of closing and
reopening.

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+      auto llvmIr = toString(module.get());
+      std::fputs(llvmIr.c_str(), tmp);
+      fclose(tmp); // Close properly so the piped command can open without race
```
skimo-openhub: What's the purpose of this comment?
You should _always_ close a FILE properly.

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+      std::string fileName(fileCName);
+      // TODO: evolve this in a better way to find LLVM
+      auto command = std::string(" `which llc || which llc-5.0` ") +
+          std::string(" -asm-verbose ") + std::string(" < ") + fileName;
+      FILE* fpipe = popen(command.c_str(), "r");
+      if (fpipe == nullptr) {
+        CHECK(false) << "Failed to popen() on command: " << command;
```
skimo-openhub: Why not `CHECK(fpipe != nullptr)`?

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+      }
+      ScopeGuard sg2([fpipe]() { pclose(fpipe); });
```
skimo-openhub: Hmm... I would just close it after I finished reading from it.

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -351,6 +355,45 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+      char buffer[512];
+      std::stringstream outputss;
+      while (1) {
+        int charactersRead = fread(buffer, 1, sizeof(buffer), fpipe);
+        if (charactersRead == 0) {
```
skimo-openhub: Shouldn't you check for errors here?

include/tc/core/execution_engine.h
skimo-openhub: I know you're just copying code, but it would be nice if you could add some
documentation to this function (before or after you moved it).
As to `MappingOptions` being CUDA-specific, perhaps it should be renamed
or subclassed from a generic version (assuming there are any options that
make sense in general).

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -44,14 +44,7 @@ class CudaTcExecutor : public ::tc::TcExecutor {
...
             tcDefinition,
             handle) {}
-  // CudaTcExecutor(
-  //     const std::string& def,
-  //     const std::vector<const DLTensor*>& inputsInfo)
-  //     : TcExecutor(def, inputsInfo) {}
-  // CudaTcExecutor(
-  //     lang::TreeRef tree,
-  //     const std::vector<const DLTensor*>& inputsInfo)
-  //     : TcExecutor(tree, inputsInfo) {}
```
skimo-openhub: Squash into previous commit.

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -23,20 +28,30 @@
...
   CudaTcExecutor(
-      const std::string& def,
-      const std::vector<const DLTensor*>& inputsInfo)
-      : TcExecutor(def, inputsInfo) {}
-  CudaTcExecutor(
-      lang::TreeRef tree,
-      const std::vector<const DLTensor*>& inputsInfo)
-      : TcExecutor(tree, inputsInfo) {}
+      std::string id,
+      std::vector<const DLTensor*> inputsInfo,
```
skimo-openhub: Why did you change the type of the first two argument?
That is, why are they no longer const references?
Note that the second argument of the `TcExecutor` constructor
_is_ a const reference.

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -80,14 +95,14 @@ class CudaTcExecutor : public ::tc::TcExecutor {
       const std::vector<const void*>& inputs,
       const std::vector<void*>& outputs) const;
 
-  bool hasRTCFun() {
+  bool hasRuntimeCompiledFun() override {
```
skimo-openhub: This looks like an independent change.

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -80,14 +95,14 @@ class CudaTcExecutor : public ::tc::TcExecutor {
...
     return rtcFun.get() != nullptr;
   }
 
   // It is necessary to clear the RTC manually because it can throw and we
   // can't have that in the destructor.
-  void clearRTC() {
-    if (!hasRTCFun()) {
+  void clearRuntimeCompiledFun() override {
+    if (!hasRuntimeCompiledFun()) {
```
skimo-openhub: This looks like an independent change.

src/core/tc_executor.cc
```diff
@@ -25,31 +28,23 @@ using namespace dlutils;
...
 int toTypeToken(DLDataType dtype) {
   return lang::TypeInfo(lang::TypeInfo::Code(dtype.code), dtype.bits)
       .toScalarToken();
 }
 } // namespace
 
 TcExecutor::TcExecutor(
-    const std::string& TcDefinition,
-    const std::vector<const DLTensor*>& inputsInfo)
-    : TcExecutor(parseOneFunction(TcDefinition), inputsInfo) {}
```
skimo-openhub: The removal of this constructor, along with `parseOneFunction`,
looks like an independent change.

include/tc/core/cuda/cuda_execution_engine.h
```diff
@@ -28,10 +28,9 @@
 
 namespace tc {
 
-/// The goal for this new shiny API is to provide a different pathway for being
-/// able to execute the kernels for multiple TC i.e. given the language which
-/// can have multiple TCs, people should be able to run things by just calling
-/// out the run function with the name of function and the inputs to run on.
+/// Given the language which can have multiple TCs, run underlying runtime
```
skimo-openhub: add comma before "which"

include/tc/core/tc_executor.h
```diff
@@ -90,8 +97,26 @@ class TcExecutor {
...
+  }
+
   const static size_t InvalidHandle = std::numeric_limits<size_t>::max();
 
+  std::string identifier;
+  std::vector<dlutils::DLTensorUPtr> inputsInfo;
+  std::string options;
+  /// When run is called this is used to find the most recently compiled
+  /// version.
+  size_t objectLocalHandle;
```
ftynse: do we still need this thing?  it's redundant with the object's position in `tcExecutors_` list

include/tc/core/execution_engine-inl.h
```diff
@@ -25,8 +25,6 @@
 
 namespace tc {
 
-using namespace dlutils;
-
```
skimo-openhub: Squash into previous commit.

include/tc/core/tc_executor.h
```diff
@@ -127,13 +127,20 @@ class TcExecutor {
   void checkInputsCompliant(
       const std::vector<const DLTensor*>& inputsInfo) const;
 
+  // This data structure contains the basic information that a TcExecutor
+  // needs to run a compiled kernel.
+  // This information corresponds to information required to:
+  // 1. build the kernel signature (kernelName/kernelParams)
+  // 2. retrieve the mapped kernel from cache (kernelName, options, input,
+  //    output)
+  // 3. actually run the kernel (supporting input, output tensors)
```
skimo-openhub: Looks good.

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -35,15 +35,10 @@ class CudaTcExecutor : public ::tc::TcExecutor {
   CudaTcExecutor(
       std::string id,
       std::vector<const DLTensor*> inputsInfo,
-      const MappingOptions& options,
+      const std::string& options,
```
skimo-openhub: These changes can be squashed into c91c2c6c81e223aea93a829eb504dd62dc8681c1 to resolve
@ftynse's remark.

ftynse:
I considered a template ExecutionEngine but changed my mind because ExecutionEngine may be needed in python and I could not see how to export a particular template instantiation.
cc @prigoyal

prigoyal:
yeah, the templated instantiations will likely not play well with pybinds. Thinking a bit in future, we will abstract away the current pybinds to bind to execution engine and now the aten compiler as it does right now. This will allow pybinds to be usable and exportable to other frameworks.

src/core/tc_executor.cc
prigoyal: why is this constructor getting dropped? we want to allow taking `treeRef` or the string as the input. :)

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -338,9 +351,48 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
+      // Close and reopen as FILE*
+      auto tmp = fopen(fileCName, "wb+");
+      auto llvmIr = toString(module.get());
+      std::fputs(llvmIr.c_str(), tmp);
+      fclose(tmp); // Close properly so the piped command can open without race
+
+      std::string fileName(fileCName);
+      // TODO: evolve this in a better way to find LLVM
+      auto command = std::string(" `which llc || which llc-5.0` ") +
```
prigoyal: can this be rather handled at cmake level? doing this here increases chances of spurious build failures that people might encounter

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -23,20 +28,18 @@
...
-      const std::vector<const DLTensor*>& inputsInfo)
-      : TcExecutor(def, inputsInfo) {}
-  CudaTcExecutor(
-      lang::TreeRef tree,
-      const std::vector<const DLTensor*>& inputsInfo)
-      : TcExecutor(tree, inputsInfo) {}
+      std::string id,
+      std::vector<const DLTensor*> inputsInfo,
+      const std::string& options,
+      lang::TreeRef tcDefinition,
```
prigoyal: `const`?

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -23,20 +28,18 @@
...
+      size_t handle)
```
prigoyal: `const`?

include/tc/core/cuda/cuda_tc_executor.h
```diff
@@ -23,20 +28,18 @@
...
+      : TcExecutor(id, inputsInfo, options, tcDefinition, handle) {}
```
prigoyal: why do we need to pass the options, handle and tcdef as well? isn't handle obtained for the `tcdef` itself?

include/tc/core/cuda/cuda_tc_executor.h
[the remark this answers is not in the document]
prigoyal: agreed with @ftynse on this one, cognitively, I also found it better to see `hasRTCFunction` so that we can relate things better to what's going on :)

include/tc/core/cuda/cuda_tc_executor.h
prigoyal: rename also to `clearRTCFunction` following the comment above.

include/tc/core/execution_engine-inl.h
```diff
@@ -0,0 +1,228 @@
...
+  }
+}
+
+// support define if we pass the parsed TreeRefs.
+template <typename ExecutorType>
+void ExecutionEngine<ExecutorType>::define(
+    const std::vector<lang::TreeRef>& treeRefs) {
+  std::lock_guard<std::mutex> lg(tcExecutorMutex_);
+  for (auto& ref : treeRefs) {
+    auto name = lang::Def(ref).name().name();
```
prigoyal: can you try to reuse the code from previous define? the earlier define should basically dispatch call to this define. If we rewrite code everywhere, it makes hard in future to maintain codebase as we need to change things everywhere.

include/tc/core/execution_engine-inl.h
```diff
@@ -0,0 +1,228 @@
...
+    tcNameMap_.emplace(std::make_pair(name, ref));
```
prigoyal: let's try to reuse the code.

include/tc/core/execution_engine-inl.h
```diff
@@ -0,0 +1,228 @@
...
+              compareDLTensorVectorMetadata(
+                     extractRawPtrs(e->inputsInfo), inputs);
+        });
+    if (e != executors_.end()) {
+      return (*e)->inferOutputTensorInfo();
+    }
+  }
+
+  // Otherwise, create a new executor and add it to executor_ with
+  // null options.  It will be used for further size queries but
```
prigoyal: nit: extra space before `It`

include/tc/core/execution_engine-inl.h
```diff
@@ -0,0 +1,228 @@
...
+    const std::string& options) {
+  // Check if we already have a handle for this name+size+options combination.
+  // If so, return it.
+  size_t handle = getHandle(name, inputs, MappingOptions(options));
+  if (handle != TcExecutor::InvalidHandle) {
+    return handle;
+  }
+
+  // Otherwise we need to compile.
+  std::unique_ptr<ExecutorType> p(new ExecutorType(
```
prigoyal: can we chose a better and descriptive name rather than `p`?

include/tc/core/execution_engine-inl.h
```diff
@@ -0,0 +1,228 @@
...
+// Steal TcExecutor and give it back under lock
+// Run outside of lock on owning TcExecutor.
+template <typename ExecutorType>
+Duration ExecutionEngine<ExecutorType>::run(
+    size_t handle,
+    const std::vector<const DLTensor*>& inputs,
+    const std::vector<DLTensor*>& outputs,
+    bool profile,
+    std::function<bool(const ExecutorType*)> pruningFunction) {
+  std::unique_ptr<TcExecutor> p(nullptr);
```
prigoyal: better naming please

src/core/polyhedral/codegen_llvm.cc
nicolasvasilache: Yes, that's why it's marked TODO. 

One possible way is to have a CLANG_PREFIX macro in the same way we have CUDA_HOME / CUB_HOME.
However I am not sure yet what the best solution is because we need to experiment a bit in our build system where setting CUDA_HOME / CUB_HOME seems to be problematic.

In any case, realize that this is completely unrelated to build failures. 
This **may** result in runtime errors when passing ```--llvm_dump_asm=1``` if the user does not have clang in his/her path.

src/core/polyhedral/codegen_llvm.cc
nicolasvasilache: This does almost exactly that, via scoping / RAII (ok technically```LOG(INFO)``` is actually printed before the close). I tend to prefer to keep related alloc/free close in code and "be done with them" early. This is also generally robust to control-flow, even though it is not a factor here.

include/tc/core/cuda/cuda_tc_executor.h
ftynse: I think the handle itself is redundant here. It's the index of executor in the list...

src/core/polyhedral/codegen_llvm.cc
nicolasvasilache: Should it tough? it's a single use 12 lines of code that are properly scoped (unless I am misunderstanding which part of the code you are referring to).
I can see how the whole chunk of code under ```if (FLAGS_llvm_dump_asm) {``` could live in a separate ```getAsm``` function, so I'll start with that.

include/tc/core/execution_engine.h
nicolasvasilache: yes that's the proper way to go after we start having a CPU mapper and can properly write / test different backends.
In the meantime I'll use string here to avoid immediate breakage in the upcoming CPU mapper.

ftynse: :+1: for string

include/tc/core/cuda/cuda_tc_executor.h
nicolasvasilache: Didn't change the type of the first 2 arguments.
This is a consolidation of 2 constructors of class (Cuda)TcExecutor  + 1 constructor of class (Cuda)ExecutorInfo into a single constructor of a (Cuda)TcExecutor class.
The signature of the resulting constructor is what makes sense for the consolidated class.

include/tc/core/cuda/cuda_tc_executor.h
nicolasvasilache: ```lang::TreeRef``` is used everywhere with by-value semantics, keeping the same behavior.
Underlying is a ```shared_ptr``` IIRC.

include/tc/core/cuda/cuda_tc_executor.h
nicolasvasilache: not for POD, by-value semantics

include/tc/core/cuda/cuda_tc_executor.h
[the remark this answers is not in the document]
nicolasvasilache: The rationale here is that whatever the backend (CPU, GPU, OpenCL, ROCm etc) there is a concept of a function that needs to be compiled at runtime before it can be evaluated. 

It happens to be called RTC / NVRTC for CUDA

Similarly to CUDA the thinking is that it should also be cleaned up when used in the autotuner.

I can't argue with the fact that this looks like premature optimization.
However the goal of this whole PR is to refactor things such that we are ready for multiplebackends.

So if y'all feel strongly about this I'll revert but I'd like this decision to be made in the context I mentioned above please :)

include/tc/core/tc_executor.h
```diff
@@ -90,8 +97,26 @@ class TcExecutor {
         << "TcExecutor::uncheckedRun is abstract and should not be called";
   }
 
+  virtual bool hasRuntimeCompiledFun() {
```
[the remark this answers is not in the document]
nicolasvasilache: See comment above, I cannot imagine that things would be significantly different for OpenCL but I don't have experience yet.

include/tc/core/cuda/cuda_tc_executor.h
ftynse: yup, it's a `shared_ptr`

include/tc/core/tc_executor.h
nicolasvasilache: nice catch thanks!

src/core/tc_executor.cc
nicolasvasilache: It's essentially dropped because it is not needed; a.k.a dead code.
We can resuscitate in the future if needed.

include/tc/core/cuda/cuda_tc_executor.h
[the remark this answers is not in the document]
ftynse: We need to think of a better name.  "runtime" and "compiled" in the same phrase is non-obvious :)  For LLVM, the produced result is called a module IIRC.  Can do this later.

src/core/tc_executor.cc
nicolasvasilache: @skimo-openhub this whole commit is about consolidating 2 classes and 3 constructors into 1 class and 1 constructor. I'd claim this is (and removing the dead code that results) is very much dependent change.
Makes sense?

include/tc/core/cuda/cuda_tc_executor.h
prigoyal: agreed, would be nice to see this fixed :)

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -195,8 +210,46 @@ class IslAstExprInterpeter {
   }
 };
 
-DEFINE_bool(llvm_dump_before_opt, false, "Print IR before optimization");
-DEFINE_bool(llvm_dump_after_opt, false, "Print IR after optimization");
+namespace {
+// In a separate process, call the LLC and dump the ASM
+// Since this is for debugging purposes only atm, separate process is fine.
+string emitAsm(const string& llvmIr) {
+  // Create a temporary file (do not use racy tmpname;
```
wsmoses: This seems hacky, I'm fairly confident LLVM has a nice way to emit assembly...

include/tc/core/execution_engine-inl.h
prigoyal: @nicolasvasilache do you plan to reuse the code? if not, can you explain why is it better to keep same code in two places which are doing essentially same thing :)

include/tc/core/execution_engine-inl.h
prigoyal: same comment as below :)

src/core/polyhedral/codegen_llvm.cc
skimo-openhub: Extracting it out (as you did) was my main point.
The other point was that I see several independent pieces in the code that communicate
with the system.  IMO, it would be better to extract those out and to make them available
at a "support" level such that they can get reused later.  Otherwise, we are likely to end
up with several similar implementations spread over the code.
Up to you.

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -206,6 +210,46 @@ class IslAstExprInterpeter {
   }
 };
 
+namespace {
+// In a separate process, call the LLC and dump the ASM
+// Since this is for debugging purposes only atm, separate process is fine.
+string emitAsm(const string& llvmIr) {
```
skimo-openhub: Doesn't compile.  Use `std::string`.

skimo-openhub:
On Tue, Mar 06, 2018 at 10:25:27PM +0000, Nicolas Vasilache wrote:
> >  int toTypeToken(DLDataType dtype) {
>    return lang::TypeInfo(lang::TypeInfo::Code(dtype.code), dtype.bits)
>        .toScalarToken();
>  }
>  } // namespace
>  
>  TcExecutor::TcExecutor(
> -    const std::string& TcDefinition,
> -    const std::vector<const DLTensor*>& inputsInfo)
> -    : TcExecutor(parseOneFunction(TcDefinition), inputsInfo) {}
> 
> @skimo-openhub this whole commit is about consolidating 2 classes and 3 constructors into 1 class and 1 constructor. I'd claim this is (and removing the dead code that results) is very much dependent change.
> Makes sense?

It doesn't look like the dead code is a *result* of the consolidation.
It was dead ever _before_.  See pr/dev/cpu-dead

skimo

include/tc/core/cuda/cuda_tc_executor.h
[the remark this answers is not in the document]
nicolasvasilache: @ftynse agreed but what does RTC stand for ;) ? Care to take this to NVIDIA ? :D :D 
But generally yes, let's find a better name for the abstraction and it happens to be called RTC in CUDA-land.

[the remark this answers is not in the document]
ftynse: RTC stands for Real-Time Clock, or réseau téléphonique commuté :P
Even if I take it somewhere, the answer is likely a bacwards-compatibility argument: it has been called this way, let's not change.

include/tc/core/execution_engine-inl.h
nicolasvasilache: yup getting to those in order of commits, I was hoping to finish yesterday before leaving but got interrupted ;)

include/tc/core/execution_engine-inl.h
nicolasvasilache: Note that this code has just been moved but wasn't implemented in this PR.
Still good point, I'll remove the copy-pasta in a followup commit thanks.

include/tc/core/execution_engine-inl.h
nicolasvasilache: Same as above, this is just moved code but your comment makes sense I'll address in a separate commit.

nicolasvasilache:
I'd like to understand a bit better what in the templated instantiations pose a fundamental problem with pybind. 

My claim is the value of templating this in the core library is high: to add a new backed one only needs to implement an Executor instead of both an Executor and an ExecutionEngine (esp since we have 2-3 backends in the works atm).

Generally the way one the way to reduce the impedence mismatch between non-templated "types" from scripting land and templated types from C++ land is to use simple dynamic dispatch.
It should work fine at the pybind level, does this make sense?

Still for the purpose of not blocking everyone on this I can strip the commit form this PR and submit a followup issue and PR just for this topic which requires more discussion.

src/core/polyhedral/codegen_llvm.cc
nicolasvasilache: I looked into the LLVM / LLC doc and got a nice refresher on machine instructions instruction selection etc. Still I couldn't find a way to easily plugin the code you shared with me. Care to add the option yourself in a separate PR? It'll take you significantly less time than it took me, and I can just scrap this :)
Another possibility is to land this and have you do a followup cleanup PR; calling a separate LLC process is done only in the debug mode and it's not like we don't have similar external ```popen``` behavior somewhere in the codebase ;)

Your call, just lmk what you prefer, thanks!

include/tc/core/cuda/cuda_tc_executor.h
prigoyal: is this getting fixed?

src/core/polyhedral/codegen_llvm.cc
wsmoses: I'd prefer removing this for now, then adding in later with the proper way unless its absolutely necessary right now simply because of the messiness we already have from popen/path-based things (aka ldconfig library lookup).

include/tc/core/cuda/cuda_tc_executor.h
nicolasvasilache: need to dig deeper, for now this just consolidates the constructors and kills some code but doesn't yet change the behavior. Prob a followup commit

include/tc/core/cuda/cuda_tc_executor.h
nicolasvasilache: Ah yes I missed the point about const, addressing.

nicolasvasilache:
Ok, every one-off comment has been addressed, thanks for your reviews!
Running tests one last time locally to confirm all is good.

There is one remaining topic which is the templated ExecutionEngine.
@ftynse suggested that:
```
apparently, you can create a type alias with using ExecutionEngineCuda = ExecutionEngine<TcExecutorCuda> and export that alias as a class through pybind
```
Works for me!

include/tc/core/execution_engine-inl.h
nicolasvasilache: addressed in c7effdb

include/tc/core/execution_engine-inl.h
nicolasvasilache: addressed in c7effdb

PR #103, opened 2018-03-06
Use database's langs instead of hardcoded lang

moskomule:
Hi, I found codes in `test_python/layers` use hard-coded langs. I think it is better to use `database`'s langs to test if they are correct. Is there any reason you do not use `database`'s langs? If not, I can change rest of test codes.


moskomule:
I see. Thank you

PR #111, opened 2018-03-06
[Hotfix] Temporarily make dump-cuda flag dump to cout

prigoyal:
this changeset somehow was missed when I moved the pytorch integration code to this repo.

using env GLOG_logtostderr is also fine however people often run things in python interpreter where setting this env variable is not friendly, or is an extra non-convenient step.

making this change and will refresh things. users complained in the slack channel about it.

### DIFF
--- a/src/core/tc_executor.cc
+++ b/src/core/tc_executor.cc
@@ -302,7 +302,9 @@ void TcExecutor::compileWithTcMapper() {
   // that.
   std::tie(execInfo_.cudaSource, execInfo_.grid, execInfo_.block) =
       mappedScop->codegen(execInfo_.kernelSpecializedName);
-  LOG_IF(INFO, FLAGS_dump_cuda) << "generatedCuda: " << execInfo_.cudaSource;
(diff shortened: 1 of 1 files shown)


nicolasvasilache:
Then please provide a function in python that prints what you want from the interpreter.
Removing the ability to print the cuda code to ```LOG(INFO)``` (and thus inspecting what the autotuner did after a few generations) is a regression IMO.

prigoyal:
@ftynse that doesn't work well when people want to do things in python interpreter

prigoyal:
merging this now after @ftynse 's accept

ftynse:
We still send the generated code to `INFO` with `--debug_tc_mapper`.

My reasons for accepting this are:
- this functionality was advertised in the release and pre-approved;
- a proper fix will be made available.

prigoyal:
yeah, this is a special flag that was added just for the release purpose. I do like your proposal and will do work on that as part of the issue that was created :) thanks for the help

PR #115, opened 2018-03-07
[Hotfix] emit macros for inf/inff

### LINKED ISSUE #63: max=! fails with inff not known to nvrtc
```
FAILS
def softmax(float(N, D) I) -> (O, expsum, maxVal) {
        maxVal(n) max=! I(n, d)
        expsum(n) +=! exp(I(n, d) - maxVal(n))
        O(n, d) = exp(I(n, d) - maxVal(n)) / expsum(n)
}
Error: https://gist.github.com/prigoyal/0318a964368f86b3a6a09228c7bfbf71

PASSES (bang removed from max=)
def softmax(float(N, D) I) -> (O, expsum, maxVal) {
        maxVal(n) max= I(n, d)
        expsum(n) +=! exp(I(n, d) - maxVal(n))
        O(n, d) = exp(I(n, d) - maxVal(n)) / expsum(n)
}

ftynse:
Max and Min reductions with default initialization require (negative)
infinity values.  NVRTC does not allow for including system headers, in
particular float.h and math.h that have standard defines for INIFINITY
and HUGE_VAL, resectively.  Currently, we emit "inff" and "inf" values
that are not defined.  Provide macros to define "inff" and "inf" using
CUDA intrinsics to convert from IEEE754 bit representation of infinities
to float and double.  Macros are used instead of proper constants
because CUDA does not allow calling the necessary intrincics in const
initialization.

A proper solution would require emitting
`std::numeric_limits<float>::infinity()` instead of "inff" and moving
bit-definitions in our re-implemented part of numeric_limits.

Closes #63

### DIFF
--- a/docs/source/framework/pytorch_integration/layers_database.rst
+++ b/docs/source/framework/pytorch_integration/layers_database.rst
@@ -139,7 +139,7 @@ Softmax
 .. code::
 
     def softmax(float(N, D) I) -> (O, maxVal, expDistance, expSum) {
-        maxVal(n) max= I(n, d)
+        maxVal(n) max=! I(n, d)
         expDistance(n, d) = exp(I(n, d) - maxVal(n))
         expSum(n) +=! expDistance(n, d)
         O(n, d) = expDistance(n, d) / expSum(n)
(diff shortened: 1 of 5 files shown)


prigoyal:
thanks @ftynse I will include this in our conda binaries as well since people are trying to run things with it and running into issues. :)

PR #118, opened 2018-03-07
Add unit test for broadcasted fcrelu

prigoyal:
this version of fcrelu works, adding it here


prigoyal:
@ftynse does it make sense to land this? I think while we work on other solutions and figure out broadcasting more properly, we should have test case supporting what's already in arxiv.

prigoyal:
yeah, there is a task to get parity between C++ test and python test, so this will be covered there.

PR #120, opened 2018-03-07
Log exception rather than just passing

prigoyal:
show the error message explicitly, one user had wrong TC but we were not displaying the error message.

### DIFF
--- a/tensor_comprehensions/tc_unit.py
+++ b/tensor_comprehensions/tc_unit.py
@@ -204,7 +204,8 @@ class TcAutotuner(object):
         try:
             best_options = self.autotuner.tune(cache_file, tc_name, inputs, options, [options])
             return best_options
-        except RuntimeError:
+        except Exception as e:
+            logger.error('Raised exception: {}'.format(e))
             return options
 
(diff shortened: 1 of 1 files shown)


tensor_comprehensions/tc_unit.py
```diff
@@ -204,7 +204,8 @@ def tune_and_store(self, tc_name, inputs, mapping_options, cache_file=""):
         try:
             best_options = self.autotuner.tune(cache_file, tc_name, inputs, options, [options])
             return best_options
-        except RuntimeError:
+        except Exception as e:
+            logger.error('Got exception: {}'.format(e))
```
wsmoses: is got the best word here? Maybe Raised?

prigoyal: yeah, will correct :) raised sounds better

ftynse:
At some point, we discussed CHECK macros throwing exceptions instead of killing everything... This would make them visible in python too.

prigoyal:
yep, all the exceptions thrown in C++ will be visible here which is I guess fine and users can provide info to us for debugging :)

PR #121, opened 2018-03-07
Update conda hash for refreshed packages

prigoyal:
two more fixes landed:
1. log the error on terminal
2. max=! reduction from Alex

they are part of package now. The updated packages are in my conda channel


ftynse:
hmm, can't we have this generated/updated automagically?  like the thing @skimo-openhub now has for bindings?

prigoyal:
yeah, we can do this but I am going to do this via nightly builds on Jenkins CI and then we will get refreshed conda packages at whatever frequency we want.

I didn't automate it so far primarily for above reason and sometimes I have to carefully pick which master to build, sometimes conda package fails to build and I apply local patches. this will get better with Jenkins CI where gpu tests will run and nightly package will build. :)

PR #125, opened 2018-03-08
Repro error on isl null input

prigoyal:
relates to https://github.com/facebookresearch/TensorComprehensions/issues/124

error:
```
[prigoyal:~/local/TensorComprehensions: (master)]$ ./build/test/test_autotuner --gtest_filter=ATenCompilationUnitTest.Graph2 --debug_tc_mapper=true
Note: Google Test filter = ATenCompilationUnitTest.Graph2
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from ATenCompilationUnitTest
[ RUN      ] ATenCompilationUnitTest.Graph2
unknown file: Failure
C++ exception with description "/home/prigoyal/local/TensorComprehensions/third-party-install/include/isl/interface/isl.h:3185: NULL input" thrown in the test body.
[  FAILED  ] ATenCompilationUnitTest.Graph2 (2758 ms)
[----------] 1 test from ATenCompilationUnitTest (2758 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (2758 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ATenCompilationUnitTest.Graph2

 1 FAILED TEST
```

PR #127, opened 2018-03-08
[hotfix] Support min max bounds

### LINKED ISSUE #124: Exception thrown: "isl.h:3185: NULL input thrown in the test body"
one of our user reported an issue with TC
```
def graph2(float(N, C, H, W) I, float(N, C, R, T) J, float(KH, KW) W1) -> (O, Out) {
        O(n, c, h, w) +=! J(n, c, h + kh, w + kw) * W1(kh, kw)
        Out(i, j) +=! I(n, i, h, w) * O(n, j, h, w)
}
```
which consistently throws error:
```
unknown file: Failure
C++ exception with description "/home/prigoyal/local/TensorComprehensions/third-party-install/include/isl/interface/isl.h:3185: NULL input" thrown in the test body.
```

ftynse:
Loop bounds inferred by Halide may involve min or max operations, which
are not convertible into affine functions.  For the sake of bound
computation, compute lists of affine functions by recursively
flattening arguments of nested min and max operations.  Use all
affine functions from a list to define loop bounds.  In particular,
allow min in upper bounds and max in lower bounds to exploit
```
  x < min(a, min(b, c)) <=> x < a and x < b and x < c,
  x > max(a, max(b, c)) <=> x > a and x > b and x > c
```
equivalences.  Do not allow other cases.

Closes #124 #125

### DIFF
--- a/src/core/halide2isl.cc
+++ b/src/core/halide2isl.cc
@@ -82,44 +82,141 @@ isl::aff makeIslAffFromInt(isl::space space, int64_t val) {
   return isl::aff(isl::local_space(space), v);
 }
 
-isl::aff makeIslAffFromExpr(isl::space space, const Expr& e) {
+std::vector<isl::aff> makeIslAffBoundsFromExpr(
+    isl::space space,
+    const Expr& e,
+    bool allowMin,
+    bool allowMax);
+
+namespace {
+/*
+ * Convert Halide binary expression "op" into a list of isl affine functions by
+ * converting its LHS and RHS into lists of affs and concatenating those lists.
+ * This is intended to be used with Min/Max operations in upper/lower bound
+ * computations, respectively.  Essentially, this allows for replacements
+ *   x < min(a,min(b,c)) <=> x < a AND x < b AND x < c
+ *   x > max(a,max(b,c)) <=> x > a AND x > b AND x > c
+ */
+template <typename T>
+inline std::vector<isl::aff>
+concatAffs(isl::space space, T op, bool allowMin, bool allowMax) {
+  std::vector<isl::aff> result;
+
+  for (const auto& aff :
+       makeIslAffBoundsFromExpr(space, op->a, allowMin, allowMax)) {
+    result.push_back(aff);
+  }
+  for (const auto& aff :
+       makeIslAffBoundsFromExpr(space, op->b, allowMin, allowMax)) {
+    result.push_back(aff);
+  }
+
+  return result;
+}
+
+/*
+ * Convert Halide binary expression "op" into an isl affine function by
+ * converting its LHS and RHS into affs and combining them with "combine"
+ * into a single expression.  LHS and RHS are expected to only produce a single
+ * expression.
+ * This is intended for use with operations other than Min/Max that do not
+ * commute nicely in bounds, for example
+ *   x < a + max(b,c)  NOT <=>  x < a + b AND x < a + c for negative values.
+ */
+template <typename T>
+inline isl::aff combineSingleAffs(
+    isl::space space,
+    T op,
+    isl::aff (isl::aff::*combine)(isl::aff) const) {
+  auto left = makeIslAffBoundsFromExpr(space, op->a, false, false);
+  auto right = makeIslAffBoundsFromExpr(space, op->b, false, false);
+  CHECK_EQ(left.size(), 1);
+  CHECK_EQ(right.size(), 1);
+
+  return (left[0].*combine)(right[0]);
+}
+
+} // end namespace
+
+/*
+ * Convert Halide expression into list of isl affine expressions usable for
+ * defining constraints.  In particular, an expression starting with (nested)
+ * Max operations can be used for lower bounds
+ *   x > max(a,b) <=> x > a AND x > b
+ * while an expression starting with (nested) Min operations can be used for
+ * upper bounds
+ *   x < min(a,b) <=> x < a AND x < b.
+ * Arguments "allowMin" and "allowMax" control whether Min and Max operations,
+ * respecitvely, are allowed to be present in the expression. Note that they
+ * can only appear before any other operation and cannot appear together in an
+ * expression.
+ * If a Halide expression cannot be converted into a list of affine expressions,
+ * return an empty list.
+ */
+std::vector<isl::aff> makeIslAffBoundsFromExpr(
+    isl::space space,
+    const Expr& e,
+    bool allowMin,
+    bool allowMax) {
+  CHECK(!(allowMin && allowMax));
+
+  const Min* minOp = e.as<Min>();
+  const Max* maxOp = e.as<Max>();
+
   if (const Variable* op = e.as<Variable>()) {
     isl::local_space ls = isl::local_space(space);
     int pos = space.find_dim_by_name(isl::dim_type::param, op->name);
     if (pos >= 0) {
-      return isl::aff(ls, isl::dim_type::param, pos);
+      return {isl::aff(ls, isl::dim_type::param, pos)};
     } else {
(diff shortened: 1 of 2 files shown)


prigoyal:
looks good from my side, can't speak about the halide specific changes but test case looks good to me :) thanks for the quick hotfix, much appreciated :)

src/core/halide2isl.cc
```diff
@@ -82,44 +82,141 @@ isl::aff makeIslAffFromInt(isl::space space, int64_t val) {
...
+  const Max* maxOp = e.as<Max>();
+
   if (const Variable* op = e.as<Variable>()) {
     isl::local_space ls = isl::local_space(space);
     int pos = space.find_dim_by_name(isl::dim_type::param, op->name);
     if (pos >= 0) {
-      return isl::aff(ls, isl::dim_type::param, pos);
+      return {isl::aff(ls, isl::dim_type::param, pos)};
     } else {
+      // FIXME: thou shalt not rely upon set dimension names
```
abadams: Halide references all dimensions by name, so the only alternative here is to also pass a map from name to set dimension index along with every Expr that we may want to turn into an isl:aff.... Except the isl::space can already *be* that map for us if we set up the names correctly.

ftynse: I'm just leaving a comment in the code before @skimo-openhub complains...

src/core/halide2isl.cc
```diff
@@ -82,44 +82,141 @@ isl::aff makeIslAffFromInt(isl::space space, int64_t val) {
...
-        return isl::aff(ls, isl::dim_type::set, pos);
+        return {isl::aff(ls, isl::dim_type::set, pos)};
       }
     }
     LOG(FATAL) << "Variable not found in isl::space: " << space << ": " << op
                << ": " << op->name << '\n';
-    return isl::aff();
+    return {};
+  } else if (minOp != nullptr && allowMin) {
+    return concatAffs(space, minOp, allowMin, allowMax);
```
abadams: Once you've allowed a min, should allowMax still be true? What happens with x < max(min(max(min(....))))

ftynse: The trick with multiple constraints does not work, so it should not be allowed.  There's a check above that allowMin and allowMax should not be true at the same time.

abadams: Ah, didn't see that check

src/core/halide2isl.cc
```diff
@@ -82,44 +82,141 @@ isl::aff makeIslAffFromInt(isl::space space, int64_t val) {
...
+ * expression.
+ * If a Halide expression cannot be converted into a list of affine expressions,
+ * return an empty list.
+ */
+std::vector<isl::aff> makeIslAffBoundsFromExpr(
+    isl::space space,
+    const Expr& e,
+    bool allowMin,
+    bool allowMax) {
+  CHECK(!(allowMin && allowMax));
```
nicolasvasilache: Feels like this could actually happen with reasonable combinations of positive and negative indices.
In which case maybe this should be reported as a user-facing TC well-formedness error but scorch the earth via CHECK is too aggressive?

test/test_tc_mapper_bugs.cc
```diff
@@ -696,6 +696,25 @@ TEST(LayerNorm, ReferenceBelongsToTwoGroups) {
   atCompl.compile("layernorm", inputs, options);
 }
 
+TEST(Halide2Isl, MinInUpperBound) {
```
nicolasvasilache: Looks a bit light on the testing side, could wee see stuff with strides and constants such that ```gcd(x.y)==1``` to shake off a little bit more the different paths added in this PR (in particular ```div```)?
Is there anything we can do to "expect" that we get a combination of ```min``` or ```max``` and ```div```?

src/core/halide2isl.cc
```diff
@@ -82,44 +82,141 @@ isl::aff makeIslAffFromInt(isl::space space, int64_t val) {
...
+ * expression.
+ * This is intended for use with operations other than Min/Max that do not
+ * commute nicely in bounds, for example
+ *   x < a + max(b,c)  NOT <=>  x < a + b AND x < a + c for negative values.
+ */
+template <typename T>
+inline isl::aff combineSingleAffs(
+    isl::space space,
+    T op,
+    isl::aff (isl::aff::*combine)(isl::aff) const) {
```
nicolasvasilache: quickly eyeballing I can't figure out easily what this ```isl::aff::*combine``` is, care to explain briefly?

src/core/halide2isl.cc
```diff
@@ -82,44 +82,141 @@ isl::aff makeIslAffFromInt(isl::space space, int64_t val) {
...
+ * converting its LHS and RHS into lists of affs and concatenating those lists.
+ * This is intended to be used with Min/Max operations in upper/lower bound
+ * computations, respectively.  Essentially, this allows for replacements
+ *   x < min(a,min(b,c)) <=> x < a AND x < b AND x < c
+ *   x > max(a,max(b,c)) <=> x > a AND x > b AND x > c
+ */
+template <typename T>
+inline std::vector<isl::aff>
+concatAffs(isl::space space, T op, bool allowMin, bool allowMax) {
+  std::vector<isl::aff> result;
```
nicolasvasilache: Optionally you could write it this way with a proper ```operator+=```, but not worth changing, just mentioning.
```
std::vector<isl::aff> result(makeIslAffBoundsFromExpr(space, op->a, allowMin, allowMax));
results += makeIslAffBoundsFromExpr(space, op->b, allowMin, allowMax);
return result;
```

ftynse:
This is a hotfix, better solution requires constructing pw affs and sets
from those and looked more involved.

prigoyal:
Hi @ftynse , would you like to have this PR merged ? or wait out to implement permanent solution :) either is fine by me, just wanted to check on the status :)

src/core/halide2isl.cc
ftynse: This particular function and the approach it implements does not support any combination of min/max in the same expression.  You are not supposed to call it with both `allowMin` and `allowMax`, so it is better to kill the run early.  It will _not_ kill the run if there is a min/max combo in the inferred bounds because we never call it the way that fails the check. 

There is another check below that will kill everything.  Before, we would report a cryptic error from isl and die anyway, so this is an improvement IMO.

test/test_tc_mapper_bugs.cc
ftynse: This patchset intends to hotfix the problem reported in #124 hence the test.

> Is there anything we can do to "expect" that we get a combination of min or max and div?

I'm actually not sure if we can get `max` because we always start tensor subscripts at zero...

src/core/halide2isl.cc
ftynse: Yes, considered that but decided that `+=` in the context where we operate on affine functions with another overloaded `operator+` can be misleading.

ftynse:
@prigoyal as I told online, this is a quickfix for #124 reported by a user.  I know this is a limited solution.  I was expecting it to get merged the same day.  If somebody has resources to implement the proper solution, please go ahead, submit a PR and close this.  I'm fine either way.

prigoyal:
thanks, I have no preferences. Hotfix is fine but this is also not super urgent.

ftynse:
Strange. All other user reported bugs were considered urgent.

prigoyal:
@ftynse, I didn't make my wording clear. I didn't mean that this is not an important feature but I meant that this is more of a feature not bug [IMO]. User has been given the workaround they are using atm. So this will be part of next package releases in 1 month and we don't need to merge it quickly and rather we can spend time on finding better solution if that's the preference.  hope that clarifies :)

prigoyal:
cc @hmansell

PR #135, opened 2018-03-09
Relax cuda/cudnn version requirement

### LINKED ISSUE #82: Relax cuda versioning asserts from test harness
We assert cuda8, cudnn6 atm but it is there only for reproducibility reasons from some time ago. We should relax this.

@nicolasvasilache is working on it , creating the task for tracking for users.

prigoyal:
we used to have cuda8-cudnn6  strictly for reproducibility reasons but these are not needed anymore. 

ran all gpu tests and all pass

closes #82


examples/example_fixture.h
```diff
@@ -91,33 +91,33 @@ struct Benchmark : public ::testing::Test {
   void SetUp() {
     if (!FLAGS_disable_reproducibility_checks) {
       auto cudnnVersion = cudnnGetVersion();
-      CHECK_EQ(6021, cudnnVersion)
+      CHECK_GE(6021, cudnnVersion)
           << "[CUDNN][VERSION] Enforce strict result reproducibility";
```
ftynse: let's change this text to something less "strict"

examples/example_fixture.h
```diff
@@ -91,34 +91,34 @@ struct Benchmark : public ::testing::Test {
   void SetUp() {
     if (!FLAGS_disable_reproducibility_checks) {
```
ftynse: sorry for iteration, can we also rename this flag?  (and change autotuning commands stored somewhere in the code base)?

prigoyal:
flag has been renamed

nicolasvasilache:
coming late to the party but cool and thanks ! :D

PR #137, opened 2018-03-12
fix some compilation errors that happen on Buck

salexspb:


include/tc/core/polyhedral/schedule_tree.h
```diff
@@ -135,6 +135,9 @@ namespace detail {
 // [1] Verdoolaege, Guelton, Grosser & Cohen (2014). "Schedule trees". In
 // IMPACT 2014.
 //////////////////////////////////////////////////////////////////////////////
+std::ostream& operator<<(std::ostream& os, const ScheduleTree& tree);
```
prigoyal: I believe these were already landed in master? here https://github.com/facebookresearch/TensorComprehensions/pull/134/files

might need to rebase :)

include/tc/core/polyhedral/schedule_tree_elem.h
```diff
@@ -18,6 +18,7 @@
 #include <memory>
 #include <unordered_set>
 #include <vector>
+#include <sstream>
```
prigoyal: can you move this in alphabetical order? clang-format might complain

ftynse:
Do we have to include empty fields
```
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:
```

in the commit message?

include/tc/core/polyhedral/schedule_tree_elem.h
ftynse: s/may/will :)

include/tc/core/polyhedral/schedule_tree.h
ftynse: Depending on how long `dev` will live separated from master, we may need to backport fixes by merging master into dev.

salexspb: @prigoyal , @ftynse , @nicolasvasilache  could you guys give more insights on what is your strategy about dev vs. master? I.e. why did you suggest us using dev before? I assumed dev has more stuff we depend on. But if fixes land in master, shouldn't be master be a better place to work from? 

 And how should we on FBCode side proceed with syncing internal changes? 

Thanks!

include/tc/core/polyhedral/schedule_tree_elem.h
salexspb: Yeah, clang format is a pitta for syncing files with fbcode due to config differences :( We have this problem in Caffe2 as well. Hopefully we will find a way to sync. clang configs for OSS and fbcode. 

You can imagine number of lint errors I see when I imported the whole code base in another direction :)

salexspb:
@ftynse , I think this is my git on a devserver is configured this way. I'll check this next time. No special leaks though! :)

include/tc/core/polyhedral/schedule_tree.h
prigoyal: @salexspb my understanding for the dev/master is as follows:
1. because users also try to build TC from source, we want to have a stable master. Hence, we maintain a dev branch where the "disruptive" changes will go. The hotfixes, normal features can go in the master branch.
2. We will sync dev and master branch in 2 weeks (for example) or more frequently.

I think for your use case, you should use master to have stable versions in fbcode. I think you were using dev this time probably because of the uncuda stuff. If that's not urgent or blocker however, then I would recommend sticking to master branch or at least moving forwards.

For syncing on fbcode side, we need to write bash scripts [just like caffe2, pytorch, onnx] to pull code inside fbcode and update.

ftynse: @salexspb  I don't think we have a precise strategy...

I was the one pushing for dev/master separation, so here's my take on it.
People will download TC from git and try building it, so master should be quite stable.  Essentially, master is equivalent to rolling release.  All development happens in the dev branch.  When dev is stable enough, we merge it into master.  (Later, we can have an extra "staging" branch mid-way between branch and master, but we are not there yet). Only hotfixes for user-visible issues that cannot wait until the next dev merge go into master.  New features don't however small they are.  Otherwise, we risk divergence between branches and pesky merge conflicts.  Merged hotfixes should be backported into dev regularly, like a weekly batch or so.

I am willingly not specific about what gets merged when, I believe this should be based on the size, nature and completeness of the changes rather on some abstract man-month count.

No idea about fbcode, but I'd generally recommend external dependers to depend on master rather than dev.  Now, if you want to build without cuda, it has to be dev until it gets merged.  The reasonable "release" state for that to get merged would be the introduction of CPU backend.

salexspb: I think we will work from master then. As I see it being more stable indeed and also people sometimes merge things there directly.

My personal take is that master / dev is an extra overhead. Especially when you have more collaborators, it becomes hard to maintain and merge. I am as user will have to constantly pay attention which features or fixes I need go to master or dev.

 In Caffe2 we don't do that. Instead we have CI covering all the tests for all the platforms and don't merge things that break tests. And we have a lot of tests :) Of course issues happen, then it is top priority to fix those. 

Also there is a concept of releases which are supposed to be more stable than master. We are not great at doing that in Caffe2. In fact, this week we will meet with other OSS teams at Facebook with more mature OSS release process to learn some of the lessons.

include/tc/core/polyhedral/schedule_tree_elem.h
salexspb: yeah, as I mentioned, it is a problem. As your oss clang might just be of a different version than our internal. And then configs might be non compatible. And I agree, we need to make sure clang config for OSS TC, Caffe2 and PyTorch is the same :)

ftynse: Our CI runs clang-format-5.0 (the same we have prebuilt in docker).

prigoyal:
Hi @salexspb , would you like to rebase this PR and address comments? :)

ftynse:
It looks like both these changes were subsumed by #181 and #134

prigoyal:
yep, that looks correct. I'll close out this PR then. Thanks @ftynse and @salexspb :)

PR #146, opened 2018-03-13
Harden conda install check

prigoyal:
more properly checking whether conda is installed or not


prigoyal:
Hi @abadams , @ftynse @zdevito would you be able to give your review sometime soon? these fixes are required for Jenkins setup and it's not trivial to detect these before hand unless Jenkins test run after PR is landed

PR #149, opened 2018-03-13
promotion to registers (aka private memory)

ftynse:
Assuming mapping to threads starts from the innermost coincident
schedule dimension and from the thread x, promote to registers in each
subtree below the band member mapped to thread x.  Split bands if
necessary to ensure that this member is the last one in the band.
For each such band, collect references to tensors accessed below it.
Group together the references that have overlapping footprints and at
least one of them is a write to ensure the most recent value is read.
For each group, consider promotion to registers if the footprint
contains only one element (hence promotable to a register) and if each
element is accessed by at most one thread (registers are private to
threads).  Do not promote to registers if these references were already
promoted to shared memory as this would require either copying from
shraed memory to registers, or demoting from shared memory first.
Do not insert synchroniztaions around these copies as no two threads are
accessing the same value.  The compiler could load from memory to a
register anyway for most arithmetic operations.

### DIFF
--- a/src/core/polyhedral/memory_promotion_heuristic.cc
+++ b/src/core/polyhedral/memory_promotion_heuristic.cc
@@ -530,7 +530,8 @@ void promoteToSharedGreedy(
           continue;
         }
 
-        scop.promoteGroupToShared(
+        scop.promoteGroup(
+            Scop::PromotedDecl::Kind::SharedMem,
             tensorId,
             std::move(group),
             bandNode,
--- a/src/core/polyhedral/memory_promotion_heuristic.cc
+++ b/src/core/polyhedral/memory_promotion_heuristic.cc
@@ -617,7 +617,6 @@ void promoteToRegistersBelowThreads(
       // per-thread-group access relations.
       auto points = activeDomainPoints(root, band);
       auto partialSched = partialSchedule(root, band);
-      auto activeStmts = activeStatements(root, band);
 
       size_t nMappedThreads = 0;
       for (int j = 0; j < points.dim(isl::dim_type::param); ++j) {
@@ -666,7 +665,6 @@ void promoteToRegistersBelowThreads(
               tensorId,
               std::move(group),
               band,
-              activeStmts,
               partialSched);
         }
       }
--- a/src/core/polyhedral/memory_promotion_heuristic.cc
+++ b/src/core/polyhedral/memory_promotion_heuristic.cc
@@ -558,23 +558,6 @@ void promoteGreedilyAtDepth(
   mapCopiesToThreads(mscop, unrollCopies);
 }
 
-namespace {
-isl::val getParamValIfFixed(isl::union_set uset, int pos) {
-  auto val = isl::val::nan(uset.get_ctx());
-  for (auto set : isl::UnionAsVector<isl::union_set>(uset)) {
-    auto currentVal = set.plain_get_val_if_fixed(isl::dim_type::param, pos);
-    if (currentVal.is_nan()) {
-      return currentVal;
-    }
-    if (!val.is_nan() && val != currentVal) {
-      return isl::val::nan(uset.get_ctx());
-    }
-    val = currentVal;
-  }
-  return val;
-}
-} // namespace
-
 // Assuming the mapping to threads happens in inverse order, i.e. the innermost
 // loop is mapped to thread x, promote below that depth.
 void promoteToRegistersBelowThreads(
@@ -625,7 +608,7 @@ void promoteToRegistersBelowThreads(
           if (id != mapping::ThreadId::makeId(i)) {
             continue;
           }
-          if (getParamValIfFixed(points, j) ==
+          if (isl::getParamValIfFixed(points, j) ==
               isl::val::zero(points.get_ctx())) {
             continue;
           }
--- a/src/core/polyhedral/memory_promotion_heuristic.cc
+++ b/src/core/polyhedral/memory_promotion_heuristic.cc
@@ -300,9 +300,7 @@ bool isCoalesced(
       auto partialSchedule = isl::map::from_union_map(partialScheduleUMap);
       auto scheduleToNextX = makeNextElementMap(
           partialSchedule.get_space().range(), threadIdxxDepth);
-      auto scheduledAccess = isl::map(access)
-                                 .gist_domain(access.domain())
-                                 .apply_domain(partialSchedule);
+      auto scheduledAccess = isl::map(access).apply_domain(partialSchedule);
       auto accessedByAdjacentX = scheduleToNextX.apply_domain(scheduledAccess)
                                      .apply_range(scheduledAccess);
 
@@ -349,9 +347,7 @@ bool isPromotableToRegisterBelowThreads(
                    threadIdxxScheduleDepthState,
                    originalAccesses.domain().intersect(activePoints));
 
-  auto scheduledAccesses =
-      originalAccesses.gist_domain(originalAccesses.domain())
-          .apply_domain(schedule);
+  auto scheduledAccesses = originalAccesses.apply_domain(schedule);
 
   // Scheduled accesses contain maps from schedule dimensions to tensor
   // subscripts.  Compute the relation that between the schedule dimensions
@@ -460,7 +456,6 @@ void promoteToSharedGreedy(
   size_t remainingMemory = maxMemory;
   for (auto bandNode : bands) {
     auto groupMap = TensorReferenceGroup::accessedBySubtree(bandNode, scop);
-    auto activeStmts = activeStatements(root, bandNode);
     auto partialSched = partialSchedule(root, bandNode);
     auto activePoints = activeDomainPoints(root, bandNode);
 
@@ -535,7 +530,6 @@ void promoteToSharedGreedy(
             tensorId,
             std::move(group),
             bandNode,
-            activeStmts,
             partialSched,
             true);
         remainingMemory -= memoryRequirement;
--- a/src/core/polyhedral/memory_promotion_heuristic.cc
+++ b/src/core/polyhedral/memory_promotion_heuristic.cc
@@ -121,9 +121,9 @@ void mapCopiesToThreads(MappedScop& mscop, bool unroll) {
 /*
  * Transform schedule bands into a union_map.
  * Takes all partial schedules at leaves as MUPAs (without accounting for
- * intermediate non-band nodes), transforms them into union maps and intersects
+ * intermediate non-band nodes), intersects
  * their domain with the filters between the root and the
- * current leaves.
+ * current leaves and transforms them into union maps.
  * Mapping filters are ignored.
  */
 isl::union_map fullSchedule(const detail::ScheduleTree* root) {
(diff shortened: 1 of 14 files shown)


prigoyal:
@tensorCompBot retest this please

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -311,6 +311,47 @@ bool isCoalesced(
...
+
+  auto bands =
+      ScheduleTree::collectDFSPreorder(root, detail::ScheduleTreeType::Band);
+  std::function<bool(ScheduleTree * st)> containsDepth = [&](ScheduleTree* st) {
+    auto depthBefore = st->scheduleDepth(root);
+    auto band = st->elemAs<ScheduleTreeElemBand>();
+    auto depthAfter = depthBefore + band->nMember();
+    return depthBefore < depth && depthAfter >= depth;
+  };
+  return functional::Filter(containsDepth, bands);
```
nicolasvasilache: functional style rules :)

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -145,21 +145,24 @@ isl::union_map fullSchedule(const detail::ScheduleTree* root) {
...
     pathToRoot.push_back(node);
     for (auto n : pathToRoot) {
       if (auto filterNode = n->elemAs<ScheduleTreeElemFilter>()) {
-        current = current.intersect_domain(filterNode->filter_);
+        domain = domain.intersect(filterNode->filter_);
       }
     }
 
-    schedule = schedule.unite(current);
+    prefixMupa = isl::manage(isl_multi_union_pw_aff_intersect_domain(
```
nicolasvasilache: ouch, do we really need to go back to C and non-RAAI world?

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -315,6 +315,48 @@ bool isCoalesced(
...
+    const ThreadIdxxScheduleDepthState& threadIdxxScheduleDepthState,
+    const TensorReferenceGroup& group,
+    isl::union_map schedule,
+    size_t nThreads,
+    isl::union_set activePoints) {
+  auto originalAccesses = group.originalAccesses();
+
+  // Return early if more than one element needs to be stored in registers.
+  // TODO: support arrays in registers if they are only accessed with constant
+  // subscripts, e.g. if the inner loops are fully unrolled.
```
nicolasvasilache: Quick question regarding unrolling. 
Are you planning to unroll at this point in the mapping or to have some deferred mechanism that also involves codegen, or something else?

nicolasvasilache: or maybe you just meant full tiles here?

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -315,6 +315,48 @@ bool isCoalesced(
...
+    return false;
+  }
+
+  // Since this function is only supposed to be called on groups seen _below_
+  // thread mapping, all refs in the group must all have the same thread-x
+  // depth.
+  auto depth = computeThreadIdxxScheduleDepth(
+                   threadIdxxScheduleDepthState,
+                   originalAccesses.domain().intersect(activePoints)) +
+      1;
```
nicolasvasilache: nit: indenting is a bit sad here, is it what ```clang_format``` forces on us in this instance ?

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -315,6 +315,48 @@ bool isCoalesced(
...
+
+  auto scheduledAccesses =
+      originalAccesses.gist_domain(originalAccesses.domain())
+          .apply_domain(schedule);
+
+  for (auto sa : isl::UnionAsVector<isl::union_map>(scheduledAccesses)) {
```
nicolasvasilache: By the time I get to this, a few weeks after last looking at shared promotion I have forgotten what structure scheduledAccesses have, mind adding a small comment about what the structure of the ```union_map``` please?

src/core/polyhedral/memory_promotion_heuristic.cc
ftynse: I meant situations like accessing 
```c
for (int i = 0; i < 2; ++i) {
  A[2*threadIdx.x + i];
```
that would require having `register_A[2]` accessed as `register_A[i]`, which would go to local memory instead of registers.  Unless the `i` loop is unrolled, there is not point in promoting this access.

Because autotuner-controlled unrolling happens earlier in the transformation pipeline, I was planning to just if the loop is requested for unroll.

src/core/polyhedral/memory_promotion_heuristic.cc
ftynse: yes, I tried to convince it otherwise but failed...

ftynse: maybe having `1 + functioncall()` could help

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +545,114 @@ void promoteGreedilyAtDepth(
   mapCopiesToThreads(mscop, unrollCopies);
 }
 
+namespace {
+isl::val getParamValIfFixed(isl::union_set uset, int pos) {
```
nicolasvasilache: This one feels reminiscent of a function that @abadams was asking for and that @skimo-openhub exposed for us from ISL. Maybe I am remembering wrong but maybe there is a way to avoid writing this function at all?
If we do need to write it indeed, it looks like a good candidate to move to ```include/tc/external/detail/islpp.h```.

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +545,114 @@ void promoteGreedilyAtDepth(
...
+}
+} // namespace
+
+void promoteToRegistersBelowThreads(
+    Scop& scop,
+    const ThreadIdxxScheduleDepthState& threadIdxxScheduleDepthState,
+    size_t nRegisters) {
+  using namespace tc::polyhedral::detail;
+
+  // Assuming the mapping happens to threads happens in inverse order, i.e.
```
nicolasvasilache: nit: happens twice

nicolasvasilache: Comment at the level of the function rather than inside?

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +545,114 @@ void promoteGreedilyAtDepth(
...
+
+    // Collect all bands where a member is located at the given depth.
+    auto bands = bandsContainingScheduleDepth(root, depth);
+    // We may have no band members mapped to thread x in case when we
+    // force-mapped everything to one thread.
+    if (bands.size() == 0) {
+      continue;
+    }
+
+    // Keep only those bands for which the this depth was recorded.
```
nicolasvasilache: spurious ```the```

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +545,114 @@ void promoteGreedilyAtDepth(
...
+      // will involve all mapping parameters when we take them below the
+      // mapping.  Skip mapping parameters obviously mapped to 0, because they
+      // do not correspond to band members that should be fixed to obtain
+      // per-thread-group access relations.
+      auto points = activeDomainPoints(root, band);
+      size_t nMappedThreads = 0;
+      for (int j = 0; j < points.dim(isl::dim_type::param); ++j) {
+        auto id = points.get_space().get_dim_id(isl::dim_type::param, j);
+        for (size_t i = 0; i < mapping::ThreadId::kMaxDim; ++i) {
+          if (not(id == mapping::ThreadId::makeId(i))) {
```
nicolasvasilache: is this better than ```!=``` or is there an underlying compilation issue because missing operator?

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +545,114 @@ void promoteGreedilyAtDepth(
...
+
+    for (auto band : bands) {
+      // Find out how many threads are actually mapped.  Active domain points
+      // will involve all mapping parameters when we take them below the
+      // mapping.  Skip mapping parameters obviously mapped to 0, because they
+      // do not correspond to band members that should be fixed to obtain
+      // per-thread-group access relations.
+      auto points = activeDomainPoints(root, band);
+      size_t nMappedThreads = 0;
+      for (int j = 0; j < points.dim(isl::dim_type::param); ++j) {
```
nicolasvasilache: nit: use ```d``` for dimension?

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +545,114 @@ void promoteGreedilyAtDepth(
...
+        }
+      }
+
+      auto groupMap = TensorReferenceGroup::accessedBySubtree(band, scop);
+      for (const auto& tensorGroups : groupMap) {
+        auto tensorId = tensorGroups.first;
+
+        // TODO: sorting of groups and counting the number of promoted elements
+        // TODO: check if nvrtc is smart enough to reuse a register for
+        // variables in disjoint scopes or do we need to reuse the variable?
```
nicolasvasilache: I would certainly hope that nvrtc would do this but that may be a good question to engage with our NVIDIA friends. Pinging Vinod to see what he thinks

test/test_mapper_memory_promotion.cc
```diff
@@ -455,6 +455,31 @@ TEST_F(MapperMemoryPromotionRAW, throwIfCopiesBelowThreads) {
...
+  O(i,j) = O(i,j) + C(i,j)
+}
+)TC";
+
+    auto mappingOptions = MappingOptions::makeNaiveMappingOptions()
+                              .tile(tileSizes)
+                              .useSharedMemory(false)
+                              .usePrivateMemory(true);
+    auto mscop = makeMappedScop(tc, mappingOptions, parameters);
+    return std::get<0>(mscop->codegen("fun"));
```
nicolasvasilache: In the following changesets, after register promotion is activated can we please check that promotion indeed happens in this unit test?

src/core/polyhedral/memory_promotion_heuristic.cc
nicolasvasilache: cc @vinodgro

vinodgro: yes it should allocate the same register to different lifetimes or scopes.

src/core/polyhedral/memory_promotion_heuristic.cc
ftynse: it uses the function that @skimo-openhub mentioned, but that function is only working for _sets_ not unions of sets; so I had to write a wrapper

src/core/polyhedral/memory_promotion_heuristic.cc
ftynse: it was missing, introduced in a follow-up PR; I can cherry-pick it here

src/core/polyhedral/scop.cc
```diff
@@ -179,25 +179,46 @@ void checkFiltersDisjointStatements(const ScheduleTree* root) {
...
+    }
+
+    auto groupId = kvp.second.groupId;
+    if (promotedDecls_.count(groupId) != 0 &&
+        promotedDecls_[groupId].tensorId == tensorId) {
+      // FIXME: allow double promotion if copies are inserted properly,
+      // in particular if the new promotion is strictly smaller in scope
+      // and size than the existing ones (otherwise we would need to find
+      // the all the existing ones and change their copy relations).
+      std::cerr << "FIXME: not promoting because another promotion of tensor "
```
nicolasvasilache: actually this spews non trivial amounts of text, how about ```LOG(WARNING)``` ?

ftynse: just kill it

prigoyal:
@caffe2bot retest this please

PR #150, opened 2018-03-13
Recursive chown and create a conda env

prigoyal:
1. do recursive chown otherwise errors can still happen
2. create a separate conda env where we will install things


prigoyal:
@zdevito or @abadams or @ftynse  may I have a review on this? sorry for so many PRs but these are blockers to CI jenkins.

PR #151, opened 2018-03-13
filter PLDI 2LUT tests temporarily from CI

prigoyal:
2LUT tests in PLDI example/ are right now OOM. They are blocking the CI progress so I am blocking these 4 tests from running on CI for now temporarily.

1. 2LUT tests are also running somewhere so we are not removing testing some functionality
2. these tests require modification and the goal is to have CI functional for gpu and we can always enable these 4 tests later. 

This has been discussed over messenger and Albert has consented that if needed, we can disable for now. and like in 1) above, there are 2LUT tests  running as part of non-pldi tests on CI so we test functionality fully.

### DIFF
--- a/.jenkins/run_test.sh
+++ b/.jenkins/run_test.sh
@@ -8,9 +8,12 @@ source /etc/lsb-release
 # condition: if 16.04 and conda, run only python tests
 # condition: if any and non-conda, run test.sh only
 
+# TODO: modify 2LUT tests from example_MLP_model and enable on CI
+
 if [[ "$DISTRIB_RELEASE" == 14.04 ]]; then
   echo "Running TC backend tests"
-  ./test.sh
+  FILTER_OUT=example_MLP_model ./test.sh
+  ./build/examples/example_MLP_model --gtest_filter=-*2LUT*
   if [[ $(conda --version | wc -c) -ne 0 ]]; then
     source activate tc-env
     echo "Running TC PyTorch tests"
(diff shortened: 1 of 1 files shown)


nicolasvasilache:
Still in meetings, how about reducing the default size rather than
disabling ?

prigoyal:
I am deep into getting CI setup and didn't want to distract myself with changing sizes and figuring out options etc. :) there is still more work left on CI and I was prioritizing it first given the merge of dev/master is blocked atm. 

we are also running Tc2Lut test which is Caffe2 test so I am not super worried that we are leaving behind functionality testing. :)

prigoyal:
I am happy to have PR from someone and close this out :)

prigoyal:
@ftynse would you like to look into the small sizes for 2LUT and options accordingly? I remember last time you had enabled these tests by changing options :)

prigoyal:
I have tried to use 1e-5 as the sizes in PR #152 but the CI is still throwing errors. So I am trying out this PR now.

ftynse:
> Still in meetings, how about reducing the default size rather than disabling ?

We already have Caffe2.Tc2LUTOp test with smaller sizes, it seems to run fine on CI.

ftynse:
> @ftynse would you like to look into the small sizes for 2LUT and options accordingly? I remember last time you had enabled these tests by changing options :)

I don't recall that.  Two possibilities: at some point, we had command line options for sizes that I could change; I disabled shared memory promotion for indirect accesses but it's irrelevant for sizes.

prigoyal:
thanks @ftynse  you can look at PR #152 I have tried that solution _first_ but there is still some trouble with CI. so I think more tuning on sizes is needed :)

PR #153, opened 2018-03-13
Install and compile with ccache for faster builds

prigoyal:
installing this will significantly speed up the build time

### DIFF
--- a/docker/common/install_base.sh
+++ b/docker/common/install_base.sh
@@ -13,6 +13,7 @@ apt-get install -y --no-install-recommends \
   wget \
   unzip \
   vim \
+  autoconf \
   automake \
   libtool \
   valgrind \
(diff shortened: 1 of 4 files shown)


docker/common/install_ccache.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+set -ex
+
+# Install ccache from source.
+# Needs 3.4 or later for ccbin support
+pushd /tmp
+git clone https://github.com/ccache/ccache -b v3.4.1
```
ftynse: why can't we have it from binary repositories?  ubuntu has it.

ftynse: do we want some specific version?  why?

prigoyal: may I request that I go over the setup once using the recommended practice and then  I will be able to answer questions better. This PR is still marked as WIP. Thanks for cooperation. :)

prigoyal: to answer the question: trusty has ccache 3.1 version which doesn't work with nvcc well. The nvcc support was added in version 3.4.3 so we install this way.

prigoyal:
Hi @abadams , @zdevito can you help review this PR? I am not aware of who else is using ccache on the team. I have tested the docker images with these and TC builds fine and tests fine. Thanks :)

prigoyal:
thanks @zdevito :)

PR #156, opened 2018-03-14
Fix compilation issue: braces around scalar initializer

mingzhe09088:
This diff is to fix the following two compilation issues: 

tc/tc/include/tc/c2/copy_op.h:50:41: error: braces around scalar initializer [-Werror,-Wbraced-scalar-init]
                                .unroll({128});

tc/tc/test/test_basic.cc:90:23: error: braces around scalar initializer [-Werror,-Wbraced-scalar-init]
  std::vector<int> v1{{1}};


prigoyal:
thanks @mingzhe09088 , looks great. I'll test out the gpu tests for it once and we can merge it then :)

prigoyal:
running gpu tests on this one now and will merge afterwards :)

mingzhe09088:
Thank you. I will build TC OSS on my devserver. So I could test my diff before sending it out.

PR #158, opened 2018-03-14
use ccache in c2 build if installed

prigoyal:
caffe2 requires passing CUDA_NVCC_EXECUTABLE as cmake arg and doesn't catch it from the env. Hence ccache will only work if the argument is passed.

refactor the caffe2 build to be less redundant and enable ccache if it is installed. this saves build time.


prigoyal:
Hi @zdevito , would you be able to review this PR around ccache ? thanks :)

PR #161, opened 2018-03-15
[wip] Register promotion improvements

ftynse:
* If a group of references was promoted into shared memory, but it could
be also promoted to registers while covering exactly the same statement
instances accessing it, demote it from shared memory before promoting to
registers.

Stacked on #149

### DIFF
--- a/include/tc/core/polyhedral/scop.h
+++ b/include/tc/core/polyhedral/scop.h
@@ -324,9 +324,8 @@ struct Scop {
     return promotedDecls_;
   }
 
-  const std::
-      unordered_map<isl::id, std::vector<PromotionInfo>, isl::IslIdIslHash>&
-      activePromotions() const {
+  const std::vector<std::pair<isl::union_set, PromotionInfo>>&
+  activePromotions() const {
     return activePromotions_;
   }
 
@@ -377,7 +376,6 @@ struct Scop {
       isl::id tensorId,
       std::unique_ptr<TensorReferenceGroup>&& gr,
       detail::ScheduleTree* tree,
-      const std::unordered_set<isl::id, isl::IslIdIslHash>& activeStmts,
       isl::union_map schedule,
       bool forceLastExtentOdd = false);
 
@@ -468,9 +466,10 @@ struct Scop {
   std::unordered_map<isl::id, size_t, isl::IslIdIslHash> groupCounts_;
   // groupId -> (tensorId, groupSizes)
   std::unordered_map<isl::id, PromotedDecl, isl::IslIdIslHash> promotedDecls_;
-  // stmtId -> (group, partial schedule, groupId)
-  std::unordered_map<isl::id, std::vector<PromotionInfo>, isl::IslIdIslHash>
-      activePromotions_;
+  // (domain, group, partial schedule, groupId)
+  // Note that domain is a non-unique key, i.e. multiple groups can be listed
+  // for the same domain, or for partially intersecting domains.
+  std::vector<std::pair<isl::union_set, PromotionInfo>> activePromotions_;
 };
 
 std::ostream& operator<<(std::ostream& os, const Scop&);
--- a/include/tc/core/polyhedral/scop.h
+++ b/include/tc/core/polyhedral/scop.h
@@ -383,6 +383,8 @@ struct Scop {
       isl::union_map schedule,
       bool forceLastExtentOdd = false);
 
+  void demoteGroup(isl::id groupId);
+
   // Given a tree node under which the promotion copy statements were
   // introduced, insert syncthread statements before and after the copies.
   // The tree should match the structure:
--- a/include/tc/core/polyhedral/scop.h
+++ b/include/tc/core/polyhedral/scop.h
@@ -306,8 +306,11 @@ struct Scop {
   void promoteEverythingAt(std::vector<size_t> pos);
 
   struct PromotedDecl {
+    enum class Kind { SharedMem, Register };
+
     isl::id tensorId;
     std::vector<size_t> sizes;
+    Kind kind;
   };
 
   struct PromotionInfo {
@@ -356,7 +359,8 @@ struct Scop {
   // Assumes such argument exists.
   const Halide::OutputImageParam& findArgument(isl::id id) const;
 
-  // Promote a tensor reference group to shared memory, inserting the copy
+  // Promote a tensor reference group to a storage of a given "kind",
+  // inserting the copy
   // statements below the given node.  Inserts an Extension node below the give
   // node, unless there is already another Extension node which introduces
   // copies.  The Extension node has a unique Sequence child, whose children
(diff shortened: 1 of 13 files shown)


prigoyal:
@caffe2bot retest this please

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -315,6 +314,67 @@ bool isCoalesced(
   return true;
 }
 
+/*
+ * Check if the given "group" can be promoted to registers for the given active
+ * domain points under full "schedule" where "nThreads" consecutive dimensions
+ * are mapped to threads (the innermost of them being mapped to thread x) and
+ * the depth of this mapping can be obtained from threadIdxxScheduleDepthState.
+ *
+ * In parciular, the group's footprint must contain only one element and the
```
nicolasvasilache: particular

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -315,6 +314,67 @@ bool isCoalesced(
...
+      computeThreadIdxxScheduleDepth(
+                   threadIdxxScheduleDepthState,
+                   originalAccesses.domain().intersect(activePoints));
+
+  auto scheduledAccesses =
+      originalAccesses.gist_domain(originalAccesses.domain())
+          .apply_domain(schedule);
+
+  // Scheduled accesses contain maps from schedule dimensions to tensor
+  // subscripts.  Compute the relation that between the schedule dimensions
```
nicolasvasilache: extra ```that```

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +563,111 @@ void promoteGreedilyAtDepth(
...
+      // Find out how many threads are actually mapped.  Active domain points
+      // will involve all mapping parameters when we take them below the
+      // mapping.  Skip mapping parameters obviously mapped to 0, because they
+      // do not correspond to band members that should be fixed to obtain
+      // per-thread-group access relations.
+      auto points = activeDomainPoints(root, band);
+      size_t nMappedThreads = 0;
+      for (int j = 0; j < points.dim(isl::dim_type::param); ++j) {
+        auto id = points.get_space().get_dim_id(isl::dim_type::param, j);
+        for (size_t i = 0; i < mapping::ThreadId::kMaxDim; ++i) {
```
nicolasvasilache:
```
if (!MappingId::isThreadId(id)) {
  continue;
}
```

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +563,111 @@ void promoteGreedilyAtDepth(
...
+          if (id != mapping::ThreadId::makeId(i)) {
+            continue;
+          }
+          if (getParamValIfFixed(points, j) ==
```
nicolasvasilache: You could extend to a templated 
```
getParamValIfFixed<T>(points, j)
```
and just compare to 0

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -503,5 +563,111 @@ void promoteGreedilyAtDepth(
...
+                  *group,
+                  fullSched,
+                  nMappedThreads,
+                  points)) {
+            continue;
+          }
+          if (!hasReuse(*group, fullSched, depth)) {
+            continue;
+          }
+          // TODO: if something is already in shared, but reuse it within one
```
nicolasvasilache: Don't you have it backwards here?
First promote to registers at some depth below threadId mappings.
Then promote remaining stuff to shared if extra reuse remains to be exploited or coalescing is bad.

If you first promote to shared then promote again to private a bunch of issues can occur:
1. missed opportunities to promote to shared because of incorrect size estimate
2. extra complexity to undo promotion to shared
3. ```no point in keeping it in shared _if_ it gets promoted into a register``` is only true modulo proper coalescing

src/core/polyhedral/memory_promotion.cc
```diff
@@ -321,8 +321,11 @@ void addSingletonReferenceGroups(
   // access relations have a shape :: [D -> ref] -> O
   // use currying to isolate the D part before intersecting with the domain
   // Compute initial groups with single reference per group.
-  accesses = accesses.curry().intersect_domain(domain).uncurry();
   for (auto a : isl::UnionAsVector<isl::union_map>(accesses)) {
+    if (isl::union_map(a.curry()).intersect_domain(domain).is_empty()) {
```
nicolasvasilache: makes sense

src/core/polyhedral/memory_promotion_heuristic.cc
```diff
@@ -145,21 +145,24 @@ isl::union_map fullSchedule(const detail::ScheduleTree* root) {
...
     pathToRoot.push_back(node);
     for (auto n : pathToRoot) {
       if (auto filterNode = n->elemAs<ScheduleTreeElemFilter>()) {
-        current = current.intersect_domain(filterNode->filter_);
+        domain = domain.intersect(filterNode->filter_);
       }
     }
 
-    schedule = schedule.unite(current);
+    prefixMupa = isl::manage(isl_multi_union_pw_aff_intersect_domain(
```
nicolasvasilache: Export this rather than fallback to old days?
Maybe it is exported later?

include/tc/core/polyhedral/scop.h
```diff
@@ -331,7 +331,9 @@ struct Scop {
 
   std::vector<std::pair<isl::union_set, Scop::PromotionInfo>> activePromotions(
       isl::union_set activePoints,
-      isl::id tensorId) const;
+      isl::id tensorId) const {
+    return promotionsAtIndexes(activePromotionsIndexes(activePoints, tensorId));
```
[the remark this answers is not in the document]
ftynse: Both are correct in English ;) But ok if you insist.

src/core/polyhedral/memory_promotion_heuristic.cc
ftynse: I wish, but ids in _isl space_ are not `MappingId` and there is no easy way to convert them.

src/core/polyhedral/memory_promotion_heuristic.cc
ftynse: We can just have a comparison operator between `isl::val` and `int`.  I don't think we should narrow `isl::val` to int in a call.

ftynse:
> Don't you have it backwards here?

There are pros and cons of doing it both ways, or even in a single promotion pass.

If you want to promote to registers first, and then the same reference group to shared, you will have to find and modify the copy expressions that you inserted in _all_ register scopes.  If you do the other way around, you can just demote _one_ shared promotion.  It will free some space, but it's straightforward to promote something else there.

Conceptually, I found it simpler to always know where your data is currently located (global, shared, register), so I went for shared, then registers.

> First promote to registers at some depth below threadId mappings.
> Then promote remaining stuff to shared if extra reuse remains to be exploited or coalescing is bad.

Practice shows that we may want to use shared _even if there is no reuse_ for latency hiding reasons.  This is actually one of the main reasons why you saw perf regressions with registers compared to shared-only.

> 
> If you first promote to shared then promote again to private a bunch of issues can occur:
> 
>     missed opportunities to promote to shared because of incorrect size estimate
We can keep the collected data and call the promotion once again (it's greedy).

>     extra complexity to undo promotion to shared
Cutting a tree branch is far easier than rewriting the functions for copies.

>     no point in keeping it in shared _if_ it gets promoted into a register is only true modulo proper coalescing
> 
It is even trickier than that.

PR #163, opened 2018-03-15
Use correct ccache repo path

prigoyal:
the default ccache install path was incorrect and was failing build on Jenkins CI.

@mingzhe09088 if you have ccache setup, then you can pass the variable `CCACHE_WRAPPER_DIR` with the build command. Set this variable to the `ccache` bin path. You can find that path by running
`which ccache`


mingzhe09088:
@prigoyal Thanks!

PR #166, opened 2018-03-16
Add pybindings for gflags

### LINKED ISSUE #113: Investigate and fix python and C++ flags logging
right now the LOG_IF doesn't play nice when we consider user experience trying to do the cuda dump in python interpreter. THey normally can do 
```
env GLOG_logtostderr=1 python my_test.py -v
```
and but this is not friendly when people have to do the stuff in python interpreter. For this reason, I had added a separate flag dump_cuda and GlobalDebugInit from python was setting that flag. However, having LOG_IF necessitates that we set env variable and doesn't play nice with logging.

this should be fixed more properly.

prigoyal:
I have been thinking about how to go about it easily and the solution was in plain-sight. I added pybind for the flags.

Pros:
1. we can now replace std::cout -> LOG_IF
2. The flags can be reset many times and the init doesn't have to be just once

Cons:
1. bindings will have to be modified in future everytime a new flag is added that we want to expose to people

closes #113

### DIFF
--- a/docs/source/framework/pytorch_integration/debugging.rst
+++ b/docs/source/framework/pytorch_integration/debugging.rst
@@ -28,7 +28,7 @@ Example usage
     import tensor_comprehensions as tc
     import torch
 
-    tc.GlobalDebugInit(["--debug_tc_mapper=true", "--debug_lang=false"])
+    tc.GlobalDebugInit(debug_tc_mapper=True, debug_lang=False)
 
     matmul = tc.define(tc.database['matmul']['lang'], name='matmul')
     mat1, mat2 = torch.randn(3, 4).cuda(), torch.randn(4, 5).cuda()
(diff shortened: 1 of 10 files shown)


prigoyal:
thanks @ftynse , will do that

I will also test out this PR in a separate env to make sure it isn't flaky. :)

prigoyal:
I also tested this solution in docker image and it works as expected.

prigoyal:
@caffe2bot retest this please

prigoyal:
@caffe2bot retest this please

prigoyal:
PR is building on Jenkins CI now https://ezyang.github.io/pytorch-ci-hud/build/tensorcomp-pull-request - had to manually trigger it.

Thank you @nicolasvasilache :) yep, I'll do it in the follow-up PR. definitely, those two should be enabled easily :)

prigoyal:
@caffe2bot retest this please

prigoyal:
@caffe2bot retest this please

PR #168, opened 2018-03-17
Pr/fix cpu build

nicolasvasilache:
This PR should come after [this](https://github.com/facebookresearch/TensorComprehensions/pull/167).
Somehow when rebasing the former dev I must have made a mistake and compilation was broken for CPU.
We still have some refactoring to go on the way to a CPU mapper but at least compilation and tests should pass.

Note however that our CIs did not catch the error.


src/core/polyhedral/mapped_scop.cc
```diff
@@ -572,6 +573,9 @@ std::tuple<std::string, tc::Grid, tc::Block> MappedScop::codegen(
   code << "extern \"C\" {" << std::endl
        << emitCudaKernel(specializedName, *mappedScopForCodegen) << "}"
        << std::endl;
+#else
+  code << "TC was compiled with -DWITH_CUDA=0 which disabled cuda compilation";
```
ftynse: how about using `#error` to fail compilation with a meaningful message?

nicolasvasilache: This was a hotfix for CPU only build but it does not seem to affect anyone atm.
I have a much bigger refactoring in progress to clearly separate between: core, cpu, cuda and cross-compile cuda.
I'll revisit this PR in the bigger context and abandon for now.

nicolasvasilache:
Abandoning this one

PR #182, opened 2018-03-21
emitMappedTensorAccess: use isl AST expression generator

skimo-openhub:
It is not clear why a single piece is expected, but there shouldn't
be any need for such an assumption.
Use the AST expression generator instead of manually trying
to pick the affine expression apart.

### DIFF
--- a/src/core/polyhedral/codegen_cuda.cc
+++ b/src/core/polyhedral/codegen_cuda.cc
@@ -754,13 +754,8 @@ void emitMappedTensorAccess(
   auto astToPromoted =
       isl::pw_multi_aff(promotion).pullback(astToScheduledOriginal);
 
-  auto pma = isl::PMA(astToPromoted);
-  CHECK_EQ(pma.size(), 1) << "expected one piece, got " << astToPromoted;
-  auto ma = isl::MA(pma[0].second);
-  context.ss << promotionInfo.groupId.get_name();
-  for (int i = 0; i < ma.size(); ++i) {
-    context.ss << "[" << toString(ma[i]) << "]";
-  }
+  auto astBuild = isl::ast_build::from_context(astToPromoted.domain());
+  context.ss << astBuild.access_from(astToPromoted).to_C_str();
 }
 
 void emitDirectSubscripts(


ftynse:
@caffe2bot retest this please

ftynse:
I would have liked to understand why the problem was happening and have a non-randomized test case for it...

skimo-openhub:
On Wed, Mar 21, 2018 at 05:45:28AM -0700, ftynse wrote:
> I would have liked to understand why the problem was happening and have a non-randomized test case for it...

Me too.
How come we don't have a reproducible test case?
I thought Nicolas changed the error message to print the seed.

skimo

ftynse:
@skimo-openhub  It's not about seed, the assertion was firing during the autotuning, which is randomized by nature.  CI does not log the offending options, so we need to find them again.

skimo-openhub:
On Wed, Mar 21, 2018 at 02:51:21PM +0000, ftynse wrote:
> @skimo-openhub  It's not about seed, the assertion was firing during the autotuning, which is randomized by nature.  CI does not log the offending options, so we need to find them again.

Ah.  Couldn't the autotuner dump the representation of
the offending candidate solution when an exception is caught?

skimo

ftynse:
Sorry, there is another problem.  This was not an exception, but a `CHECK` macro.  Those don't throw but terminate immediately.

skimo-openhub:
On Wed, Mar 21, 2018 at 08:03:39AM -0700, ftynse wrote:
> Sorry, there is another problem.  This was not an exception, but a `CHECK` macro.  Those don't throw but terminate immediately.

Could we change those to throw an exception instead
(at least in the case where the code gets run under the autotuner)?

skimo

skimo-openhub:
On Wed, Mar 21, 2018 at 03:19:00PM +0000, ftynse wrote:
> > (at least in the case where the code gets run under the autotuner)?
> That would be pretty much all of them, except the autotuner itself.
> 
> We should consider some general way to replace all CHECKs by exceptions.

I created an issue to keep track of this: #193

skimo

PR #184, opened 2018-03-21
emitLLVMKernel: avoid relying on isl set variable names

skimo-openhub:
codegenISL relied on set variable names in two ways.
First, it assumed that the variable names of the domain
of the schedule were preserved through the entire
AST generation process.
Explicitly set the names of the space sent to
halide2isl::makeIslAffFromExpr instead of assuming
they are still available in the schedule.
halide2isl::makeIslAffFromExpr itself still relies on
the names in the space itself being preserved, but
the space does not get modified in between,
so the risk of the names disappearing is reduced.

Second, codegenISL would manually set the names of the range
of the schedule, expecting them to be preserved across subsequent operations,
but this is again not guaranteed.
Convert the corresponding expressions to isl::ast_expr objects instead.
As a nice bonus, the code generation for subscripts and variables
now share the same mechanism, while before one used isl::ast_expr
objects, while the other used an isl::pw_multi_aff.
As a result, a lot of code can be removed.

### DIFF
--- a/src/core/halide2isl.cc
+++ b/src/core/halide2isl.cc
@@ -316,6 +316,20 @@ struct ScheduleTreeAndDomain {
   isl::union_set domain;
 };
 
+/*
+ * Helper function for extracting a schedule tree from a Halide Stmt,
+ * recursively descending over the Stmt.
+ * "s" is the current position in the recursive descent.
+ * "set" describes the bounds on the outer loop iterators.
+ * Return the schedule tree corresponding to the subtree at "s",
+ * along with a separated out domain.
+ *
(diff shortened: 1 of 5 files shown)


src/core/halide2isl.cc
```diff
@@ -376,8 +382,10 @@ ScheduleTreeAndDomain makeScheduleTreeHelper(
     }
 
     // Recursively descend.
+    outer.push_back(op->name);
```
nicolasvasilache: Let's please make this functional style to avoid push/pop.
I.e. just construct a new vector, append op->name to it and let it go out of scope.

PR #187, opened 2018-03-21
Bunch of doc fixes and use correct reduction operators

### LINKED ISSUE #176: Bug of doc example
- OS: Linux64
- How you installed TC (docker, conda, source): conda
- Python version: Python3.6.4
- CUDA/cuDNN version: CUDA8.0/cuDNN7.0.5
- Conda version (if using conda): 4.4.11

1. There are a doc typo bug of `tc.options` about `group conv`, the desc of `conv` is same to `group conv`. 
qq20180320-110155 2x

2. There are a doc typo bug of `avgpool` example, the `avgpool` is not sum op. 
qq20180320-142650 2x

3. There are a doc typo bug of `maxpool` example, the `max=` should be `max=!`, or it will got wrong answer if the image size not divided by the kernel size. 
qq20180320-165027 2x

By the way, I think the doc is obscure to understand.

prigoyal:
bunch of doc fixes spotted by our user 

closes #176

### DIFF
--- a/tensor_comprehensions/library/layers.yaml
+++ b/tensor_comprehensions/library/layers.yaml
@@ -61,13 +61,13 @@
 - name: avgpool
   lang: |
     def avgpool(float(B, C, H, W) input) -> (output) {{
-      output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
+      output(b, c, h, w) +=! input(b, c, h * {sH} + kh, w * {sW} + kw) / ({kH} * {kW}) where kh in 0:{kH}, kw in 0:{kW}
     }}
 
 - name: maxpool
   lang: |
     def maxpool(float(B, C, H, W) input) -> (output) {{
-      output(b, c, h, w) max= input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
+      output(b, c, h, w) max=! input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
     }}
 
 - name: scale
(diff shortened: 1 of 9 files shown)


test_python/layers/test_avgpool_autotune.py
```diff
@@ -31,7 +31,7 @@ def test_avgpool_autotune(self):
         # NOTE: take note of use of {{ }} below for handling TC with scalars
         LANG = """
         def avgpool(float(B, C, H, W) input) -> (output) {{
-            output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
```
ftynse: I think we should have a warning for such uninitialized reductions if the target tensor is not input.

tensor_comprehensions/library/layers.yaml
```diff
@@ -61,7 +61,7 @@
 - name: avgpool
   lang: |
     def avgpool(float(B, C, H, W) input) -> (output) {{
-      output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) where kh in 0:{kH}, kw in 0:{kW}
```
ftynse: `+=!` here

tensor_comprehensions/library/layers.yaml
```diff
@@ -61,7 +61,7 @@
...
+      output(b, c, h, w) += input(b, c, h * {sH} + kh, w * {sW} + kw) / ({kH} * {kW}) where kh in 0:{kH}, kw in 0:{kW}
     }}
 
 - name: maxpool
```
ftynse: also change to `+=!` for `maxpool` here, for consistency

ftynse:
Also, this PR changes some `.py` files which I wouldn't consider as "doc fixes"

prigoyal:
Thanks @ftynse. I thought it was a good idea to make related changes in the
same PR. I will change the title to be more appropriate.

Yes, for redundancy there is a task open. Thank you.

ftynse:
> I thought it was a good idea to make related changes in the same PR.

It is a good idea indeed!  Otherwise, I wouldn't have noticed the mismatch between the documentation and the code.

test_python/layers/test_avgpool_autotune.py
prigoyal: agreed. Would you like to create an issue for it? I believe there was a discussion on this but there was no resolution IIRC.

ftynse: #190

Issue #193, opened 2018-03-23
autotuner should catch compilation failures


skimo-openhub:
By its very nature, the autotuner is not deterministic.
It may sometimes uncover compilation failures on the CI,
but these cause an abort instead of printing the feature vector
that could be used to reproduce the issue.
See, e.g., #181.

We probably need to redefine the `CHECK` macros to throw an exception,
at least when run under the autotuner.
Unfortunately, calling `google::InstallFailureFunction` to set a function that
throws an exception does not have the desired effect.
See also https://github.com/google/glog/issues/249

skimo-openhub:
Closed by #460.

PR #197, opened 2018-03-23
Decouple cuda

nicolasvasilache:
This PR starting breaking the hard dependence of the mapper on CUDA-specific concepts.
This is a necessary stepping stone towards a functional CPU mapper.
While there is still work to do to get a functional CPU mapper this can be reviewed and land independently.


examples/example_MLP_model.cc
```diff
@@ -24,7 +24,7 @@
 #include <ATen/ATen.h>
```
salexspb: It would help a lot to name examples / tests that require CUDA with 'cuda' in their name (or in  a cuda subfolder), but this @doesn't have to be in the same PR, of course

include/tc/core/cuda/cuda_mapping_options-inl.h
```diff
@@ -0,0 +1,256 @@
+/**
```
salexspb: why isn't this a .cc file? -inl would be needed for template implementations, here it looks like it should work inside .cc. .cc files are faster to compile if you later on do incremental changes to them.

include/tc/core/cpu/cpu_mapping_options-inl.h
```diff
@@ -0,0 +1,40 @@
...
+namespace tc {
+
+CpuMappingOptions::CpuMappingOptions()
+    : ownedProto_(), genericMappingOptionsView(ownedProto_) {}
+
+/// Construct from a serialized protocol buffer message.
+CpuMappingOptions::CpuMappingOptions(const std::string& str)
+    : CpuMappingOptions() {
+  bool parsed = ownedProto_.ParseFromString(str);
+  CHECK(parsed) << "could not parse protobuf string";
```
salexspb: on a separate note, in Caffe2 earlier days we had to cut all the CHECKs out. Because they can crash a process and are not compatible with production workloads. So we use CAFFE_ENFORCE and CAFFE_THROW which use exceptions instead.

include/tc/core/cuda/cuda_mapping_options.h
```diff
@@ -0,0 +1,248 @@
...
+
+  /// Compare with another message.
+  inline bool operator==(const CudaMappingOptions& options) const;
+  inline bool operator!=(const CudaMappingOptions& options) const;
+
+  /// Construct from a serialized protocol buffer message.
+  inline explicit CudaMappingOptions(const std::string& str);
+
+  std::string toProtobufSerializedString() const {
+    return ownedProto_.SerializeAsString();
```
salexspb: so even though classes are separated, underlying protobuf representations stay the same. Just wondering, did you consider protobuf separate as well? So different backends don't pollute the same protobuf with options that make sense only in their context.

nicolasvasilache: Full separation requires copy-pasting the contents of the protos and then require reimplementing the mutation functions for all classes, so I am not sure it buys us much?

include/tc/core/cuda/cuda_mapping_options-inl.h
nicolasvasilache: good point, will address, thanks!

examples/example_MLP_model.cc
nicolasvasilache: Will do, I barely managed to break the SDK dependence in the cmake and I am not very happy about the form it took.  Suggestions welcome.

include/tc/core/cpu/cpu_mapping_options-inl.h
ftynse: yep, help on moving away from these cheks is much appreciated!  I was considering to define my own macros and just search/replace in the codebase, but if there is a better way -- let's do it.

include/tc/core/cuda/cuda_mapping_options-inl.h
ftynse: except when functions are inline, and we probably want the two-liner "fluent interface" calls to stay inline

examples/example_MLP_model.cc
ftynse: ideally, we want the same example to run on different backends in the future :)

ftynse:
I still have to read it fully...

include/tc/core/cpu/cpu_mapping_options-inl.h
skimo-openhub: It may be easier to redefine CHECK.  See also #193.

src/core/CMakeLists.txt
```diff
@@ -27,8 +26,12 @@ add_library(
   polyhedral/schedule_print.cc
   polyhedral/scop.cc
   polyhedral/separation.cc
-  polyhedral/tighten_launch_bounds.cc
   polyhedral/unroll.cc
+
+  # Builds and runs without the CUDA SDK and used in cross-compilation tests
+  polyhedral/codegen_cuda.cc
+  polyhedral/cuda/cuda_mapping_types.cc
+  polyhedral/cuda/cuda_tighten_launch_bounds.cc
```
skimo-openhub: Looks like an independent change.

examples/example_MLP_model.cc
salexspb: @ftynse , sure, but we need somehow to enable CPU only mode to run correctly for it :) May be we can use macro approach for it

include/tc/core/cpu/cpu_mapping_options-inl.h
salexspb: I would not redefine check, actually. As it is very magical. In Caffe2 we just did a codemod and replaced all CHECKs with CAFFE_ENFORCE / CAFFE_THROW

Also this gives ability to add custom exception being thrown. Which could have extra information.

include/tc/core/cuda/cuda_mapping_options.h
salexspb: How about this?

message MappingOptionsProto {
  optional CudaMappingOptionsProto cuda_options,
  optional CPUMappingOptionsProto cpu_option,
  <the  rest of common options goes here>
}

the in the code we won't accept a message where both cpu and cuda mapping options are set. This emulates union support which is not available in protobuf (is available in thrift).

include/tc/core/mapping_options-inl.h
```diff
@@ -22,18 +22,15 @@ namespace tc {
 //
```
salexspb: this file also seem to be doable in .cc instead of .h

src/proto/mapping_options.proto
```diff
@@ -42,25 +42,31 @@ message MappingOptionsProto {
...
   optional uint64 unroll = 7;
   // Apply tiling to imperfectly nested loops if possible.  Same tilesizes are
   // used.
   required bool tile_imperfectly_nested = 8;
+  // reserved 5, 6, 9 to 13; can only activate with proto3
+  // Reserved: 5, 6, 9-13 -> factored out into CudaMappingOptionsProto
+  required bool match_library_calls = 14;
+}
+
+message CudaMappingOptionsProto {
```
salexspb: @nicolasvasilache , looks like I misread this part before. The split actually looks good here. I.e. fields are not being flattened out. Somehow initially I thought that this file is not changed.

include/tc/core/cuda/cuda_mapping_options.h
```diff
@@ -158,50 +158,6 @@ class CudaMappingOptions {
     return ownedProto_.SerializeAsString();
   }
 
-  /// The following only call the operation on CudaMappingOptions and return
-  /// reference to self for the purpose of chaining
```
skimo-openhub: This comment seems useful enough to be preserved, with some modification.

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -123,7 +111,25 @@ PYBIND11_MODULE(mapping_options, m) {
...
+          [](tc::CudaMappingOptions& instance, bool fix) {
+            instance.fixParametersBeforeScheduling(fix);
+          },
+          "Perform automatic loop scheduling taking into account specific tensor sizes.\nMay produce faster kernels but significantly increases compilation time.\n Note that the mapping will be performed for specific tensor sizes anyway")
+      .def(
+          "unroll",
+          [](tc::CudaMappingOptions& instance, uint64_t factor) {
+            instance.unroll(factor);
+          },
+          "Perform loop unrolling on the generated code and produce at most the given number of statements");
```
skimo-openhub: Is this file automatically generated?

src/core/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+################################################################################
+# tc_core
+#
+# Core CPU library with cross-compilation capabilities linked from
+# tc_core_cuda
```
skimo-openhub: It also appears to be linked from `tc_core_cpu`.

src/core/CMakeLists.txt
```diff
@@ -84,24 +87,83 @@ install(
...
     PATHS ${CUDA_TOOLKIT_ROOT_DIR}
     PATH_SUFFIXES lib lib64 targets/x86_64-linux/lib targets/x86_64-linux/lib/stubs)
 
   add_library(
-    tc_core_cuda
+    tc_cuda
 
     SHARED
 
+    # This block of file is copy-pasted from tc_core_cuda_no_sdk
```
skimo-openhub: "files"

src/core/CMakeLists.txt
```diff
@@ -84,24 +87,83 @@ install(
...
+    # Don't know a better way atm
```
skimo-openhub: I'm afraid I can't help here.  Not a cmake expert.

test/test_mapper_llvm.cc
```diff
@@ -82,7 +82,7 @@ def fun(float(N, M) A, float(N, M) B) -> (C) {
 
   ExecutionEngine<CpuTcExecutor> engine;
   engine.define(tc);
-  auto options = tc::CudaMappingOptions::makeNaiveCudaMappingOptions();
+  auto options = tc::MappingOptions::makeNaiveMappingOptions();
```
skimo-openhub: This part should be squashed into cb8fbbedea88968dd58e4e224009fd22ce2138e7.

src/core/CMakeLists.txt
ftynse: declare a variable with this list and reuse it

src/core/polyhedral/mapped_scop.cc
```diff
@@ -584,71 +584,76 @@ std::tuple<std::string, tc::Grid, tc::Block> MappedScop::codegen(
 
 std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
     std::unique_ptr<Scop>&& scopUPtr,
-    const MappingOptions& options) {
```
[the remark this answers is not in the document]
nicolasvasilache: hmm rebase fun, thanks!

examples/example_MLP_model.cc
```diff
@@ -520,8 +520,8 @@ TEST_F(ProductionModel, 1LUT) {
   auto D = FLAGS_D;
   auto L1 = FLAGS_L1;
   auto E1 = FLAGS_E1;
-  auto options = tc::MappingOptions::makeNaiveMappingOptions()
-                     .tile({1, 32})
+  auto options = tc::CudaMappingOptions::makeNaiveCudaMappingOptions()
+                     .tile(1, 32)
```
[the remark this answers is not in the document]
nicolasvasilache: Just because I dropped initializer list for tile (coming from forwarding macros).
I could work harder to bring initializer list back but at this point with the variadic template it feels unnecessary.
We could also for consistency drop initializer list everywhere, they don't play as nicely with templates.

include/tc/autotuner/genetic_autotuner.h
```diff
@@ -37,19 +37,19 @@ class GeneticAutotuner {
 
   void storeCaches(const std::string& filename);
 
-  std::vector<MappingOptions> load(
+  std::vector<CudaMappingOptions> load(
```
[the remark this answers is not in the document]
nicolasvasilache: yes we shouldn't and it is a followup PR
At least this makes it very explicit everywhere we can that currently the only behavior is Cuda specific.
Anything that is not templated by a TcExecutor is future work.

include/tc/core/cuda/cuda_mapping_options.h
```diff
@@ -0,0 +1,248 @@
...
+  FORWARD_FUN(intraTileScheduleAllowSkewing);
+  FORWARD_FUN(intraTileSchedulePositiveOrthant);
+
+#undef FORWARD_FUN
+
+ private:
+  CudaMappingOptionsProto ownedProto_;
+
+ public:
+  MappingOptionsView genericMappingOptionsView;
```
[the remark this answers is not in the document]
nicolasvasilache: Unclear, so far my experience has been that having a concrete object inherit from a view creates all sorts of hellish allocation order issues and tricky segfaults.
This would be solved more easily if proto supported inheritance but unfortunately it does not.

tensor_comprehensions/pybinds/pybind_options.cc
nicolasvasilache: no it's manual and part of the python bindings that @prigoyal did

src/core/CMakeLists.txt
nicolasvasilache: easy enough :) thanks!

tensor_comprehensions/pybinds/pybind_options.cc
skimo-openhub: Then why did these entries get reordered by this commit?
That makes it more difficult to see what actually changed.

PR #198, opened 2018-03-23
[RFC] Cuda Profiling

ttheodor:
I used CUPTI to add a profiling functionality in `CudaRTCFun`. This is still work in progress: there is no integration with caches yet, nor the autotuner.

There are several performance metrics (listed [here](http://docs.nvidia.com/cuda/cupti/r_main.html#metrics-reference). Each metric requires measuring (possibly multiple typed of) hardware events. Since not all events can be measured at the same time, a kernel must be launched multiple times (t's 12 times with the current list of metrics that I hardcoded).


include/tc/core/cuda/cuda_profile.h
```diff
@@ -0,0 +1,66 @@
...
+  CudaMetric(const char* name_, CUdevice device);
+
+  std::string name;
+  CUpti_MetricID id;
+  uint32_t numberEvents;
+
+  CUpti_MetricValue value;
+};
+
+using KernelTy = std::function<void(void)>;
```
nicolasvasilache: ```KernelType``` ?

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+    uint64_t numInstances,
+    uint64_t numTotalInstances,
+    uint64_t sum,
+    uint64_t normalized,
+    uint64_t* values) {
+  char eventName[128];
+  size_t eventNameSize = sizeof(eventName) - 1;
+  TC_CUPTI_CHECK(cuptiEventGetAttribute(
+      eventId, CUPTI_EVENT_ATTR_NAME, &eventNameSize, eventName));
+  eventName[127] = '\0';
```
nicolasvasilache: ```eventName[eventNameSize]``` ?

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+  CHECK_GT(eventNameSize, 0);
```
nicolasvasilache: Move before and also check < 128?

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+};
+
+void logEvent(
+    CUpti_EventID eventId,
+    uint64_t numInstances,
+    uint64_t numTotalInstances,
+    uint64_t sum,
+    uint64_t normalized,
+    uint64_t* values) {
+  char eventName[128];
```
nicolasvasilache: constexpr blah = 128; ?

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+  if (numInstances > 1) {
+    for (uint64_t k = 0; k < numInstances; k++) {
+      if (k != 0) {
+        ss << ", ";
+      }
+      ss << values[k];
+    }
+  }
+  ss << ')';
+  LOG(INFO) << ss.str();
```
nicolasvasilache: cleanup plz, maybe it already is cleaned up?

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+
+  ss.str("");
+  ss.clear();
+
+  ss << eventName << " (normalized) (" << sum << " * " << numTotalInstances
+     << ") / " << numInstances << " = " << normalized;
+  LOG(INFO) << ss.str();
```
nicolasvasilache: cleanup plz, maybe it already is cleaned up?

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+        &numTotalInstances));
+    TC_CUPTI_CHECK(cuptiEventGroupGetAttribute(
+        group,
+        CUPTI_EVENT_GROUP_ATTR_INSTANCE_COUNT,
+        &numInstancesSize,
+        &numInstances));
+    TC_CUPTI_CHECK(cuptiEventGroupGetAttribute(
+        group, CUPTI_EVENT_GROUP_ATTR_NUM_EVENTS, &numEventsSize, &numEvents));
+    eventIdsSize = numEvents * sizeof(CUpti_EventID);
+    eventIds = (CUpti_EventID*)malloc(eventIdsSize);
```
nicolasvasilache:
```
ScopeGuard([]{free(eventIds);});
```

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+    for (uint32_t j = 0; j < numEvents; j++) {
+      values.clear();
+      valuesSize = sizeof(uint64_t) * numInstances;
+      TC_CUPTI_CHECK(cuptiEventGroupReadEvent(
+          group,
+          CUPTI_EVENT_READ_FLAG_NONE,
+          eventIds[j],
+          &valuesSize,
+          values.data()));
+      CHECK_EQ(numInstances, valuesSize / sizeof(uint64_t));
```
nicolasvasilache: If these CHECKs get hit, consider throwing instead

src/core/cuda/cuda_rtc.cc
```diff
@@ -203,4 +204,64 @@ Duration CudaRTCFunction::Launch(
...
+    const std::array<size_t, 3>& grid,
+    const std::array<size_t, 3>& block,
+    unsigned int shared_mem,
+    cudaStream_t stream,
+    std::vector<int> params,
+    std::vector<void*> outputs,
+    std::vector<const void*> inputs) const {
+  int dev;
+  TC_CUDA_RUNTIMEAPI_ENFORCE(cudaGetDevice(&dev));
+  if (perGpuModule_.count(dev) == 0) {
```
nicolasvasilache: Factor out and reuse across functions plz

include/tc/core/execution_engine-inl.h
```diff
@@ -155,6 +155,42 @@ Duration ExecutionEngine<ExecutorType>::run(
   return res;
 }
 
+// Steal the executor instance and give it back under lock.
+// Run outside of lock on owning ExecutorType.
+template <typename ExecutorType>
+typename ExecutorType::ProfilingInfoType ExecutionEngine<ExecutorType>::profile(
```
nicolasvasilache: Let's please definitely **not** copy-paste this piece as it involves tricky concurrent code.
Let's instead factor out the code in another function and pass it a std::function since the only difference seems to be to call ```run``` vs ```profile```.

include/tc/core/execution_engine.h
```diff
@@ -72,10 +72,15 @@ class ExecutionEngine {
       std::function<bool(const ExecutorType*)> pruningFunction =
           [](const ExecutorType*) { return false; });
 
-  /// "Low-latency" execution mode in which we just propagate raw pointers to
-  /// data in GPU address space.
-  /// No tensor-related information can be checked so it is the user's
-  /// responsibility to ensure that shapes and strides match.
+  typename ExecutorType::ProfilingInfoType profile(
```
nicolasvasilache: This is likely going to be non-core team facing API, please document the function

src/core/CMakeLists.txt
```diff
@@ -156,15 +156,21 @@ if (WITH_CUDA)
...
+      ${LLVM_INCLUDE_DIRS} 
+      ${CUDA_TOOLKIT_ROOT_DIR}/extras/CUPTI/include)
   target_link_libraries(
     tc_cuda
 
     ${CUDA_CUDA_LIBRARIES}
     ${CUDA_curand_LIBRARY}
     ${CUDA_LIBRARIES}
     ${CUDA_NVRTC_LIBRARIES}
+    ${CUDA_cupti_LIBRARY}
```
skimo-openhub: `CUDA_cupti_LIBRARY` doesn't seem to be defined anywhere.

I get
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
CUDA_cupti_LIBRARY (ADVANCED)
    linked by target "tc_cuda" in directory /home/skimo/git/c2isl/src/core
```

ttheodor: `find_package(cuda)` should have set it. Is there a cupti directory under `cuda_root_dir/extras` in your machine?

skimo-openhub:
On Tue, Apr 03, 2018 at 08:29:43AM -0700, Theodoros Theodoridis wrote:
> ttheodor commented on this pull request.
> >    target_link_libraries(
>      tc_cuda
>  
>      ${CUDA_CUDA_LIBRARIES}
>      ${CUDA_curand_LIBRARY}
>      ${CUDA_LIBRARIES}
>      ${CUDA_NVRTC_LIBRARIES}
> +    ${CUDA_cupti_LIBRARY}
> 
> `find_package(cuda)` should have set it. Is there a cupti directory under `cuda_root_dir/extras` in your machine?

No, I don't have that on any of my machines.
Do I need to install something extra?
Then please update the documentation in this commit.

skimo

src/core/CMakeLists.txt
ttheodor: You shouldn't have to install it separately (I didn't).
```
The CUPTI library is supported on all platforms supported by the CUDA Toolkit, and is available on the CUDA Downloads page as part of the CUDA Tools SDK.
```

ttheodor:
You shouldn't have to install it separately (I didn't).
```
The CUPTI library is supported on all platforms supported by the CUDA Toolkit, and is available on the CUDA Downloads page as part of the CUDA Tools SDK.
```

skimo-openhub:
On Tue, Apr 03, 2018 at 08:29:43AM -0700, Theodoros Theodoridis wrote:
> ttheodor commented on this pull request.
> >    target_link_libraries(
>      tc_cuda
>  
>      ${CUDA_CUDA_LIBRARIES}
>      ${CUDA_curand_LIBRARY}
>      ${CUDA_LIBRARIES}
>      ${CUDA_NVRTC_LIBRARIES}
> +    ${CUDA_cupti_LIBRARY}
> 
> `find_package(cuda)` should have set it. Is there a cupti directory under `cuda_root_dir/extras` in your machine?

Do you mean `find_package(CUDA REQUIRED)`?
That's conditioned on `WITH_CUDA` and I'm building with `-DWITH_CUDA=0`.
Where exactly does it set `CUDA_cupti_LIBRARY` and what is it looking
for exactly?

skimo

ttheodor:
Yes, I meant `find_package(CUDA REQUIRED)`.  `find_package(CUDA REQUIRED)` will use `FindCUDA.cmake` which is shipped together with cmake. `FindCUDA.cmake` searches for all CUDA related libraries and sets the appropriate variables (such as `CUDA_cupti_LIBRARY`).

skimo-openhub:
On Tue, Apr 03, 2018 at 04:07:55PM +0000, Theodoros Theodoridis wrote:
> You shouldn't have to install it separately (I didn't).
> ```
> The CUPTI library is supported on all platforms supported by the CUDA Toolkit, and is available on the CUDA Downloads page as part of the CUDA Tools SDK.
> ``` 

So I should install this "CUDA Tools SDK"?

skimo

ttheodor:
`CUDA Tools SDK` is the standard thing you get when you install cuda with your package manager which is probably already installed in your system. By the way, what cmake version are you using?

nicolasvasilache:
@skimo-openhub if you are building with ```WITH_CUDA=0 ... ./build.sh``` (not ```-DWITH_CUDA=0```, not sure if there is a difference) then you shouldn't see this. 
I suspect an issue in your build.
I don't think CUPTI will install if you have no physical GPU even though the SDK itself will install fine for cross compilation.

skimo-openhub:
On Wed, Apr 04, 2018 at 12:10:50AM -0700, Nicolas Vasilache wrote:
> @skimo-openhub if you are building with ```WITH_CUDA=0 ... ./build.sh``` (not ```-DWITH_CUDA=0```, not sure if there is a difference) then you shouldn't see this. 

Thanks.  That works on the machine without a card.

> I suspect an issue in your build.
> I don't think CUPTI will install if you have no physical GPU even though the SDK itself will install fine for cross compilation.

I get the same error on the machine with a card.
I can run `test.sh` just fine on this machine on master,
but I can't configure it on `cupti`.

skimo

skimo-openhub:
On Tue, Apr 03, 2018 at 11:20:36AM -0700, skimo-openhub wrote:
> On Tue, Apr 03, 2018 at 05:18:59PM +0000, Theodoros Theodoridis wrote:
> > `CUDA Tools SDK` is the standard thing you get when you install cuda with your package manager which is probably already installed in your system.
> 
> Then it doesn't include the CUPTI thing.
> (I'm on Ubuntu 17.04)

For the machine with a card, it's Ubuntu 17.10

> > By the way, what cmake version are you using?
> 
> cmake version 3.7.2

and cmake version 3.9.1

skimo

nicolasvasilache:
@caffe2bot  retest this please

src/core/cuda/cuda_profile.cc
```diff
@@ -0,0 +1,465 @@
...
+double metricValueAsDouble(const CudaMetric& metric) {
+  auto valueKind = getValueKind(metric);
+  if (valueKind == CUPTI_METRIC_VALUE_KIND_DOUBLE) {
+    return metric.value.metricValueDouble;
+  } else if (valueKind == CUPTI_METRIC_VALUE_KIND_PERCENT) {
+    return metric.value.metricValuePercent;
+  } else {
+    CHECK(false) << "Invalid metric value conversion.";
+  }
+}
```
skimo-openhub: These functions result in warnings about control reaching
end of non-void function.

skimo-openhub:
On Tue, Apr 03, 2018 at 05:39:57PM +0200, Sven Verdoolaege wrote:
> On Tue, Apr 03, 2018 at 08:29:43AM -0700, Theodoros Theodoridis wrote:
> > ttheodor commented on this pull request.
> > >    target_link_libraries(
> >      tc_cuda
> >  
> >      ${CUDA_CUDA_LIBRARIES}
> >      ${CUDA_curand_LIBRARY}
> >      ${CUDA_LIBRARIES}
> >      ${CUDA_NVRTC_LIBRARIES}
> > +    ${CUDA_cupti_LIBRARY}
> > 
> > `find_package(cuda)` should have set it. Is there a cupti directory under `cuda_root_dir/extras` in your machine?
> 
> No, I don't have that on any of my machines.
> Do I need to install something extra?

Installing the package `libcupti-dev` solved the problem for me.

skimo

src/core/cuda/cuda_profile.cc
ttheodor: The compiler can't figure out that `CHECK(false)` always leads to program termination. I could return dummy values to make the compiler happy.

ttheodor: Or use something like `__builtin_unreachable`

skimo-openhub: Please do.

skimo-openhub:
On Wed, Apr 04, 2018 at 09:58:34AM +0000, Theodoros Theodoridis wrote:
> ttheodor commented on this pull request.
> > +  size_t valueKindSize = sizeof(valueKind);
> +  TC_CUPTI_CHECK(cuptiMetricGetAttribute(
> +      metric.id, CUPTI_METRIC_ATTR_VALUE_KIND, &valueKindSize, &valueKind));
> +  return valueKind;
> +}
> +
> +double metricValueAsDouble(const CudaMetric& metric) {
> +  auto valueKind = getValueKind(metric);
> +  if (valueKind == CUPTI_METRIC_VALUE_KIND_DOUBLE) {
> +    return metric.value.metricValueDouble;
> +  } else if (valueKind == CUPTI_METRIC_VALUE_KIND_PERCENT) {
> +    return metric.value.metricValuePercent;
> +  } else {
> +    CHECK(false) << "Invalid metric value conversion.";
> +  }
> +}
> 
> The compiler can't figure out that `CHECK(false)` always leads to program termination. I could return dummy values to make the compiler happy. 

Please do.

skimo

src/core/cuda/cuda_profile.cc
```diff
@@ -116,7 +116,11 @@ void CUPTIAPI bufferCompleted(
   // since we launched only 1 kernel, we should have only 1 kernel record
   TC_CUPTI_CHECK(cuptiActivityGetNextRecord(buffer, validSize, &record));
 
+#if (CUPTI_API_VERSION >= 10)
```
skimo-openhub: This test doesn't seem to match the commit message.
You're doing something special for versions "> 9.0" not "< 9.0".
I don't know which of the two is correct.
However, you seem to be modifying new code, so this should be squashed in.
There is no point in introducing known broken commits.

ttheodor: The commit message refers to CUDA version. The check is on CUPTI version (their versions are tied but they use different numbers).

ttheodor: I'll squash it in the right commit.

skimo-openhub:
On Wed, Apr 04, 2018 at 02:46:15PM +0000, Theodoros Theodoridis wrote:
> ttheodor commented on this pull request.
> > @@ -116,7 +116,11 @@ void CUPTIAPI bufferCompleted(
>    // since we launched only 1 kernel, we should have only 1 kernel record
>    TC_CUPTI_CHECK(cuptiActivityGetNextRecord(buffer, validSize, &record));
>  
> +#if (CUPTI_API_VERSION >= 10)
> 
> The commit message refers to CUDA version. The check is on CUPTI version (their versions are tied but they use different numbers).

Thanks.  I missed that.

skimo

PR #201, opened 2018-03-23
Improve the build script

apaszke:
- Make it less verbose by default (e.g. avoid printing commands, including the output of `stat` for all built files at every run)
- Get rid of a ton of "Dependee * is newer than the depender *" messages in non-verbose mode (bug in CMake)
- Don't force rebuilds of TC every time the script is ran (this was caused by unnecessarily copying CUB includes at every run)

This brings down the time needed for a no-op rebuild (no changes) from over 20s (with ccache set up!!) to less than 2s, drastically reducing the length of a single iteration on the code.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,5 @@
 #! /bin/bash
-set -ex
+set -e
 
 export TC_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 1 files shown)


prigoyal:
thanks @apaszke for improvements. PR passes all builds and merging it now.

PR #203, opened 2018-03-23
Add TC format (a pretty printer for AST)

apaszke:
This currently doesn't handle `where`, `let`, range constraints, and `<=>`, but this can be added later if needed.


apaszke:
Please don't merge yet! I've found a small bug!

apaszke:
Ok, should be fixed now!

Issue #162, opened 2018-03-15
Improve SIGINT/SIGTERM during autotuning


ttheodor:
When the [signal handling logic](https://github.com/facebookresearch/TensorComprehensions/blob/dev/src/autotuner/genetic_autotuner.cc#L151) receives a SIGINT/SIGTERM it dumps the compilation caches, waits for the tuning sessions to finish and then throws an exception.

Maybe something got lost during refactoring but the current logic is not very useful: 
- The caches are stored after autotuning even if a signal was not raised.
- If the user has to wait until autotuning exits normally why is an exception thrown.
- Not being to immediately stop the autotuner with a SIGTERM is annoying.

@prigoyal what are the constraints imposed by the python bindings?

prigoyal:
thanks @ttheodor, let me repro and look into python side for this. I'll report back. Right now, the autotuning is indeed stopped once the signal is caught and only after the current generation has finished running.

the constraint from the python side is that abort in autotuning as we had earlier would kill python interpreter. so we rather throw exception

ttheodor:
The behavior I observed is that it only stops after all remaining generations have finished (which is obviously not the intended functionality).

prigoyal:
Hi @ttheodor, thank you for the insight. It should not wait for all generations to finish. Perhaps something got broken in the code logic. are you on master branch or dev branch?

ttheodor:
On dev, can you confirm that it works on your machine?

ftynse:
seems to work for me as expected (abort at the end of the current generation)
```
[ RUN      ] ATenCompilationUnitTest.LayerNorm
Generation 0    Jobs(Compiled, GPU)/total  (8, 2)/8   (best/median/worst)us: 64/75/75^CAutotuning aborted.
No filepath provided, not saving cache
Generation 0    Jobs(Compiled, GPU)/total  (8, 8)/8   (best/median/worst)us: 62/75/132
unknown file: Failure
C++ exception with description "Abort requested" thrown in the test body.
[  FAILED  ] ATenCompilationUnitTest.LayerNorm (13082 ms)
```

ttheodor:
Can you run it it with ```--smoke_check=0 --gtest_filter='*TensorDot' --tuner_threads=8``` (choose a high number of threads).

ftynse:
works too, but it takes a non-negligible amount of time to finish the generation
```
[...]

PR #206, opened 2018-03-24
Genetic tuner: std::abort on SIGTERM

ttheodor:
Adresses #162

### DIFF
--- a/src/autotuner/genetic_autotuner.cc
+++ b/src/autotuner/genetic_autotuner.cc
@@ -83,6 +83,11 @@ std::vector<MappingOptions> GeneticAutotuner::load(
   return tc::autotune::restoreCandidates(tcName, inputs, outputs);
 }
 
+namespace {
+volatile std::sig_atomic_t sigint_ = 0;
+volatile std::sig_atomic_t sigterm_ = 0;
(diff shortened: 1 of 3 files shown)


src/autotuner/genetic_autotuner.cc
```diff
@@ -148,15 +153,23 @@ llvm::Optional<MappingOptions> GeneticAutotuner::tune(
   });
   while (not tunerFinished) {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
-    if (killRequested_) {
+    if (sigint_) {
+      std::cerr
+          << "Autotuning will stop after the current generation has finished."
+          << std::endl;
+      tuner.stopAfterCurrentGeneration();
+      tunerThread.join();
```
prigoyal: Hi @ttheodor , I am wondering what will happen in the case where: autotuning was running , sigint was caught, I want to do another autotuning. In that case, it seems like sigint_ will be still set and autotuning will not run. Is that correct? Thanks

prigoyal: thanks a lot @ttheodor , sounds great. :)

Issue #109, opened 2018-03-06
Define TC lang conventions to standardize tc defs


prigoyal:
we are seeing TC's get written in ways that can be confusing to look at. Nicolas has great idea on defining the TC lang convention. Feel free to add the .md file for it :)

PR #215, opened 2018-03-26
TC coding guide

nicolasvasilache:
This introduces a coding style guide for TC and addresses #109.
It then goes on to update various places in the codebase where TCs are hard coded.

### DIFF
--- a/CodingConventions.md
+++ b/CodingConventions.md
@@ -594,4 +594,4 @@ Coding Conventions for writing Tensor Comprehensions
 
 Please see the following documentation
 [entry](https://facebookresearch.github.io/TensorComprehensions/coding_conventions.html)
-on how to write Tensor Comprehensions in a standard, legible, fashion.
+on how to write Tensor Comprehensions in a standard legible fashion.
--- a/CodingConventions.md
+++ b/CodingConventions.md
@@ -588,3 +588,10 @@ changes) while working on a feature and even in "WIP" pull requests,
 as long as the pieces are recombined (e.g., through an interactive rebase)
 into logical units when the feature is ready for merging.
 Force-pushing in PR branches is fine.
+
+Coding Conventions for writing Tensor Comprehensions
+====================================================
+
+Please see the following documentation
+[entry](https://facebookresearch.github.io/TensorComprehensions/coding_conventions.html)
+on how to write Tensor Comprehensions in a standard, legible, fashion.
--- a/docs/doxygen/index.md
+++ b/docs/doxygen/index.md
@@ -13,7 +13,7 @@ with a few basic functionalities.
 
 Tensor Comprehension Notation
 -----------------------------
-TC borrow three ideas from Einstein notation that make expressions concise:
+TC borrows three ideas from Einstein notation that make expressions concise:
 
 1. Loop index variables are defined implicitly by using them in an expression and their range is aggressively inferred based on what they index.
 2. Indices that appear on the right of an expression but not on the left are assumed to be reduction dimensions.
@@ -21,22 +21,22 @@ TC borrow three ideas from Einstein notation that make expressions concise:
 
 Let's start with a simple example is a matrix vector product:
 
-    def mv(float(R,C) A, float(C) B) -> (o) {
-      o(i) +=! A(i,j) * B(j)
+    def mv(float(R,C) A, float(C) x) -> (o) {
+        o(r) +=! A(r,r_c) * x(r_c)
     }
 
 `A` and `x` are input tensors. `o` is an output tensor.
-The statement `o(i) += A(i,j) * b(j)` introduces two index variables `i` and `j`.
-Their range is inferred by their use indexing `A` and `B`. `i = [0,R)`, `j = [0,C)`.
-Because `j` only appears on the right side,
-stores into `o` will reduce over `j` with the reduction specified for the loop.
+The statement `o(r) +=! A(r,r_c) * x(r_c)` introduces two index variables `r` and `r_c`.
+Their range is inferred by their use indexing `A` and `x`. `r = [0,R)`, `r_c = [0,C)`.
+Because `r_c` only appears on the right side,
+stores into `o` will reduce over `r_c` with the reduction specified for the loop.
 Reductions can occur across multiple variables, but they all share the same kind of associative reduction (e.g. +=)
 to maintain invariant (3). `mv` computes the same thing as this C++ loop:
 
     for(int i = 0; i < R; i++) {
       o(i) = 0.0f;
       for(int j = 0; j < C; j++) {
-        o(i) += A(i,j) * B(j);
+        o(i) += A(i,j) * x(j);
       }
     }
 
@@ -50,7 +50,7 @@ We provide a few basic examples.
 **Simple matrix-vector**:
 
     def mv(float(R,C) A, float(C) B) -> (o) {
-      o(i) += A(i,j) * B(j)
+        o(r) +=! A(r,r_c) * B(r_c)
     }
 
 **Simple matrix-multiply:**
@@ -59,21 +59,20 @@ Note the layout for B is transposed and matches the
 traditional layout of the weight matrix in a linear layer):
 
     def mm(float(X,Y) A, float(Y,Z) B) -> (R) {
-      R(i,j) += A(i,j) * B(j,k)
+        R(x,z) +=! A(x,r_y) * B(r_y,z)
     }
 
 **Simple 2-D convolution (no stride, no padding):**
 
     def conv(float(B,IP,H,W) input, float(OP,IP,KH,KW) weight) -> (output) {
-      output(b, op, h, w) += input(b, ip, h + kh, w + kw) * weight(op, ip, kh, kw)
+        output(b, op, h, w) +=! input(b, r_ip, h + r_kh, w + r_kw) * weight(op, r_ip, r_kh, r_kw)
     }
 
 **Simple 2D max pooling:**
 
-Note the similarity with a convolution with a
-"select"-style kernel):
+Note the similarity with a convolution with a "select"-style kernel:
 
     def maxpool2x2(float(B,C,H,W) input) -> (output) {
-      output(b,c,i,j) max= input(b,c,2*i + kw, 2*j + kh)
-        where kw = [0, 2[, kh = [0, 2[
+        output(b,c,h,w) max=! input(b,c,2*h + r_kw, 2*w + r_kh)
+            where r_kw in 0:2, r_kh in 0..2
     }
--- a/docs/doxygen/index.md
+++ b/docs/doxygen/index.md
@@ -28,7 +28,7 @@ Let's start with a simple example is a matrix vector product:
 `A` and `x` are input tensors. `o` is an output tensor.
 The statement `o(r) +=! A(r,r_c) * x(r_c)` introduces two index variables `r` and `r_c`.
 Their range is inferred by their use indexing `A` and `x`. `r = [0,R)`, `r_c = [0,C)`.
-Because `r_c` only appears on the right side,
+Because `r_c` only appears on the righthand side,
 stores into `o` will reduce over `r_c` with the reduction specified for the loop.
 Reductions can occur across multiple variables, but they all share the same kind of associative reduction (e.g. +=)
 to maintain invariant (3). `mv` computes the same thing as this C++ loop:
--- a/docs/source/coding_conventions.rst
+++ b/docs/source/coding_conventions.rst
@@ -3,7 +3,7 @@ Coding Conventions
 
 In order to increase readability across Tensor Comprehensions written by
 multiple authors and to reduce the amount of surprising behavior, the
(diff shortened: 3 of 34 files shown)


nicolasvasilache:
@caffe2bot retest this please

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+    def matmul(float(M, K) L, float(K, M) U) -> (LU) {
+        LU(m1, m2) +=! L(m1, r_k) * U(r_k, m2) where r_k in m1:M, r_k in 0:m2+1
+    }
+
+The reader may remark that this is an inefficient way of performing
+matrix-multiplication of triangular matrices.
+Lowering such operations efficient from TC is the subject of future work.
+
+Prefix gradient tensors names with :code:`g_`
+---------------------------------------------
```
apaszke: I can of course make AD follow this pattern, however, provided we disallow mutating inputs, this is not the whole story. Consider a variant of matmul that also has a second output which is the doubled result:
```
def mm(double(M, K) A, double(K, N) B) -> (O, W) {
  O(i, j) +=! A(i, k) * B(k, j)
  W(i, j) = O(i, j) * 2
}
```

So far so good, however if you want to apply reverse mode AD to it, you will need to add gradients w.r.t. W to the gradient w.r.t. O:
```
def grad_mm(double(M, K) A, double(K, N) B,
            double(M, N) O, double(M, N) W,
            double(M, N) seed_d_O, double(M, N) seed_d_W) -> (d_A, d_B) {
  d_O(i, j) = seed_d_d_O(i, j)
  d_O(i, j) += (seed_d_W(i, j) * 2)
  d_A(i, k) = 0
  d_A(i, k) += (d_O(i, j) * B(k, j))
  d_B(k, j) = 0
  d_B(k, j) += (d_O(i, j) * A(i, k))
}
```

Here, you can see that the arguments are gradient "seeds" (this is the name used sometimes in the literature IIRC), and might not reflect the real gradients of a particular value alone, as can be seen in the case of O. On the other hand, is true that the seed is the real gradient of W, which is why my code skips instantiating `d_W`, and uses the seed alone.

If you're writing down the conventions for gradient TCs then it would be good to take that into account. Otherwise, if you're fine with what I did for AD, I can just call them `seed_g_<name>`.

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+
+to:
+
+.. code::
+
+    def matmul(float(M, K) A, float(K, N) B) -> (C) {
+        C(m, n) +=! A(m, k) * B(k, n)
+    }
+
+Filter non-rectangular regions with deta-dependencies
```
skimo-openhub: data

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+
+.. code::
+
+    def matmul(float(M, K) L, float(K, M) U) -> (LU) {
+        LU(m1, m2) +=! L(m1, r_k) * U(r_k, m2) where r_k in m1:M, r_k in 0:m2+1
+    }
+
+The reader may remark that this is an inefficient way of performing
+matrix-multiplication of triangular matrices.
+Lowering such operations efficient from TC is the subject of future work.
```
skimo-openhub: efficiently

CodingConventions.md
```diff
@@ -588,3 +588,10 @@ changes) while working on a feature and even in "WIP" pull requests,
 as long as the pieces are recombined (e.g., through an interactive rebase)
 into logical units when the feature is ready for merging.
 Force-pushing in PR branches is fine.
+
+Coding Conventions for writing Tensor Comprehensions
+====================================================
+
+Please see the following documentation
+[entry](https://facebookresearch.github.io/TensorComprehensions/coding_conventions.html)
+on how to write Tensor Comprehensions in a standard, legible, fashion.
```
ftynse: I'd remove commas here.

docs/doxygen/index.md
```diff
@@ -13,30 +13,30 @@ with a few basic functionalities.
...
     }
 
 `A` and `x` are input tensors. `o` is an output tensor.
-The statement `o(i) += A(i,j) * b(j)` introduces two index variables `i` and `j`.
-Their range is inferred by their use indexing `A` and `B`. `i = [0,R)`, `j = [0,C)`.
-Because `j` only appears on the right side,
-stores into `o` will reduce over `j` with the reduction specified for the loop.
+The statement `o(r) +=! A(r,r_c) * x(r_c)` introduces two index variables `r` and `r_c`.
+Their range is inferred by their use indexing `A` and `x`. `r = [0,R)`, `r_c = [0,C)`.
+Because `r_c` only appears on the right side,
```
ftynse: right _hand_ side?

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
+Coding Conventions
+==================
+
+In order to increase readability across Tensor Comprehensions written by
+multiple authors and to reduce the amount of surprising behavior, the
+following conventions should be adopted when writing TC. Generally in TC one
```
ftynse: Comma after "Generally in TC"

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+should increment nesting by 4 whitespaces at each level and align tensor names
```
ftynse: mention that we use spaces and not tabs
(never miss an opportunity to start a holy war)

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+and indices where appropriate to make memory access patterns emerge. Since
```
ftynse: does this mean that inline whitespace is encouraged?  mention this specifically?

Is this example what you want?  Do we encourage inline whitespace on both LHS and RHS?
```
A(i,j+1) += B(i, j   , k)
C(i,j)   += B(i, j-1, k)
```

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+these two goals can easily be conflicting, use your best judgement to tradeoff
+between the two goals. Such examples are provided below.
+
+Use indices named after parameters
+----------------------------------
+
+Use upper-case names for parameters and input/output tensors.
```
ftynse: `output`, `input` and many other tensor names in the examples do not use upper case.  I suppose we don't want to  uppercase the entire name, just to capital case it.

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+Use lower-case names for indices to match the name of the parameter
+corresponding to the dimension upon which they iterate.
```
ftynse: What if we have 
```
def (float(N) X, float(M) Y) -> (Z) {
  Z(xxx) = X(xxx) + Y(xxx)
```
what should we use instead of `xxx`?

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+requirement to make range inference non-ambiguous. To simulate non-rectangular
+iteration spaces, one can use the following:
+
+.. code::
+
+    def matmul(float(M, K) L, float(K, M) U) -> (LU) {
+        LU(m1, m2) +=! (r_k >= m1 and r_k =< m2) ? L(m1, r_k) * U(r_k, m2) : 0
+    }
+
+However, the following is incompatible with range inference and will fail
```
ftynse: Can we formalize "the following" here?  I don't really like restrictions-by-simple-example.  It seems that "using iteration variables in `where` expression is forbidden" is what you want. 

Also, this is a _semantic_ restriction and not a _style_ recommendation.

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+the semantic checks in the TC compiler:
+
+.. code::
+
+    def matmul(float(M, K) L, float(K, M) U) -> (LU) {
+        LU(m1, m2) +=! L(m1, r_k) * U(r_k, m2) where r_k in m1:M, r_k in 0:m2+1
+    }
+
+The reader may remark that this is an inefficient way of performing
```
ftynse: Why inefficient?  How this is relevant to the style?  A code can perfectly respect the coding guidelines yet remain inefficient...

docs/source/coding_conventions.rst
ftynse: I'd prefer `grad_` over `g_`.  Otherwise it smells like the Hungarian notation, and that is a questionable choice.

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+matrix-multiplication of triangular matrices.
+Lowering such operations efficient from TC is the subject of future work.
+
+Prefix gradient tensors names with :code:`g_`
+---------------------------------------------
+
+When implementing backward operations, pass the inputs to the backwards pass
+in the same order as the outputs to the forward passs and use the same tensor
```
ftynse: outputs _of_ the forward pass?

docs/source/coding_conventions.rst
```diff
@@ -0,0 +1,119 @@
...
+name prefixed by :code:`g_`. For instance:
+
+.. code::
+
+     def conv(float(N,C,H,W) I, float(M,C,KH,KW) Wt) -> (O) {
+         ...
+     }
+
+     def conv_bw(float(N,C,H,W) I, float(M,C,KH,KW) Wt, float(N,M,HO,WO) g_O) -> (g_I) {
```
ftynse: shouldn't this be called `g_conv` as per the previous item?

docs/source/framework/pytorch_integration/autograd_with_tc.rst
```diff
@@ -27,11 +27,11 @@ Examples
...
      CONV_LANG = """
      def convolution(float(N,C,H,W) I, float(M,C,KH,KW) W1) -> (O) {{
-        O(n, m, h, w) +=! I(n, c, {sh} * h + kh, {sw} * w + kw) * W1(m, c, kh, kw)
+       O(n, m, h, w) +=! I(n, r_c, {sh} * h + r_kh, {sw} * w + r_kw) * W1(m, r_c, r_kh, r_kw)
      }}
-     def convolution_grad(float(N,C,H,W) I, float(M,C,KH,KW) W1, float(N,M,H,W) O_grad) -> (I_grad, W1_grad) {{
-        I_grad(n, c, h, w) +=! O_grad(n, m, {sh} * h - kh, {sw} * w - kw) * W1(m, c, kh, kw)
-        W1_grad(m, c, kh, kw) +=! O_grad(n, m, {sh} * h - kh, {sw} * w - kw) * I(n, c, h, w)
+     def convolution_grad(float(N,C,H,W) I, float(M,C,KH,KW) W1, float(N,M,H,W) g_O) -> (g_I, g_W1) {{
+        g_I(n, c,  h,  w) +=! g_O(  n, r_m, {sh} *   h - r_kh, {sw} *   w - r_kw) * W1(r_m, c, r_kh, r_kw)
```
ftynse: we never said that we prefer aligning

```
  X(i, j)
FOO(m,n)
```

over
```
X  (i,j)
FOO(i,j)
```

docs/source/coding_conventions.rst
nicolasvasilache: How about @apaszke's ```d_```? It is the actual mathematical notation so that should settle it.

docs/source/coding_conventions.rst
nicolasvasilache: unspecified, use your best judgement, I'd certainly not limit whitespacing to only RHS.

docs/source/coding_conventions.rst
nicolasvasilache: no particular request on the function name on my end

docs/source/framework/pytorch_integration/autograd_with_tc.rst
nicolasvasilache: we didn't, use your best judgement, do you see an issue with this alignment?

docs/source/coding_conventions.rst
nicolasvasilache: Whatever? 
I would go for ```n``` if ```min(n, m) == n``` or ```m``` if ```min(n, m) == m``` statically

docs/source/coding_conventions.rst
nicolasvasilache: We can specifically mention that it is a "semantic restriction" and point to the proper place that discusses inference for more details.
Still I think i wouldn't hurt to also surface it in the coding conventions but we can always move it to another section in the docs.

docs/source/coding_conventions.rst
nicolasvasilache: same comment as above, please let me know if you want me to move to another section in the docs

docs/source/coding_conventions.rst
nicolasvasilache: @apaszke the seed stuff lgtm, no need to change IMO

PR #216, opened 2018-03-26
Fix tc_lang runtime error in FBcode

mingzhe09088:
This diff fixed a runtime error with tc_lang which was caused by incorrect file path. The added "ROOT_PATH" will be passed by the buck TARGET.

### DIFF
--- a/test/test_lang.cc
+++ b/test/test_lang.cc
@@ -29,10 +29,10 @@
 using namespace lang;
 
 #ifdef ROOT_PATH
-  const std::string expected_file_path =
-      std::string(ROOT_PATH) + "src/lang/test_expected/";
+const std::string expected_file_path =
+    std::string(ROOT_PATH) + "src/lang/test_expected/";
 #else
-  const std::string expected_file_path = "src/lang/test_expected/";
+const std::string expected_file_path = "src/lang/test_expected/";
 #endif
 
 static inline void barf(const char* fmt, ...) {
--- a/test/test_lang.cc
+++ b/test/test_lang.cc
@@ -29,7 +29,8 @@
 using namespace lang;
 
 #ifdef ROOT_PATH
-  const std::string expected_file_path = std::string(ROOT_PATH) + "src/lang/test_expected/";
+  const std::string expected_file_path =
+      std::string(ROOT_PATH) + "src/lang/test_expected/";
 #else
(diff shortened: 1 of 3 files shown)


test/test_lang.cc
```diff
@@ -28,7 +28,12 @@
 
 using namespace lang;
 
+#ifdef ROOT_PATH
```
nicolasvasilache: We have been using ```TC_DIR``` across our build for this.
How about removing the case disjunction which is going to be always painful to maintain and got for something similar to ```CUDA_HOME```?
In other words:
1. ```const std::string expected_file_path = std::string(TC_DIR) + "src/lang/test_expected/";``` always,
2. ```TC_DIR``` is exported both in fbcode and CMakeLists (you can rename all occurrences to ```TC_ROOT_DIR``` in a separate commit if you prefer to have ```ROOT``` in the name),
3. have a ```#ifndef TC_ROOT_DIR``` trigger an error to ensure errors at compile time rather than runtime 

Also, let's take the opportunity to move all those expected test files in ```test/expected``` rather than keeping them polluting the source tree.

nicolasvasilache:
cc salexpb

mingzhe09088:
@nicolasvasilache I've updated my diff and its working on FBcode. On OSS side, TC_DIR is not defined. To define TC_DIR on OSS side, I need to add -DTC_DIR. Should I add it to build.sh or CMakefile?

prigoyal:
Hi @mingzhe09088, I think adding `TC_DIR` to build.sh should suffice. Thanks so much for your efforts. :)

nicolasvasilache:
@mingzhe09088 TC_DIR is defined in build.sh already. 
You should additionally pass is to CMakeLists.txt and have it defined in the build rules to emit a ```gcc ... -DTC_DIR=/path ...```
You can just to the build subdir and run say ```VERBOSE=1 make -j 4``` to see the compilation commands emitted by CMake

Atm the CI above errors out with:
```
/root/TensorComprehensions/test/test_lang.cc:32:2: error: #error "TC_DIR must be defined"
 #error "TC_DIR must be defined"
```

Which is exactly the problem; you can also take a look at ```git grep CUDA_HOME``` to see how this propagated in the CMakeLists.txt.

mingzhe09088:
@prigoyal Could you take a look at the fix and merge it?

PR #219, opened 2018-03-26
Parallel CPU mapper

### LINKED ISSUE #34: Determine whether a loop should be parallelized in CPU backend from ISL
ISL internally dictates if loops should be run serially, or may be run in parallel. Presently we are conservative and always run loops serially (for CPU). We ought use the information from ISL to set the CPU parallelization flag appropriately for each loop.

wsmoses:
We now properly generate parallel cpu backend code. This requires the ISL patch to go in first.

Resolves #34


test/test_mapper_llvm.cc
```diff
@@ -66,6 +65,176 @@ def fun(float(N, M) A, float(N, M) B) -> (C) {
   checkRtol(Cc - C, {A, B}, N * M);
 }
 
+TEST(LLVMCodegen, BasicParallel) {
+  string tc = R"TC(
+def fun(float(N, M) A, float(N, M) B) -> (C) {
+  C(i, j) = A(i, j) + B(i, j)
```
nicolasvasilache: we have TC coding style in #215 could you please update so this PR is future proof :)

test/test_mapper_llvm.cc
```diff
@@ -66,6 +65,176 @@ def fun(float(N, M) A, float(N, M) B) -> (C) {
...
+
+body.i:                                           ; preds = %__cilk_sync.exit
+  call void @__cilkrts_leave_frame(%struct.__cilkrts_stack_frame* nonnull %__cilkrts_sf) #0
+  br label %__cilk_parent_epilogue.exit
+
+__cilk_parent_epilogue.exit:                      ; preds = %__cilk_sync.exit, %body.i
+  ret void
+}
+)LLVM";
+  EXPECT_EQ(correct_llvm, toString(mod->getFunction("kernel_anon")));
```
nicolasvasilache: This seems overly ambitious :), care to relax a bit?
Also, we only use vanilla LLVM internally atm so this is not a portable test, can you please separate test_mapper_llvm from test_mapper_tapir which should only compile conditionally?

test/test_mapper_llvm.cc
```diff
@@ -66,6 +65,176 @@ def fun(float(N, M) A, float(N, M) B) -> (C) {
...
+  auto N = 40;
+  auto M = 24;
+
+  auto ctx = isl::with_exceptions::globalIslCtx();
+  auto scop = polyhedral::Scop::makeScop(ctx, tc);
+  auto context = scop->makeContext(
+      std::unordered_map<std::string, int>{{"N", N}, {"M", M}});
+  scop = Scop::makeSpecializedScop(*scop, context);
+  scop =
+      Scop::makeScheduled(*scop, SchedulerOptionsView(SchedulerOptionsProto()));
```
nicolasvasilache: Let's please rebase on top of #197, for this case you should create a CpuMappingOptions 
I am adding that in a followup because #197 is big enough and is already having enough merge conflicts as it is.

include/tc/core/mapping_options.h
```diff
@@ -295,6 +295,7 @@ class SchedulerOptionsView {
   /// Construct a view that refers to a protocol buffers message.
   SchedulerOptionsView(const SchedulerOptionsView&) = default;
   SchedulerOptionsView(SchedulerOptionsProto& buf) : proto(buf) {}
+  SchedulerOptionsView(SchedulerOptionsProto&& buf) : proto(buf) {}
```
nicolasvasilache: This does not seem like a good idea independently.
We may want to do it nonetheless in the big picture but atm I would punt on adding such extensions.

test/test_mapper_llvm.cc
nicolasvasilache: Also, why not:
```
SchedulerOptionsProto p;
SchedulerOptionsProtoView pv(p);
scop = Scop::makeScheduled(*scop, pv);
```
?

wsmoses:
Regarding ISL, this PR requires the pr/coincident in ISL to function (which has yet to be merged and not sure has been updated w c++ bindings).

skimo-openhub:
On Tue, Mar 27, 2018 at 02:14:15AM -0700, William Moses wrote:
> Regarding ISL, this PR requires the pr/coincident in ISL to function (which has yet to be merged and not sure has been updated w c++ bindings).

You need to reference the ntv_dev_cpp branch, which gets updated
_after_ merging (in this case pr/coincident) into ntv_dev.
In particular, you should mark you TC changes as WIP until
you have updated the isl submodule to the right commit
(when it materializes).

skimo

test/test_mapper_llvm.cc
wsmoses: Resolve for both. However, as after this our CPU backend starts emitting parallel code [and we start doing some scheduling for CPU], it may be wise to nudge the internal folks to add Tapir :P

wsmoses:
@caffe2bot retest this please

wsmoses:
@caffe2bot retest this please

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -492,8 +483,7 @@ class LLVMCodegen {
 
     // TODO: integrate query ISL as to whether the relevant loop ought be
     // parallelized
```
skimo-openhub: stale comment

src/core/polyhedral/codegen_llvm.cc
```diff
@@ -492,8 +483,7 @@ class LLVMCodegen {
...
-    bool parallel = false;
-
+    bool parallel = isl_ast_node_for_is_coincident(node.get());
```
skimo-openhub: Please use the C++ interface in production code.

Issue #222, opened 2018-03-27
Check and remove redundant python tests (if any) + gen=2


prigoyal:
Hello all, in light of recent discussion around test time, I will go over the python tests to make sure everything runs for 2 generations and that no tests are redundant :)

as @salexspb has accurately pointed out: the test time tuning is an ongoing process and is hard to get right but has to be adjusted :)

prigoyal:
closed by #223

PR #223, opened 2018-03-27
Bump autotune default options, use small gen=2 for test and remove redundant tests

### LINKED ISSUE #195: Increase the default number of the autotuner generations in python bindings.
https://github.com/facebookresearch/TensorComprehensions/blob/479815449a3a438a69ba21b8c2eb9bf52c594443/tensor_comprehensions/tc_unit.py#L192

By default, the autotuner in python runs 2 generations with very small population size.  This is largely insufficient, even "some check" tests run longer.

@prigoyal @nicolasvasilache @ttheodor 

- [x] bump up the defaults in tc_unit.py 
- [x] modify python unit tests to still use old settings to have faster CI testing

prigoyal:
1. Bumped the autotuner defaults as per #195 . #generations are set to 10, shall this be changed to 20?

2. remove the redundant tests

3. Set the unit test settings as 2 generations and 5 pop_size.

With 3 settings above, the time for test on my terminal is 16 min - down my 6 mins. :)

closes #195 and #222

### DIFF
--- a/tensor_comprehensions/tc_unit.py
+++ b/tensor_comprehensions/tc_unit.py
@@ -30,12 +30,14 @@ logger = logging.getLogger(__name__)
 
 
 # these are quick options for finishing autotuning
-autotuner_settings = {"threads": 32}
+autotuner_settings = {
+    "threads": 32, "generations": 2, "pop_size": 5,
+}
 
 # TC prunes autotuning for kernels which require < 256 threads. So to tune small
 # size kernels, we set the min kernel threads to 1
 small_sizes_autotuner_settings = {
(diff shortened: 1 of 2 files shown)


prigoyal:
@caffe2bot retest this please

ftynse:
Given that the autotuning can be gracefully stopped, 20 generations sounds ok. I still think we should change that based on actual observations, but that can be done later

prigoyal:
thanks @ftynse , that sounds reasonable. :)

prigoyal:
@caffe2bot retest this please

Issue #179, opened 2018-03-20
Convolution outputs gibberish values when filter size > 32, with naive mapping options


jeanfeydy:
Hi all,
When playing around with large convolution filters, I discovered an unexpected behavior: with a large filter, if the input image size exceeds 32, a standard convolution outputs gibberish results.

```python
import tensor_comprehensions as tc
import torch
from torch.autograd import Variable
from matplotlib import pyplot as plt

def two_char(s) :
    if   s    == "0.0" : return ". " 
    elif s[0] == "0"   : return s[1:]
    else :               return s[:2]

def pretty_print(t) :
    l = [two_char("{:.1f}".format(i)) for i in t.data.cpu().numpy()]
    return " ".join(l)

convolution_lang = """
def convolution(float(X) A, float(Xp) K) -> (B) {
    B(x)   +=! A(xp) * K(X-1+x-xp) where xp in 0:X
}
"""

convolution = tc.define(convolution_lang, name="convolution")

for X in [30,31,32,33,34] :
    A = torch.zeros(X).cuda()
    A[10] = 1
    A = Variable(A)

    # len(C) = 2*X-1
    C = Variable( torch.arange(-X+1,X).cuda()**2 )
    K = (-C/9).exp()

    B = convolution( A, K )

    print("X =", X, "==============================")
    print("A :\n",         pretty_print(A) )
    print("K :\n",         pretty_print(K))
    print("B = A ★ K :\n", pretty_print(B))
```

```
[WARNING]: No mapping options passed, 'naive' type mapping options will be used and will likely have bad performance. See help(your_layer.__call__) for setting mapping options.
X = 30 ==============================
A :
 .  .  .  .  .  .  .  .  .  .  1. .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  . 
K :
 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .1 .2 .4 .6 .9 1. .9 .6 .4 .2 .1 .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  .  . 
B = A ★ K :
 .  .  .  .  .  .1 .2 .4 .6 .9 1. .9 .6 .4 .2 .1 .  .  .  .  .  .  .  .  .  .  .  .  .  . 
[...]

PR #226, opened 2018-03-27
emitHalideExpr: print parentheses around compound nested expressions

skimo-openhub:
Despite what the original comment claims, the expression in the output
that corresponds to a variable in the input may not be an identifier
by itself.  Since the variable may appear in a Halide expression
at any position (e.g., as an argument to "-" or "*"), parentheses
need to be put around the expression if it is anything other than
an identifier or a non-negative number.

### DIFF
--- a/src/core/polyhedral/cuda/codegen.cc
+++ b/src/core/polyhedral/cuda/codegen.cc
@@ -371,27 +371,21 @@ void emitAccess(AFF access, const CodegenStatementContext& context) {
 } // namespace
 
 void emitCopyStmt(const CodegenStatementContext& context) {
-  using detail::emitDirectSubscripts;
-
   auto stmtId = context.statementId();
 
   auto iteratorMap = context.iteratorMap();
   auto promoted = iteratorMap.range_factor_range();
   auto original = iteratorMap.range_factor_domain().range_factor_range();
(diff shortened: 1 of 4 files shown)


skimo-openhub:
Still waiting for confirmation that it actually fixed #179.

skimo-openhub:
I removed the annotation that it fixes #179.  It should be an improvement, even if it doesn't.

src/core/polyhedral/cuda/codegen.cc
```diff
@@ -495,22 +495,33 @@ isl::multi_aff makeMultiAffAccess(
...
       auto pwAff = tc::polyhedral::detail::makeAffFromMappedExpr(
           Halide::Expr(op), context);
       // Use a temporary isl::ast_build to print the expression.
       // Ideally, this should use the build at the point
       // where the user statement was created.
       auto astBuild = isl::ast_build::from_context(pwAff.domain());
-      auto s = astBuild.expr_from(pwAff).to_C_str();
+      auto expr = astBuild.expr_from(pwAff);
+      auto s = expr.to_C_str();
+      if (!is_identifier_or_nonnegative_integer(expr))
```
nicolasvasilache: braces plz

nicolasvasilache:
@caffe2bot retest this please

PR #227, opened 2018-03-27
Initial version of reverse mode autodiff

apaszke:
There are still a few things that could have been improved, but I think this can be done in subsequent PRs, and I wanted to get some feedback at this point. Some issues I see:
1. It might be nicer to link it as a part of `lang`. On the other hand, this needs `tc2halide`, which is part of the `core`, and that would create a dependency cycle between `libtc_lang.so` and `libtc_core.so`.
2. Not sure where the Python bindings should go. It's not really part of the engine, but I don't think we have a different pybind file that fits it better. I can create a new one if you want, just let me know.

### DIFF
--- a/src/core/autodiff.cc
+++ b/src/core/autodiff.cc
@@ -1,0 +1,387 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "tc/core/autodiff.h"
+#include "tc/core/tc2halide.h"
+#include "tc/lang/parser.h"
+#include "tc/lang/sema.h"
+#include "tc/lang/tc_format.h"
+#include "tc/lang/tree_views.h"
+
+#include <unordered_set>
+#include <vector>
+
+namespace tc {
+
+using namespace lang;
+
+static const lang::SourceRange dummyRange{std::make_shared<std::string>(""),
+                                          0,
+                                          0};
+
+int32_t getTcType(Halide::Type t) {
+  if (t.is_int()) {
+    switch (t.bits()) {
+      case 64:
+        return TK_INT64;
+      case 32:
+        return TK_INT32;
+      case 16:
+        return TK_INT16;
+      case 8:
+        return TK_INT8;
+    }
+  } else if (t.is_uint()) {
+    switch (t.bits()) {
+      case 64:
+        return TK_UINT64;
+      case 32:
+        return TK_UINT32;
+      case 16:
+        return TK_UINT16;
+      case 8:
+        return TK_UINT8;
+    }
+  } else if (t.is_float()) {
+    switch (t.bits()) {
+      case 64:
+        return TK_DOUBLE;
+      case 32:
+        return TK_FLOAT;
+    }
+  }
+  throw std::runtime_error("Unknown Halide type");
+}
+
+void findAccessedTensors(
+    std::unordered_set<std::string>& read_only,
+    const TreeRef& tree) {
+  if (tree->kind() == TK_ACCESS) {
+    read_only.insert(Access(tree).name().name());
+  } else {
+    for (const TreeRef& subtree : tree->trees()) {
+      findAccessedTensors(read_only, subtree);
+    }
+  }
+}
+
+void assertNoWriteAfterRead(Def def) {
+  std::unordered_set<std::string> read_only;
+  // Inputs are always read-only
+  for (Param input : def.params())
+    read_only.insert(input.ident().name());
+  for (Comprehension comp : def.statements()) {
+    findAccessedTensors(read_only, comp.rhs());
+    auto lhs_name = comp.ident().name();
+    if (read_only.count(lhs_name) > 0)
+      throw std::runtime_error(
+          "AD not supported in TCs that write to a value after reading it");
+  }
+}
+
+void findIndexVars(
+    std::unordered_set<std::string>& index_vars,
+    const TreeRef& tree,
+    bool gather_idents) {
+  if (tree->kind() == TK_IDENT && gather_idents) {
+    index_vars.insert(Ident(tree).name());
+  } else if (tree->kind() == TK_ACCESS) {
+    for (const TreeRef& idx : Access(tree).arguments()) {
+      findIndexVars(index_vars, idx, true);
+    }
+  } else if (tree->kind() == TK_BUILT_IN) {
+    // BuiltIn holds the name of a function as an ident, so we have to skip it
+    for (const TreeRef& subtree : BuiltIn(tree).arguments()) {
+      findIndexVars(index_vars, subtree, gather_idents);
+    }
+  } else {
+    for (const TreeRef& subtree : tree->trees()) {
+      findIndexVars(index_vars, subtree, gather_idents);
+    }
+  }
+}
+
+// XXX: this is a bit of a fragile hack, and can easily break when the AST will
+// get more idents in different nodes, but it's quite simple and the worst thing
+// that can happen is that we will be too conservative and throw, so it's ok.
+std::unordered_set<std::string> usedIndexVars(Comprehension comp) {
+  std::unordered_set<std::string> index_vars;
+  for (Ident idx : comp.indices())
+    index_vars.insert(idx.name());
+  findIndexVars(index_vars, comp.rhs(), false);
+  return index_vars;
+}
+
+// This struct holds a lot of the information required to perform bookkeeping
+// of gradient values. For example:
+// - do we already have a writeable gradient for a value, or only a seed
+// - should we use the seed value, or the writeable gradient in an expression
+// - is the gradient implicitly zero now (because the value was overwritten)
+struct GradInfo {
+  GradInfo(const ListView<Param>& primal_outputs) {
+    for (const Param& output : primal_outputs) {
+      primal_outputs_.insert(output.ident().name());
+    }
+  }
+
+  void addGradComprehension(
+      Ident primal_lhs_name,
+      ListView<lang::TreeRef> lhs_indices,
+      TreeRef rhs_expr) {
+    auto lhs_name = makeGradName(primal_lhs_name);
+    if (has_writeable_grad_.count(primal_lhs_name.name()) == 0) {
+      auto rhs_expr = primal_outputs_.count(primal_lhs_name.name()) > 0
+          ? Access::create(dummyRange, seedNameOf(lhs_name), lhs_indices)
+          : Const::create(
+                dummyRange,
+                Number::create(0),
+                Compound::create(TK_FLOAT, dummyRange, {}));
+
+      grad_comps_.push_back(Comprehension::create(
+          dummyRange,
+          lhs_name,
+          lhs_indices,
+          Compound::create('=', dummyRange, {}),
+          rhs_expr,
+          ListView<TreeRef>::create(dummyRange, TreeList{}),
+          Compound::create(TK_OPTION, dummyRange, {}),
+          ListView<TreeRef>::create(dummyRange, TreeList{})));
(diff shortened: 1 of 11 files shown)


prigoyal:
yay :)

apaszke:
I can of course answer some more specific questions, but most of this patch is basically a tiny bit of [reverse mode AD code](https://rufflewind.com/2016-12-30/reverse-mode-automatic-differentiation), and a ton of defensive programming to perform bookkeeping and shield off from unsupported features.

prigoyal:
oh very cool, thanks for the reference @apaszke :) I'll look at that tutorial and then look over this PR.

test/test_autodiff.cc
```diff
@@ -0,0 +1,46 @@
...
+#include <glog/logging.h>
+#include <gtest/gtest.h>
+
+#include "tc/core/autodiff.h"
+
+#include "lang_utils.h"
+
+using namespace tc;
+
+void generalTest() {
```
prigoyal: I am curious about two things atm so far from looking at the test case:

1. how do we handle temporaries like the example https://facebookresearch.github.io/TensorComprehensions/framework/pytorch_integration/autograd_with_tc.html#reordering-grad-outputs ? would it be possible to demonstrate such use case with test example?

2. How is the inputs/outputs order handled for backwards?

apaszke: #### Regarding 1 and 2:

The test case has temporaries, so you can actually see this in the example. In general the grad TC currently takes in:
- inputs
- outputs (including intermediates)
- grads w.r.t. outputs (including intermediates)
and returns grads w.r.t. inputs. Inputs and outputs are ordered in the same way as they were specified in the original TC.

#### Regarding 3.

No, that's not currently possible, but is a very straightforward extension (only need a few lines of code).

abadams:
The bulk of the work in the Halide autodiff was handling non-trivial indexing in ways that preserve parallelism. E.g. consider:

A(i) = 2 * B(some nasty expression in i)

You don't want to compute the derivatives as:

B'(some nasty expression in i) = A'(i) / 2

because you can't necessarily parallelize that over i, so it's going to be slow. This sort of thing comes up as soon as you have a stencil or broadcast (one input influences multiple outputs, when reversed, introduces a race condition). 

Dealing with bounds and shapes of the computed derivatives was also interesting in the Halide work. I think that's simpler in TC, because tensors have finite size.

I think this PR is the right sort of approach, and we should do the rewrite of B(nasty) = A'(i) / 2 into B(j) = A'(nasty) / 2 as a later pass inside lowering, using the Halide solver or polyhedral tools.

apaszke:
@abadams Thanks for the review! Great, it seems that we're on the same page! 😄 

I agree that shifting the formulas to lhs isn't perfect, but it could easily be done as a post processing step. I know how to implement this, it's just that I'd need a linear equation solver (which is not available at this level from what I understand). For now I decided to bail and put the indexing expressions there, but it's likely to cause errors downstream anyway (I don't think it's supported in TC/Halide). It would be useful to support this in general too.

Bounds are another problem, and you can see that I had to implement a few checks to make sure that they can be still auto-inferred (`usedIndexVars` + `requireAllIndexVarsOf`). This is still not perfect because there are a few simple things like `where` ranges, that also cause an error for now, but are fairly easy to add later. I didn't want to complicate the initial patch so we can quickly get this in, and then work on improvements.

abadams:
Equation solvers exist once you get down to Halide IR or polyhedral IR. It would be silly to implement yet another one at the TC-front-end-IR level. I'd just allow complex index expressions on the LHS for now, which I think is what you're doing? They may barf inside the tc2halide layer right now, but support for those is an outstanding near-term TODO - we need them for things like histogram computation.

apaszke:
Yep, that's exactly what I've been thinking. Good we're on the same page!

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+      case 64:
+        return TK_DOUBLE;
+      case 32:
+        return TK_FLOAT;
+    }
+  }
+  throw std::runtime_error("Unknown Halide type");
+}
+
+void findAccessedTensors(
```
nicolasvasilache: We have been implementing this type of functions as:
```
std::unordered_set<std::string> collectAccessedTensors(const TreeRef& tree) {...}
```
This way it can compose. 
Generally we try not to pass out parameters as mutable in but just return the result.

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    for (const TreeRef& subtree : tree->trees()) {
+      findAccessedTensors(read_only, subtree);
+    }
+  }
+}
+
+void assertNoWriteAfterRead(Def def) {
+  std::unordered_set<std::string> read_only;
+  // Inputs are always read-only
+  for (Param input : def.params())
```
nicolasvasilache: braces even around 1-liner for/if please

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    read_only.insert(input.ident().name());
+  for (Comprehension comp : def.statements()) {
+    findAccessedTensors(read_only, comp.rhs());
+    auto lhs_name = comp.ident().name();
+    if (read_only.count(lhs_name) > 0)
+      throw std::runtime_error(
+          "AD not supported in TCs that write to a value after reading it");
```
nicolasvasilache: So this would not work for FCRelu?
```
def fcrelu(float(N, M) I, float(M, P) W, float(P) Bias) -> (O) {
  O(n, p) +=! W(r_m, p) * I(n, r_m)
  O(n, p)  = O(n, p) + B(p)
}
```
Is this a fundamental limitation or a future TODO?

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+        return TK_DOUBLE;
+      case 32:
+        return TK_FLOAT;
+    }
+  }
+  throw std::runtime_error("Unknown Halide type");
+}
+
+void findAccessedTensors(
+    std::unordered_set<std::string>& read_only,
```
nicolasvasilache: Also, why convert to string at this point, returning TreeRef seems fine to me?

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    }
+  }
+}
+
+// XXX: this is a bit of a fragile hack, and can easily break when the AST will
+// get more idents in different nodes, but it's quite simple and the worst thing
+// that can happen is that we will be too conservative and throw, so it's ok.
+std::unordered_set<std::string> usedIndexVars(Comprehension comp) {
+  std::unordered_set<std::string> index_vars;
+  for (Ident idx : comp.indices())
```
nicolasvasilache: braces plz

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+      throw std::runtime_error(
+          "Not all index variables are used in gradient comprehension. "
+          "AD will require range inference to support this case.");
+  }
+
+  bool hasZeroGrad(const std::string& name) {
+    return has_zero_grad_.count(name) > 0;
+  }
+  void markZeroGrad(const std::string& name) {
+    has_zero_grad_.count(name);
```
nicolasvasilache: does this function do anything?

test/lang_utils.h
```diff
@@ -0,0 +1,110 @@
+/**
```
nicolasvasilache: This extraction is useful by itself and independent, please split in a separate commit

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    output_types.push_back(
+        TensorType::create(dummyRange, scalar_type, dim_sizes));
+  }
+
+  return output_types;
+}
+
+std::string differentiate(const std::string& source) {
+  // Parse and check the source
+  auto def = Def(Sema().checkFunction(Parser(source).parseFunction()));
```
nicolasvasilache: What happens if the TC has more than 1 def?

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+#include "tc/lang/tree_views.h"
+
+#include <unordered_set>
+#include <vector>
+
+namespace tc {
+
+using namespace lang;
+
+static const lang::SourceRange dummyRange{std::make_shared<std::string>(""),
```
nicolasvasilache: no freestanding globals please, make it a static definition inside a function in an anon namespace and return that instead

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+        grad_info.seedNameOf(returns[i].ident()),
+        output_types.at(i)));
+  }
+
+  // --------------------------------------------------------------------------
+  // Differentiate the body
+  auto body = def.statements();
+  auto it = body.end();
+  if (it == body.begin())
+    throw std::runtime_error("empty body");
```
nicolasvasilache: braces around if
append TC to the message?

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+  do {
```
nicolasvasilache: can we use a for / rbegin / rend loop?

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    Comprehension comp = *(--it);
+
+    int assign_kind = comp.assignment()->kind();
+    if (assign_kind != '=' && assign_kind != TK_PLUS_EQ_B &&
+        assign_kind != TK_PLUS_EQ)
```
nicolasvasilache: braces plz

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+      throw ErrorReport(comp)
+          << "Only =, += and +=! assignments are supported in AD";
+    if (comp.whereClauses().size() > 0 || comp.equivalent().present())
```
nicolasvasilache: braces plz

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+  for (Comprehension comp : def.statements()) {
+    findAccessedTensors(read_only, comp.rhs());
+    auto lhs_name = comp.ident().name();
+    if (read_only.count(lhs_name) > 0)
+      throw std::runtime_error(
+          "AD not supported in TCs that write to a value after reading it");
+  }
+}
+
+void findIndexVars(
```
nicolasvasilache: Same comment as above re return values vs mutable reference as input.

src/lang/test_expected/general_ad.expected
```diff
@@ -0,0 +1,8 @@
+def grad_mm(double(M, K) A, double(K, N) B, double(M, N) O, double(M, N) W, double(M, N) seed_d_O, double(M, N) seed_d_W) -> (d_A, d_B) {
+  d_O(i, j) = seed_d_d_O(i, j)
+  d_O(i, j) += (seed_d_W(i, j) * 2)
+  d_A(i, k) = 0
```
nicolasvasilache: In order to be parsed by TC this will need either a where clause or to be folded into `+=!`.

More generally I'd make the test run parsing + sema on the produced TC and print / compare against the canonical TC. Dogfooding would have caught the where case here.

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    default:
+      throw ErrorReport(expr) << "Unsupported expression kind in AD: "
+                              << kindToString(expr->kind());
+  }
+}
+
+// XXX: Sema isn't nilpotent, so we have to reparse the source
+std::vector<TreeRef> inferOutputTypes(const std::string& source) {
+  auto halide_def =
+      tc2halide::translate(isl::with_exceptions::globalIslCtx(), source, true);
```
nicolasvasilache: So this now makes lang depend on ISL which is unnecessary.
We should extract what we need from halide for range inference.

src/core/autodiff.cc
```diff
@@ -0,0 +1,387 @@
...
+    grad_comps_.push_back(Comprehension::create(
+        dummyRange,
+        lhs_name,
+        lhs_indices,
+        Compound::create(TK_PLUS_EQ, dummyRange, {}),
+        rhs_expr,
+        ListView<TreeRef>::create(dummyRange, TreeList{}),
+        Compound::create(TK_OPTION, dummyRange, {}),
+        ListView<TreeRef>::create(dummyRange, TreeList{})));
+    if (usedIndexVars(Comprehension(grad_comps_.back())) !=
```
nicolasvasilache: braces plz

nicolasvasilache (commented):
Made a first quick pass, generally this look great, thanks for your contribution.
I made a few comments, I'll finish reviewing early next week.

I have one question regarding gathers, I did not seem to see a guard against those in AD (i.e. `A(B(i))`), can you point where such cases are caught and what we throw on them?

PR #229, opened 2018-03-27
Remove the flags check for logging in two places

prigoyal:
as requested by @nicolasvasilache , I am making this change now. There might be a pybind change required likely so I'll do that after looking at CI runs. :)

### DIFF
--- a/src/core/cuda/cuda_tc_executor.cc
+++ b/src/core/cuda/cuda_tc_executor.cc
@@ -174,9 +174,7 @@ void CudaTcExecutor::compileWithTcMapper() {
   // that.
   std::tie(cudaSource, grid, block) =
       mappedScop->codegen(kernelSpecializedName);
-  if (FLAGS_dump_cuda) {
-    std::cout << "generatedCuda: " << cudaSource << std::endl;
-  }
+  LOG_IF(INFO, FLAGS_dump_cuda) << "generatedCuda: " << cudaSource;
 }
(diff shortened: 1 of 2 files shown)


src/core/cuda/cuda_tc_executor.cc
```diff
@@ -174,9 +174,7 @@ void CudaTcExecutor::compileWithTcMapper() {
   // that.
   std::tie(cudaSource, grid, block) =
       mappedScop->codegen(kernelSpecializedName);
-  if (FLAGS_dump_cuda) {
-    std::cout << "generatedCuda: " << cudaSource << std::endl;
-  }
+  LOG_IF(INFO FLAGS_dump_cuda) << "generatedCuda: " << cudaSource;
```
nicolasvasilache: missing comma?
This shouldn't compile as is

prigoyal:
I marked this PR as wip to wait and see what things look like in docker environment. There might be a likely change required for the pybinds :)

skimo-openhub:
On Tue, Mar 27, 2018 at 01:45:41PM -0700, Priya Goyal wrote:
> as requested by @nicolasvasilache , I am making this change now. There might be a pybind change required likely so I'll do that after looking at CI runs. :)

What is the motivation for this change?
Apparently, you and @nicolasvasilache know, but it's not recorded
in the commit message.

skimo

PR #231, opened 2018-03-28
halide2isl::makeScheduleTree: use isl::schedule for constructing tree

skimo-openhub:
The original code was using a domain-less ScheduleTree and
a separate domain, resulting in an unhandled special case
where no ScheduleTree gets built.
For consistency, it is better to always return a complete
schedule, with domain.  This requires some extra helper functions,
but those are already available in the isl::schedule abstraction,
so use that instead, further removing additional low-level
tree manipulations.
Since isl is used to construct the tree, it automatically
satisfies the isl invariants, so the corresponding check
can also be removed.


skimo-openhub:
@caffe2bot retest this please

PR #233, opened 2018-03-28
Allow using a user-defined key in the compilation cache

### LINKED ISSUE #230: support for autotune cache reuse for different machine
@ttheodor 
i autotune the function in server with 4gpu and get cache file.
and i try to reuse generated cache file in different server with 4gpu.
but it throws error
2 server has diffenet gpu.
@nicolasvasilache said it is key related problem.
could you give me some instrunction to reuse cache files?

thetheodor:
resolves #230


nicolasvasilache (approved):

PR #235, opened 2018-03-29
Add a fused 2 layer wavenet example

nicolasvasilache:


examples/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ set(GTEST_LIBS gtest gtest_main)
 # Examples
 ################################################################################
 set(EXAMPLES_FILES
+  dilate
```
RPrenger: This kernel is essentially 2 layers of a WaveNet.  https://arxiv.org/pdf/1609.03499.pdf

While the WaveNet uses dilated convolutions, the dilation part isn't in this kernel specifically, so a name like `wavenet_layer_2` or something similiar might be more appropriate

nicolasvasilache:
@RPrenger FYI I played locally with 3 blocks instead of 2.
I did see in the first generation a 120us pop up for the whole thing but compilation times are much slower and eventually I stopped the run.

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+#include <ATen/ATen.h>
+
+#include "tc/aten/aten_compiler.h"
+#include "tc/autotuner/genetic_autotuner_aten.h"
+#include "tc/core/cuda/cuda_mapping_options.h"
+#include "tc/core/flags.h"
+
+#include "../test/test_harness_aten_cuda.h"
+
+DEFINE_string(tuner_proto, "", "Filename to load and store proto cache ");
```
ftynse: maybe add another prefix to make clear this flag is specific to this test, rather than a generic tuner flag?

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+
+TEST(WaveNet2Layers, SimpleAutotune) {
+  // 1. Define and setup the TC compilation unit with CUDA memory
+  // management backed by ATen tensors.
+  std::string tc = R"TC(
+def wavenet2layers(float(OUT, IN, KERN) Weight0,
```
ftynse: I'd start parameter list from the new line in this case

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+      Skip0(batch, skip)  =   SkipBias0(skip)
+            where batch in 0:BATCH
+      Skip0(batch, skip) += SkipWeight0(skip, r_in) * NonLin0(batch, r_in)
+            where r_in in 0:IN # necessary because r_in gets into unresolved min/max
+       Res0(batch, out)   =   ResBias0(  out)
+            where batch in 0:BATCH
+       Res0(batch, out)  += ResWeight0(  out, r_in) * NonLin0(batch, r_in)
+            where r_in in 0:IN # necessary because r_in gets into unresolved min/max
+       Res0(batch, out)   =       Res0(batch,  out) + NonLin0(batch, out)
+            where out in 0:IN # necessary because r_in gets into unresolved min/max
```
ftynse: copy-pasta in the comment, there is no `r_in` in this expression

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+    float(OUT) Bias1,
+    float(BATCH, IN, KERN) Data1,
+    float(IN, IN) ResWeight1,
+    float(IN) ResBias1,
+    float(SKIP, IN) SkipWeight1,
+    float(SKIP) SkipBias1)
+    -> (Res0, Dilate0, NonLin0, Skip0, Res1, Dilate1, NonLin1, Skip1)
+{
+    Dilate0(batch, out)   =   Bias0(out)
+            where batch in 0:BATCH
```
ftynse: aligning parenthesis on the RHS does not look good if the `where` clause is moved to the next line

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+    Dilate0(batch, out)  += Weight0(out, r_in, r_kern) * Data0(batch, r_in, r_kern)
+    NonLin0(batch, out)   = 1 / (1 + exp(-1*(Dilate0(batch, out))))
+    NonLin0(batch, out)  *= tanh(Dilate0(batch, out + 64))
+      Skip0(batch, skip)  =   SkipBias0(skip)
```
ftynse: The parentheses on the RHS are not aligned with the previous statement, but are aligned with the following one, so I'd add an empty line before this expression.

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+            where batch in 0:BATCH
+      Skip0(batch, skip) += SkipWeight0(skip, r_in) * NonLin0(batch, r_in)
+            where r_in in 0:IN # necessary because r_in gets into unresolved min/max
+       Res0(batch, out)   =   ResBias0(  out)
```
ftynse: The parentheses on the RHS are not aligned with the previous statement, but are aligned with the following one, so I'd add an empty line before this expression.

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+            where r_in in 0:IN # necessary because r_in gets into unresolved min/max
+       Res0(batch, out)   =       Res0(batch,  out) + NonLin0(batch, out)
+            where out in 0:IN # necessary because r_in gets into unresolved min/max
+
+    Dilate1(batch, out)   =   Bias1(out)
+            where batch in 0:BATCH
+    Dilate1(batch, out)  += Weight1(out, r_in, r_kern) * Data1(batch, r_in, r_kern)
+    NonLin1(batch, out)   = 1 / (1 + exp(-1*(Dilate1(batch, out))))
+    NonLin1(batch, out)  *= tanh(Dilate1(batch, out + 64))
+      Skip1(batch, skip)  =   SkipBias1(skip)
```
ftynse: The parentheses on the RHS are not aligned with the previous statement, but are aligned with the following one, so I'd add an empty line before this expression.

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+            where batch in 0:BATCH
+      Skip1(batch, skip) += SkipWeight1(skip, r_in) * NonLin1(batch, r_in)
+            where r_in in 0:IN # necessary because r_in gets into unresolved min/max
+       Res1(batch, out)   =   ResBias1(  out)
+            where batch in 0:BATCH
+       Res1(batch, out)  += ResWeight1(  out, r_in) * NonLin1(batch, r_in)
+            where r_in in 0:IN # necessary because r_in gets into unresolved min/max
```
ftynse: copy-pasta about `r_in` in the comment

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+  //   res_weight1 = 2*(torch.cuda.FloatTensor(64,64).uniform_() - 1.5)
+  //   res_bias1 = 2*(torch.cuda.FloatTensor(64).uniform_() - 1.5)
+  //   skip_weight1 = 2*(torch.cuda.FloatTensor(256,64).uniform_() - 1.5)
+  //   skip_bias1 = 2*(torch.cuda.FloatTensor(256).uniform_() - 1.5)
+
+  // 6. Run unchecked 1000 times, to put GPU in high usage mode, use with:
+  //   nvprof --profile-from-start off executable --use_nvprof=1
+  {
+    tc::CudaProfiler cp;
+    for (int i = 0; i < 1000; ++i) {
```
ftynse: we have `FLAG_benchmark_iterations` for this

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+       res_bias1,
+       skip_weight1,
+       skip_bias1},
+      naiveOptions);
+
+  // 4. Compile and run the TC with the best option.
+  // Outputs get allocated; could also be pre-allocated and passed.
+  auto handle = atCompl.compile(
+      "wavenet2layers",
+      {weight0,
```
ftynse: let's create a `vector` of tensors once and reuse it instead of copy-pasting code every time

examples/wavenet.cc
```diff
@@ -0,0 +1,229 @@
...
+      true);
+  std::cout
+      << "wavenet2layers size weight0: " << weight0.sizes() << " ran in: "
+      << std::chrono::duration_cast<std::chrono::microseconds>(duration).count()
+      << "us\n";
+
+  // 5. TODO: the following represent reasonable initialization
+  // operations. Port them from PyTorch if we want to look at precision in the
+  // future.
+  //   weight0 = 5*(torch.cuda.FloatTensor(128,64,2).uniform_() - 0.5)
```
ftynse: These seem pretty straightforward to translate, was there any particular problem?

examples/blockdiagperm.cc
```diff
@@ -0,0 +1,129 @@
...
+#include <ATen/ATen.h>
+
+#include "tc/aten/aten_compiler.h"
+#include "tc/autotuner/genetic_autotuner_aten.h"
+#include "tc/core/cuda/cuda.h"
+#include "tc/core/cuda/cuda_tc_executor.h"
+#include "tc/core/flags.h"
+#include "tc/core/mapping_options.h"
+
+DEFINE_string(tuner_proto, "", "Filename to load and store proto cache ");
```
ftynse: this appears in every example, maybe you want an actual tuner flag instead?

nicolasvasilache: This was removed in a PR by @ttheodor because it was dead code in the mapper and gave a false impression that it was doing something in the mapper. In practice each binary may or may not want to save. For the simple examples we build small reinforcement loops by using the proto and running multiple times. One could argue that that this belongs in a harness, sure but then other things too and we end up in a position where people complain about complexity (e.g. benchmarks).  So this is our "examples" tradeoff, we may want to avoid the reinforcement loop in the future if we are picky about the flag.

PR #236, opened 2018-03-30
codegen_cuda.cc: emitCopyStmt: skip conversion to isl::map

skimo-openhub:
range_factor_domain and range_factor_range are now also
available on isl::pw_multi_aff.


ftynse:
`+0 −19,752 		include/isl/cpp.h `
this does not look right

skimo-openhub:
On Fri, Mar 30, 2018 at 11:15:08AM +0000, ftynse wrote:
> `+0 −19,752 		include/isl/cpp.h `
> this does not look right

That's why it's marked WIP.

skimo

PR #238, opened 2018-03-30
src/lang/lexer.cc: compare const char* strings using strcmp

skimo-openhub:
There is no guarantee that the compiler maps identical strings
to the same memory location.


skimo-openhub:
On Fri, Mar 30, 2018 at 09:52:51PM +0000, ftynse wrote:
> 
> I also noticed the warning.  Are we sure that `str` is a C string?

Yes,  It's always a string literal.

skimo

PR #244, opened 2018-04-02
unit test for mul reduction, use pytest + gen=2/pop=1 to improve test time

### LINKED ISSUE #245: Multiply reduction not initializing with identity of reduction
A user reported this problem here https://tensorcomprehensions.slack.com/archives/C98GR4E10/p1522443828000045

The multiply reduction *=! doesn't seem to be working i.e. it doesn't init output with the identity of reduction * which should be 1. Here the output is all still 0.

prigoyal:
This PR makes following changes:

1. A user reported some issue related to *=! here https://tensorcomprehensions.slack.com/archives/C98GR4E10/p1522443828000045 add a unit test for *=! for reasonable tensor data
2. use pytest to spin one python process for running test. If separate process is run for separate test, cuda driver initialization cost is incurred when using pytorch. Using pytest allows running all tests in one python process. This saves about 2 minutes of time on my terminal
3. Use gen=1 and population size = 2 setting as the minimal. I am not quite sure why population size=1 doesn't work.  I am not myself comfortable with having gen=1 as in past, when running python test, I have often seen errors in gen=2 for some layers but the errors don't happen in gen=1. Hence open to suggestions if anyone has any. 

overall, I have 7m of python test on my terminal.

closes #245

[the reason this PR closes #245 is by providing a test case that the *=! works as expected]

### DIFF
--- a/tensor_comprehensions/tc_unit.py
+++ b/tensor_comprehensions/tc_unit.py
@@ -31,13 +31,13 @@ logger = logging.getLogger(__name__)
 
 # these are quick options for finishing autotuning
 autotuner_settings = {
-    "threads": 32, "generations": 2, "pop_size": 5,
+    "threads": 32, "generations": 1, "pop_size": 2,
 }
 
 # TC prunes autotuning for kernels which require < 256 threads. So to tune small
 # size kernels, we set the min kernel threads to 1
 small_sizes_autotuner_settings = {
-    "threads": 32, "generations": 2, "pop_size": 5, "tuner_min_launch_total_threads": 1,
+    "threads": 32, "generations": 1, "pop_size": 2, "tuner_min_launch_total_threads": 1,
 }
 
 ###############################################################################
--- a/test_python/layers/test_multiply_reduction.py
+++ b/test_python/layers/test_multiply_reduction.py
@@ -1,0 +1,39 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##############################################################################
+
+import tensor_comprehensions as tc
+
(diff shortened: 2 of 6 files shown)


prigoyal:
Hi @zdevito , it seems like buggy to me but I am not sure if we support `*=!` reduction yet or not. Do you have some ideas? :)

tensor_comprehensions/tc_unit.py
```diff
@@ -31,13 +31,13 @@
 
 # these are quick options for finishing autotuning
 autotuner_settings = {
-    "threads": 32, "generations": 2, "pop_size": 5,
+    "threads": 32, "generations": 2, "pop_size": 2,
```
ftynse: irrelevant change

tensor_comprehensions/tc_unit.py
```diff
@@ -31,13 +31,13 @@
...
 }
 
 # TC prunes autotuning for kernels which require < 256 threads. So to tune small
 # size kernels, we set the min kernel threads to 1
 small_sizes_autotuner_settings = {
-    "threads": 32, "generations": 2, "pop_size": 5, "tuner_min_launch_total_threads": 1,
+    "threads": 32, "generations": 2, "pop_size": 2, "tuner_min_launch_total_threads": 1,
```
ftynse: irrelevant change

ftynse:
As usual, I would prefer functionality tests to be in C++.  There's `test_tc_mapper_bugs.cc` with a bunch of easy-to-replicate tests.

test_python/layers/test_multiply_reduction.py
```diff
@@ -0,0 +1,39 @@
...
+       def conv_mult(float(N,C,H,W) I, float(M,C,KH,KW) W1) -> (O) {
+           O(n, m, h, w) *=! I(n, c, h + kh, w + kw) * W1(m, c, kh, kw)
+       }
+       """
+       N, C, H, W, O, kH, kW = 64, 10, 24, 24, 10, 7, 7
+       conv_mult = tc.define(LANG, name="conv_mult")
+       I, W1 = torch.ones(N, C, H, W).cuda(), torch.ones(O, C, kH, kW).cuda()
+       # Note: There is no bias here
+       out = conv_mult(I, W1)
+       assert out.data.min() > 0
```
zdevito: We should probably have at least 1 test that checks the result of the multiply is correct.  Using a very small matrix size will work around any precision issues.

prigoyal: yaap, agreed. good point that we can just use multiply operations, not just for this particular test but for demonstrating other features.

prigoyal:
thanks @zdevito :)

PR #248, opened 2018-04-02
Use const pointers for input tensors in generated cuda

nicolasvasilache:


nicolasvasilache:
@ftynse  not sure when you did but it's always been the semantics of TC that inputs are immutable and tmp + outputs are mutable. However we never actually enforced it, I added a commit to tie that loose end. Note that nothing prevents abuse at the boundary: you can have tmp and output views that point into input tensor space.

PR #250, opened 2018-04-03
test_isl_functionality.cc: drop dead code

skimo-openhub:
The removed function was used in a test that was removed in prehistory.


skimo-openhub:
Superseded by #252.

PR #262, opened 2018-04-05
Allow computed expressions on the left-hand-side

abadams:


ftynse:
This also needs support in polyhedral dependence analysis, putting a block for now

ftynse:
I pushed the polyhedral support for scattering to this branch directly.  Requesting @nicolasvasilache to review my code.

abadams:
Given that the tests already passed without the polyhedral support, should there be new tests that exercise it?

include/tc/core/libraries.h
```diff
@@ -81,16 +87,16 @@ float fmodf ( float  x, float  y );
 //float frexpf ( float  x, int* nptr );
 float hypotf ( float  x, float  y );
 //int ilogbf ( float  x );
-//__RETURN_TYPE 	isfinite ( float  a );
-//__RETURN_TYPE 	isinf ( float  a );
-//__RETURN_TYPE 	isnan ( float  a );
+//__RETURN_TYPE        isfinite ( float  a );
```
nicolasvasilache: tab vs spaces war ? :)

include/tc/core/libraries.h
```diff
@@ -32,8 +32,14 @@ namespace c {
 
 constexpr auto types = R"C(
 // Halide type handling
-typedef int int32;
-typedef long int64;
+typedef signed char int8;
```
nicolasvasilache: Thanks for tying that loose end, deserves a separate commit?

include/tc/core/libraries.h
ftynse: probably was never checked by clang-format before that

abadams:
A(i) would have to be injective and surjective for that scatter to define a total function. We need to decide on the semantics for repeated values on A(i) and values that are never hit. It sounds like you're saying is we should permit race conditions on the writes, and we leave values that are never written as whatever is already in the buffer.

If that's the case, then I think tc2halide is correct, and we just need to get sema.h to relax and not throw that error.

ftynse:
Reviving this.  I made Sema look inside the subscript expressions on the LHS so it no longer complains at that level.  In particular, it accepts things like `O(i + 2) = A(i) + B(i)` which it previously considered to be reductions, and thus failing.

However, on the stupid scattering example `O(A(i)) = B(i)`, range inference now complains about LHS index being unbounded.  It indeed is, and as Andrew mentioned, i->A(i) must be a bijective relation for this to be correct.  We can insert a very expensive runtime check, but I'm not sure it's a good idea so I'm inclined to just disallow that for now.

Thoughts, @abadams @nicolasvasilache  ?

abadams:
My inclination would be to make O(A(i)) = B(i) work if A(i) has known bounds (e.g. it's a uint8).

skimo-openhub:
On Wed, Apr 25, 2018 at 03:25:21PM +0000, ftynse wrote:
> Reviving this.  I made Sema look inside the subscript expressions on the LHS so it no longer complains at that level.  In particular, it accepts things like `O(i + 2) = A(i) + B(i)` which it previously considered to be reductions, and thus failing.
> 
> However, on the stupid scattering example `O(A(i)) = B(i)`, range inference now complains about LHS index being unbounded.  It indeed is, and as Andrew mentioned, i->A(i) must be a bijective relation for this to be correct.  We can insert a very expensive runtime check, but I'm not sure it's a good idea so I'm inclined to just disallow that for now.

Would it be an option to simply require users to write
`O(i) = B(invA(i))` instead?

skimo

abadams:
If they knew the inverse of A, they'd probably already be done. The code that inverts A is something like:

invA(A(i)) = i

ftynse:
> My inclination would be to make O(A(i)) = B(i) work if A(i) has known bounds (e.g. it's a uint8).

And infer 0:256 as bounds for `O`?

What about `O(i + A(j+k))  = C(i,j,k)` ?  And what would the inferred bounds be in this case?

PR #263, opened 2018-04-06
disallow mapping nested bands to threads

ftynse:
mapInnermostBandsToThreads maps coincident band members to threads
while traversing the tree recursively in DFS postorder.  It is possible
that members of two nested bands get mapped to thread dimensions.
Moreover, this is necessary for parallel reduction support as the
reduction dimension is split out into a separate band.  However, such
mapping is not valid in general because members of different bands are
not necessarily permutable with each other. Since thread dimensions are
not executed in any specific order, i.e. there is no guarantee that, for
a given value of threadIdx.y, all threadIdx.x threads will be executed
before (or after) those with another value of threadIdx.y, permutability
is a necessary condition for mapping to multiple thread dimensions.
Only allow mapping a band to threads if its children are not already
mapped to threads, except for a single child node with reduction mapped
only to threadIdx.x under condition that this node was split out of a
larger permutable band.

### DIFF
--- a/src/core/polyhedral/cuda/mapped_scop.cc
+++ b/src/core/polyhedral/cuda/mapped_scop.cc
@@ -192,8 +192,9 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
     found |= detectReductions(c);
   }
   auto band = tree->elemAs<detail::ScheduleTreeElemBand>();
-  // Nested reductions are not currently supported.
-  if (!band || found) {
+  // Nested reductions or reductions in non-permutable bands are not currently
+  // supported.
+  if (!band || !band->permutable_ || found) {
     return found;
   }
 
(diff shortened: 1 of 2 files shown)


src/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -192,8 +192,9 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
     found |= detectReductions(c);
   }
   auto band = tree->elemAs<detail::ScheduleTreeElemBand>();
-  // Nested reductions are not currently supported.
-  if (!band || found) {
+  // Nested reductions or reductions in non-permutable bands are not currently
+  // supported.
+  if (!band || !band->permutable_ || found) {
```
skimo-openhub: Is this related to not mapping nested bands?

src/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -442,15 +432,19 @@ bool hasOuterSequentialMember(
...
   }
   auto children = st->children();
   auto nChildren = children.size();
   std::vector<size_t> nInner(nChildren);
+  bool allowNestedMapping = false;
   for (size_t i = 0; i < nChildren; ++i) {
-    nInner[i] = mapInnermostBandsToThreads(children[i]);
+    auto inner = mapInnermostBandsToThreads(children[i]);
+    nInner[i] = inner.first;
+    allowNestedMapping = (i == 0 ? inner.second : false);
```
skimo-openhub: This looks like complicated way of writing `i == 0 && inner.second`.

ftynse:
> In hindsight, it was probably a mistake to break the band up in the first place. Have you tried resolving this issue by not splitting up the band?

Now I did, turned out to be slightly more involved then I thought initially.  We still need to split out the reduction update band to insert synchronizations, but it can be done after mapping it to threads.

skimo-openhub:
On Mon, Apr 09, 2018 at 03:31:53AM -0700, ftynse wrote:
> > In hindsight, it was probably a mistake to break the band up in the first place. Have you tried resolving this issue by not splitting up the band?
> 
> Now I did, turned out to be slightly more involved then I thought initially.  We still need to split out the reduction update band to insert synchronizations, but it can be done after mapping it to threads.

Did you push out your changes?

skimo

ftynse:
Still need to clean them up

ftynse:
see #283

skimo-openhub:
Superseded by #283.

PR #266, opened 2018-04-06
further simplify reduction handling

skimo-openhub:
In particular, remove a use of isl::MUPA, which is set to be removed.


skimo-openhub:
All tests passed before the rebase.

PR #267, opened 2018-04-06
use standard isl support for lists intead of homegrown alternatives

skimo-openhub:


skimo-openhub:
On Sat, Apr 07, 2018 at 11:15:56AM +0000, Nicolas Vasilache wrote:
> ISL C++ seems due for an "use standard STL support for common data types instead of homegrown alternatives", no?
> 
> Do you think there is a way this could happen (in the longer term) without reimplementing ISL?

Let's get a basic C++ interface working first.

skimo

prigoyal:
@caffe2bot retest this please

skimo-openhub:
@caffe2bot retest this please

PR #273, opened 2018-04-07
Drop default cache filename

ttheodor:
To avoid surprises, protobuf cache filename is by default empty
and the user has to explicitly specify it.


nicolasvasilache:
cc @prigoyal  notice the dead code removal also ended up removing one of the pybind functions.

prigoyal:
thanks @ttheodor. This PR has broken the python test and all the PRs will fail due to this reason although the test are not push blocking. May I kindly request that we try our best as a team to merge PRs when unit test pass :)

PR #274, opened 2018-04-07
Update for LLVM JIT API Churn

wsmoses:
To quote LLVM docs "The text will be updated once the API churn dies down."

Master JIT LLVM api is undergoing major churn atm, so this adds some initial code to unblock FB folks since it seems to be time-sensitive.

Once LLVM API churn dies down, this code will be made nicer.

### DIFF
--- a/tc/core/polyhedral/llvm_jit.cc
+++ b/tc/core/polyhedral/llvm_jit.cc
@@ -25,6 +25,7 @@
 #include "llvm/IR/Verifier.h"
 #include "llvm/Passes/PassBuilder.h"
 #include "llvm/Support/DynamicLibrary.h"
+#include "llvm/Transforms/Utils/Cloning.h"
 
 #include "tc/core/flags.h"
 #include "tc/core/polyhedral/codegen_llvm.h"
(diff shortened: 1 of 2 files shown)


wsmoses:
@caffe2bot retest

tc/core/polyhedral/llvm_jit.cc
```diff
@@ -111,7 +100,64 @@ Jit::ModuleHandle Jit::addModule(std::shared_ptr<Module> M) {
...
+          ES,
+          [this](llvm::orc::VModuleKey) {
+            return llvm::orc::RTDyldObjectLinkingLayer::Resources{
+                std::make_shared<SectionMemoryManager>(), Resolver};
+          }),
+      compileLayer_(objectLayer_, orc::SimpleCompiler(*TM_)) {
+  std::string err;
+
+  auto path = find_library_path("libcilkrts.so");
+  sys::DynamicLibrary::LoadLibraryPermanently(path.c_str(), &err);
```
salexspb: if we remove previous line and pass nullptr here, tests pass now.

Issue #194, opened 2018-03-23
Modernization of TC repo


prigoyal:
We have been having some discussions to modernize the TC codebase. Here are the proposed changes:

- [x] ~Create a top level 'tc' folder, create tc/proto and move .proto files there. They should be included as tc/proto/mapping_options.pb.h~ done in #279 
- [x] ~Move subfolders of include/tc into tc/ and same for subfolders of src/~ done in #279 
    - Use best judgement for merging some of those folders if it makes sense
    - This should be synced with other big PRs that might be in progress
- [x] ~supply -WError in dev builds and in CI. Make sure that errors Mingzhe Li fixed for various virtual destructors trigger errors in OSS build as well~ done in #284  

cc @salexspb

prigoyal:
Hi TC team: @skimo-openhub , @ttheodor , @ftynse , @nicolasvasilache , @abadams , @zdevito 

as we have been discussing this over VC and have concluded to most things, I created this issue to track the progress as I work on this. I would love to hear comments from you all on this.

nicolasvasilache:
nice summary, thanks!
looks good to me.
Let's just hold off until pr/decouple-cuda is sent for review and lands otherwise merge conflicts will be massive.

prigoyal:
thank you @nicolasvasilache for the review. Do we have an ETA on pr/decouple-cuda for when it is expected to land? :) cc @salexspb

nicolasvasilache:
I'm cutting a large commit into smaller ones as we speak, it will prob require a few rounds of review but hopefully will land soon.

ftynse:
I also prefer #149 and #161 to land.  Otherwise I will feel like abandoning those PRs.

nicolasvasilache:
@skimo-openhub absolutely, the circular dependence should be a non-problem and an indication of code-smell. We should ensure we use TensorComprehensions.h internally and that will probably require refactorings (e.g. we probably should not expose the functionality from ```core/cuda/cuda.h``` at any time in ```TensorComprehensions.h ``` or we'll have to start including ```#ifdef``` spaghetti; ideally we'd avoid that unless absolutely necessary).

salexspb:
I am bit confused about TensorComprehensions.h causing circular deps if it just does includes of private code. 

On a separate note - what is the evidence that it is needed? I tried to find example of modern codebases with large external community. PyTorch is not good example as it was mainly focused on Python.

[...]

PR #279, opened 2018-04-09
Modernization of TC codebase

prigoyal:
This PR makes following changes:

1. Create a top level 'tc' folder, create tc/proto and move .proto files there. They should be included as tc/proto/mapping_options.pb.h
2. Move subfolders of include/tc into tc/ and same for subfolders of src/

addresses https://github.com/facebookresearch/TensorComprehensions/issues/194

cc @mingzhe09088  (couldn't tag as reviewer hence cc'ing)


prigoyal:
@salexspb, I will make the -Werror change in a separate PR since that also involves testing some configuration changes on CI. This PR is ready for review and the incremental change can be landed if you want.

nicolasvasilache:
Re hotfix see #281

src/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ include_directories(.)
 include_directories(../include)
 
 add_subdirectory(lang)
-add_subdirectory(proto)
+#add_subdirectory(proto)
```
nicolasvasilache: dead code

tc/aten/aten_compiler.h
```diff
@@ -22,9 +22,9 @@
 #include <ATen/ATen.h>
 #include <ATen/DLConvertor.h>
 
-#include "tc/aten/utils.h"
-#include "tc/core/execution_engine.h"
-#include "tc/lang/parser.h"
+#include "aten/utils.h"
+#include "core/execution_engine.h"
```
salexspb: Actually having includes starting from "tc", in my opinion, is what we should be targeting. this way we provide a proper isolation between tc and any other libraries when TC is included as a submodule or is included in an FBCode fashion. 


Also number of changed files would be significantly smaller in this case :)

tc/core/cpu/cpu_mapping_options.h
```diff
@@ -17,7 +17,7 @@
 
 #include <mapping_options.pb.h>
```
salexspb: this one should be, IMO, tc/proto/mapping_options.pb , but we can do it in a separate PR, of course.

tc/aten/aten_compiler.h
prigoyal: I agree. I'll address this and update PR. There will be cmake change needed only for this then. Thanks.

prigoyal:
I ran `test_python/run_test.sh` and `test.sh` both and they pass

@salexspb , if you could help with review to avoid more merge conflicts, I would really appreciate that. NOTE: the Jenkins CI is having downtime atm so running local test is best thing to do right now.

prigoyal:
yep, I ran `test_python/run_test.sh` and `test.sh` both and they pass. Thanks @salexspb :)

PR #280, opened 2018-04-09
Expose option for saving # best candidates in autotuning

### LINKED ISSUE #243: Expose the setting for saving # best candidates in autotuning
Currently, by default, we save 10 best candidates in autotuning and number `10` is also hardcoded here https://github.com/facebookresearch/TensorComprehensions/blob/master/src/autotuner/genetic_autotuner.cc#L60

Recently, some friendly users have asked if it is possible to rather set the number of candidates or maybe even save all i.e. can we expose this setting to users?

IMO, this seems like a legit change but I would like to hear views from @ttheodor if we should do this or not.  The change is quite simple. Set the gflag for it with default value 10 and then expose this flag to the pybind as well.

cc @sidelnik

prigoyal:
closes #243 

*NOTE:* This PR is stacked on modernization PR branch `modernize-tc` hence merge it only after modernization PR lands.

### DIFF
--- a/tc/autotuner/genetic_autotuner.cc
+++ b/tc/autotuner/genetic_autotuner.cc
@@ -57,7 +57,8 @@ void GeneticAutotuner::storeCaches(const std::string& filename) {
   } else {
     std::cout << "Dumping cache to " << filename << ".cuda/options"
               << std::endl;
-    tc::OptionsCache::getCache()->keepOnlyBestCandidates(10);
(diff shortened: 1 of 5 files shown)


ttheodor:
It still looks good but needs rebasing.

prigoyal:
@ttheodor , rebased PR. your review is needed again whenever you have time :) 

I ran `test.sh` and `test_python.sh` both and test pass. Jenkins CI is still not functional atm

PR #281, opened 2018-04-09
[Hotfix] fix python tests

nicolasvasilache:
This cherry-picks 479c78d0367b823bb7c61b7a8e8ee8dd385e4902 to fix the python side of things for which the proto disappeared.
This was dead C++ code that got removed but Jenkins being down over the weekend it turns out that master was broken.


nicolasvasilache:
Redundant with #282

PR #283, opened 2018-04-09
Don't split reductions to remove nested thread mapping

ftynse:
Reduction detection mechanism splits out the band containing the reduction loop into a separate band in the tree.  As a result, mapping to threads was made recursive to allow for mapping both the reduction update band and its parent band node, although mapping nested bands to threads is invalid in the general case.  However, it is necessary to split out the reduction loop in order to insert the reduction synchronization call.

Split out the reduction loop immediately before inserting the reduction synchronization.  During thread mapping, keep the reduction loop inside its original band.  This avoids the need to keep track of the relation between the reduction band and its parent, which can be replaced by position of the reduction loop in the band.  As memory promotion may split the band to insert copies, insert reduction synchronizations before promotion passes.

mapInnermostBandsToThreads maps coincident band members to threads while traversing the tree recursively in DFS postorder.  It is possible that members of two nested bands get mapped to thread dimensions, which was necessary for parallel reduction support.  The previous commit keeps the reduction in its original band, removing the need for mapping nested bands. Such mapping is not valid in general because members of different bands are not necessarily permutable with each other.  Since thread dimensions are not executed in any specific order, i.e. there is no guarantee that, for a given value of threadIdx.y, all threadIdx.x threads will be executed before (or after) those with another value of threadIdx.y, permutability is a necessary condition for mapping to multiple thread dimensions.  Only allow mapping a band to threads if its children are not already mapped to threads.  The procedure continues recursively so as to insert all relevant synchronization between children of sequence nodes even after the thread mapping is complete.

Closes #263

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -197,6 +197,11 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
     return found;
   }
 
+  // Only reductions that appear in permutable bands are mapped to threads.
+  if (!band->permutable_) {
+    return false;
+  }
+
   // For now, only support reductions with a sufficient number
   // of coincident outer band members for the remaining thread identifiers.
   auto nCoincident = band->nOuterCoincident();
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -367,8 +367,8 @@ size_t MappedScop::mapToThreads(detail::ScheduleTree* band, size_t nInner) {
   }
   // With current isl scheduler, if coincident dimensions exist in a band,
   // they are outermost.
-  // If a band has more than 3 coincident dimensions, this will choose
-  // outermost, but we may also want innermost.
+  // If a band has more than 3 coincident dimensions,
+  // then the innermost of those will be used.
   auto nOuterCoincident = bandNode->nOuterCoincident();
   if (!bandNode->permutable_ || nOuterCoincident < 1) {
     return nInner;
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -230,7 +230,6 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
   if (!isReductionMember(member, updates, scop())) {
     return false;
   }
-  auto reductionTree = bandSplitOut(scop_->scheduleRoot(), tree, reductionDim);
   // Order the init statements (if any) before the update statements
   // to ensure the band from which the reduction band has been split off
   // only contains update statements.
@@ -238,57 +237,43 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
   if (!inits.is_empty()) {
     orderBefore(scop_->scheduleRoot(), tree, inits);
   }
-  reductionFromParent_.emplace(tree, reductionTree);
-  reductionBandUpdates_.emplace(reductionTree, updateIds);
+  reductionBandUpdates_.emplace(tree, Reduction(updateIds, reductionDim));
   return true;
 }
 
 bool MappedScop::needReductionSeparation(const detail::ScheduleTree* st) {
-  // It is the parent band of the reduction band that needs to be separated.
-  if (reductionFromParent_.count(st) != 1) {
+  if (reductionBandUpdates_.count(st) != 1) {
     return false;
   }
-  st = reductionFromParent_.at(st);
-  CHECK(reductionBandUpdates_.count(st) == 1);
   // No need to separate if already separated.
   return !reductionBandUpdates_.at(st).separated;
 }
 
 isl::multi_union_pw_aff MappedScop::reductionMapSchedule(
     const detail::ScheduleTree* st) {
-  CHECK(reductionFromParent_.count(st) == 1);
-  auto parent = st;
-  st = reductionFromParent_.at(st);
   CHECK(reductionBandUpdates_.count(st) == 1);
-
   auto reductionBand = st->elemAs<detail::ScheduleTreeElemBand>();
   CHECK(reductionBand);
-  // Start with the schedule of the reduction band (in last position).
-  auto reductionSchedule = reductionBand->mupa_;
 
-  // Total size of returned schedule needs to be equal
-  // to the number of thread identifiers.
-  if (numThreads.view.size() > 1) {
-    CHECK(parent != st);
-  }
-  // Prepend last members of parent band (if any).
-  if (parent != st) {
-    auto parentBand = parent->elemAs<detail::ScheduleTreeElemBand>();
-    CHECK(parentBand);
-    auto parentSchedule = parentBand->mupa_;
-    auto nMember = parentBand->nMember();
-    CHECK_GE(nMember, numThreads.view.size() - 1);
-    parentSchedule = parentSchedule.drop_dims(
-        isl::dim_type::set, 0, nMember - (numThreads.view.size() - 1));
-    reductionSchedule = parentSchedule.flat_range_product(reductionSchedule);
-  }
+  // Drop band members following the reduction dimension and preceding those
+  // mapped to threads.
+  auto reductionSchedule = reductionBand->mupa_;
+  auto nMember = reductionBand->nMember();
+  auto reductionDim = reductionBandUpdates_.at(st).reductionDim;
+  auto nMappedThreads =
+      std::min(numThreads.view.size(), reductionBand->nOuterCoincident() + 1);
+  CHECK_GE(nMember, reductionDim);
+  CHECK_GE(reductionDim + 1, nMappedThreads);
+  reductionSchedule = reductionSchedule.drop_dims(
+      isl::dim_type::set, reductionDim + 1, nMember - (reductionDim + 1));
+  reductionSchedule = reductionSchedule.drop_dims(
+      isl::dim_type::set, 0, reductionDim - nMappedThreads + 1);
 
   return reductionSchedule;
 }
 
 detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
-  CHECK(reductionFromParent_.count(st) == 1);
-  auto reduction = reductionFromParent_.at(st);
+  auto reduction = st;
   // This function either separates full blocks (if needed) or
   // disables the reduction handling.
   reductionBandUpdates_.at(reduction).separated = true;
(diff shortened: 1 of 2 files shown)


src/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -220,65 +220,52 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
...
   // Order the init statements (if any) before the update statements
   // to ensure the band from which the reduction band has been split off
   // only contains update statements.
   // Note that this relies on the outer members being coincident.
   if (!inits.is_empty()) {
     orderBefore(scop_->scheduleRoot(), tree, inits);
   }
-  reductionFromParent_.emplace(tree, reductionTree);
-  reductionBandUpdates_.emplace(reductionTree, updates);
+  reductionBandUpdates_.emplace(tree, Reduction(updates, reductionDim));
```
skimo-openhub: Update documentation of `detectReductions` corresponding to these changes.

src/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -220,65 +220,52 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
...
   CHECK(reductionBandUpdates_.count(st) == 1);
-
   auto reductionBand = st->elemAs<detail::ScheduleTreeElemBand>();
   CHECK(reductionBand);
-  // Start with the schedule of the reduction band (in last position).
-  auto reductionSchedule = reductionBand->mupa_;
 
   // Total size of returned schedule needs to be equal
   // to the number of thread identifiers.
-  if (numThreads.view.size() > 1) {
```
skimo-openhub: The comment above is about the removed code, so it should presumably be removed too.

src/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -329,10 +316,8 @@ detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
 size_t MappedScop::mapToThreads(detail::ScheduleTree* band, size_t nInner) {
   using namespace tc::polyhedral::detail;
 
-  if (nInner >= numThreads.view.size()) {
-    return nInner;
-  }
```
skimo-openhub: Why did you remove this test?
In fact, it should be strengthened to `nInner > 0`, no?

src/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -329,10 +316,8 @@ detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
...
-  if (reductionBandUpdates_.count(band) == 1) {
+  bool mapReduction = reductionBandUpdates_.count(band) == 1;
+  if (mapReduction) {
     // A reduction is assumed to get mapped to threadIdx.x
     if (nInner != 0) {
```
skimo-openhub: This test can then be removed.

skimo-openhub:
On Tue, Apr 10, 2018 at 01:53:28AM -0700, skimo-openhub wrote:
> skimo-openhub commented on this pull request.

I keep forgetting that github doesn't seem to keep track
of the commit that a review corresponds to.
This was cdb9c07d6af6542efd3ecfcd6c3977f836d5da4f.

skimo

src/core/polyhedral/cuda/mapped_scop.cc
ftynse: `nInner` is completely removed in the following commit, I may have picked the wrong change

src/core/polyhedral/cuda/mapped_scop.cc
ftynse: It will be removed in the next commit.  The first one only deals with separating the reduction band without changing how the threads are mapped in general.

ftynse:
updated

ftynse:
rebased

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -225,65 +225,50 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
...
+  auto nMember = reductionBand->nMember();
+  auto reductionDim = reductionBandUpdates_.at(st).reductionDim;
+  auto nMappedThreads =
+      std::min(numThreads.view.size(), reductionBand->nOuterCoincident() + 1);
+  CHECK_GE(nMember, reductionDim);
+  CHECK_GE(reductionDim + 1, nMappedThreads);
+  reductionSchedule = reductionSchedule.drop_dims(
+      isl::dim_type::set, reductionDim + 1, nMember - reductionDim - 1);
+  reductionSchedule = reductionSchedule.drop_dims(
+      isl::dim_type::set, 0, reductionDim - nMappedThreads + 1);
```
skimo-openhub: nit: I would put the `+ 1` together with `reductionDim` and
probably also write `-(reductionDim + 1)` instead of `- reductionDim - 1`.
Note that I will probably end up rewriting these lines anyway in the not too distant future (because of the `isl::dim_type::set`).

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -587,6 +575,22 @@ std::tuple<std::string, tc::Grid, tc::Block> MappedScop::codegen(
       mappedScopForCodegen->numThreads);
 }
 
+// Split out reduction loops into separate bands and insert reduction
+// synchronizations after those bands.
```
skimo-openhub: "after" is confusing.  It's inserted _outside_ of those bands, isn't it?

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -587,6 +575,22 @@ std::tuple<std::string, tc::Grid, tc::Block> MappedScop::codegen(
...
+void MappedScop::insertReductionSyncs() {
```
skimo-openhub: Maybe rename the method to reflect the two operation?
`splitOutReductionsAndInsertSyncs`?

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -316,62 +312,49 @@ detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
...
-        band->child({0}), nInner, numThreads.view.size());
-    scop_->insertSyncAfter(band->child({0}));
-    return numThreads.view.size();
-  }
   // With current isl scheduler, if coincident dimensions exist in a band,
   // they are outermost.
   // If a band has more than 3 coincident dimensions, this will choose
   // outermost, but we may also want innermost.
   auto nOuterCoincident = bandNode->nOuterCoincident();
+  int firstThreadIndex = mapReduction ? 1 : 0;
```
skimo-openhub: I find this name (`firstThreadIndex`) confusing.
It seems to indicate the _number_ of threads already used.
I'd either put it in `nMappedThreads` or in a new `nMappedReductionThreads`.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -316,62 +312,49 @@ detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
...
   if (!bandNode->permutable_ || nOuterCoincident < 1) {
-    return nInner;
+    return firstThreadIndex;
   }
 
   auto nMappedThreads = std::min(
-      numThreads.view.size() - nInner, static_cast<size_t>(nOuterCoincident));
-  CHECK_GT(nMappedThreads, 0) << "not mapping to threads";
```
skimo-openhub: You removed this check.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -316,62 +312,49 @@ detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
...
-    CHECK_EQ(band->numChildren(), 1);
-    mapRemaining<mapping::ThreadId>(
-        band->child({0}), nInner, numThreads.view.size());
-    scop_->insertSyncAfter(band->child({0}));
-    return numThreads.view.size();
-  }
   // With current isl scheduler, if coincident dimensions exist in a band,
   // they are outermost.
   // If a band has more than 3 coincident dimensions, this will choose
   // outermost, but we may also want innermost.
```
skimo-openhub: This comment looks outdated (prior to this series).
This series would be a good time to fix it, because the reasoning is hard to follow now.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -316,62 +312,49 @@ detail::ScheduleTree* MappedScop::separateReduction(detail::ScheduleTree* st) {
...
   auto nOuterCoincident = bandNode->nOuterCoincident();
+  int firstThreadIndex = mapReduction ? 1 : 0;
   if (!bandNode->permutable_ || nOuterCoincident < 1) {
-    return nInner;
+    return firstThreadIndex;
```
skimo-openhub: I think this deserves a bit more explanation.
It is only correct because `MappedScop::detectReductions` has already done
the required checks.  Actually, it hasn't completely.
`MappedScop::detectReductions` doesn't have the permutable_` check.
That should be fixed first.

tc/core/polyhedral/cuda/mapped_scop.cc
ftynse: I replaced it by early return, because `nMappedThreads` can now legitimately be zero in case when reduction was mapped.

ftynse:
> Now that we only map a single band, we might as well complete the mapping immediately instead of waiting until we reach the parent, but this can be done in a follow-up.

I also thought about that.  Updated.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -467,20 +467,22 @@ size_t MappedScop::mapInnermostBandsToThreads(detail::ScheduleTree* st) {
   if (auto band = st->elemAs<detail::ScheduleTreeElemBand>()) {
     if (n == 0) {
       // If children were not mapped to threads, the current band can be mapped.
+      // First, map the coincidence and reduction dimension to threads.
+      // The fix unused thread dimensions to 0 because we cannot map parent
+      // bands anyway.
       auto nMapped = mapToThreads(st);
+      mapRemaining<mapping::ThreadId>(st, nMapped, numThreads.view.size());
```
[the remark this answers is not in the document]
ftynse: Good point about "if anything gets mapped", but I'm not convinced it should be inside `mapToThreads`.  The function has two exit points and we'd have to eventually map to remaining at both.

ftynse:
Thanks, running tests...

ftynse:
tests pass, merging

PR #284, opened 2018-04-09
Enable Werror compiler flag

prigoyal:
enabling `Werror` required as part of modernization task #194 

this should run by default on CI as well now.

also cc @mingzhe09088 (couldn't tag as reviewer hence cc'ing)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,13 @@ project (TensorComprehensions C CXX)
 
 # Display compiler flags for build modes: Release/Debug
 message(STATUS "BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+message(STATUS "CMAKE_CXX_COMPILER_VERSION is ${CMAKE_CXX_COMPILER_VERSION}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "5.4")
+  # we officially support these compiler flags for compiler version 5.4 or less
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable -Werror")
+endif()
+message(STATUS "CMAKE_CXX_FLAGS is ${CMAKE_CXX_FLAGS}")
 if (${CMAKE_BUILD_TYPE} MATCHES "Debug")
   set(CMAKE_CXX_FLAGS_DEBUG "-g")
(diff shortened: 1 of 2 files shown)


include/tc/lang/lexer.h
```diff
@@ -418,9 +418,9 @@ struct Lexer {
 
  private:
   Token lex() {
-    int kind;
-    size_t start;
-    size_t length;
+    int kind = -1;
```
prigoyal: I initialized these values by looking at other contructors for SourceRange in the code and also ran the C++/python test afterwards. @zdevito , is this safe to do?

ttheodor:
`-Werror` shouldn't be enabled by default but only optionally (e.g., on CI): different compiler versions emit different warnings.

prigoyal:
@salexspb what compiler version do you want this enabled on? I can then make the relevant changes. thanks

ttheodor:
@prigoyal  the problem is that as soon as anyone (developer or users) uses a newer/different compiler that emits more/different warnings (which are promoted to errors due to `-Werror`), they won't be able to build the project. (For example, with GCC 7.X.X you can't build this branch).

So we either support very specific compilers (and versions) or we make `-Werror` optional. 

Just to be clear, I'm not saying that we should ignore warnings, we should fix them (I already suggested  that we should enable `-Wall -Wpedantic -Wextra`). But making them hard errors is just annoying.

prigoyal:
sounds good @ttheodor . How about enabling for gcc 5.4 and 4.8.4 i.e the ones we support on CI as well? If we need this enabled for more compiler versions, then someone should fix the warnings and enable. what do you think?

ttheodor:
@prigoyal sounds reasonable

CMakeLists.txt
```diff
@@ -16,6 +16,12 @@ project (TensorComprehensions C CXX)
 
 # Display compiler flags for build modes: Release/Debug
 message(STATUS "BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
+message(STATUS "CMAKE_CXX_COMPILER_VERSION is ${CMAKE_CXX_COMPILER_VERSION}")
+if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "5.4")
```
prigoyal: @ttheodor , what do you think of this? I didn't put specific version thinking that 4.9 for example could also work. for example 4.8.4 and 4.8.5 are both okay.

ttheodor: LG, but I'd move `-Wall` outside the conditional.

prigoyal:
@ttheodor  I have rebased the PR and put the `-Wall` outside compiler version conditioning. Ready for review again :)

PR #289, opened 2018-04-11
Fix a type inconsistency issue in codegen_llvm.

dingobye:
This PR fixes a type inconsistency issue in _codegen_llvm_.

Such issue is observed after I added a test case for [concat from layers database](https://facebookresearch.github.io/TensorComprehensions/framework/pytorch_integration/layers_database.html#concat), 
```
def concat(float(M, N) A, float(M, N) B) -> (O1) {
    O1(n, i, m) = i == 0 ? A(m, n) : B(m, n) where i in 0:2
}
```
and the following assertion failure occurs:
_.../include/llvm/IR/Instructions.h:1113: void llvm::ICmpInst::AssertOK(): Assertion `getOperand(0)->getType() == getOperand(1)->getType() && "Both operands to ICmp instruction are not of the same type!"' failed._

It happens to the loop induction variable _i_, corresponding to a PHINode of i64. It is directly used to compare with a i32 constant value (i.e. 0), where inconsistency of two operands' types takes place. 

This patch adds proper type casting to fix this issue, and adds a test cast for concat.

### DIFF
--- a/tc/core/polyhedral/codegen_llvm.cc
+++ b/tc/core/polyhedral/codegen_llvm.cc
@@ -252,6 +252,10 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
     if (value->getType() != ty) {
       if (op->type.is_int()) {
         value = builder->CreateIntCast(value, ty, true);
+      } else if (op->type.is_uint()) {
+        value = builder->CreateIntCast(value, ty, false);
+      } else if (op->type.is_float()) {
+        value = builder->CreateFPCast(value, ty);
       } else {
         CHECK(false) << "Type inconsistency not handled. "
                      << "Variable " << op->name << " is " << op->type


tc/core/polyhedral/codegen_llvm.cc
```diff
@@ -248,8 +248,22 @@ class CodeGen_TC : public Halide::Internal::CodeGen_X86 {
...
+
   void visit(const Halide::Internal::Variable* op) override {
     value = getValue(iteratorMap_->at(op->name));
+
+    // Generate code for type casting if necessary.
+    llvm::Type* ty = llvm_type_of(op->type);
+    if (value->getType() != ty) {
+      if (op->type.is_int()) {
+        value = builder->CreateIntCast(value, ty, true);
+      } else {
```
wsmoses: Can you also add the corresponding floating point cast, not necessarily needed for your test case, but useful to have.

dingobye:
Hi William, thanks for your review. I updated the code as you suggested.

wsmoses:
@caffe2bot retest

dingobye:
rebased

PR #295, opened 2018-04-11
Move protobufs generated files to tc/proto/*.{h/cc}

prigoyal:
Motivation: posting excerpt from @salexspb below:

```
Currently in OSS includes are done as following:

#include "mapping_options.pb.h"

This is bad because in a bigger codebase it can clash with other files. Also it requires manual include paths which complicate the project, people don't know where to look for files, etc.

Solution we discussed is  to have tc/proto/*.proto files and then have protoc / cmake generate .cc and .h files in a way that .h is being includes with full path as tc/proto/mapping_options.pb.h

This would also unblock fbCode development. Where currently we have to fix the OSS code when doing a sync and store protobufs in a separate from OSS folder
```

### DIFF
--- a/tc/proto/CMakeLists.txt
+++ b/tc/proto/CMakeLists.txt
@@ -60,7 +60,7 @@ function(tc_protobuf_generate_cpp_py output_dir srcs_var hdrs_var python_var)
   set(${python_var} ${${python_var}} PARENT_SCOPE)
 endfunction()
 
-tc_protobuf_generate_cpp_py(${CMAKE_CURRENT_BINARY_DIR} PROTO_SRCS PROTO_HDRS PROTO_PY mapping_options.proto compcache.proto)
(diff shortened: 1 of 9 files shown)


prigoyal:
I ran `test.sh` and `test_python.sh`. Jenkins CI is still not working.

tc/proto/CMakeLists.txt
```diff
@@ -42,7 +42,6 @@ function(tc_protobuf_generate_cpp_py output_dir srcs_var hdrs_var python_var)
     list(APPEND ${srcs_var} "${output_dir}/${fil_we}.pb.cc")
     list(APPEND ${hdrs_var} "${output_dir}/${fil_we}.pb.h")
     list(APPEND ${python_var} "${output_dir}/${fil_we}_pb2.py")
-
```
ftynse: spurious change?

prigoyal:
thanks @ftynse , the spurious change happened due to some logging I was doing. It has been corrected. requesting your kind review again. Thanks. :)

PR #301, opened 2018-04-13
do not promote to registers before promoting to shared in tests

ftynse:
MapperMemoryPromotionRAW and Sum4D tests call the mapping strategy with
promotion to private enabled, which may insert copies to private, and
promotion to shared disabled.  They then call promotion to shared
separately through a dedicated call to avoid using the main heuristic
dependent on the shared memory size available on the current device.
Therefore, promotion to shared happens after promotion to private, and
neither expects that.  In particular, mapCopiesToThreads is a part of
the promotion to shared and expects to be called before the promotion to
private happens.  Do not request promotion to private in the tests that
do not check for it.


ftynse:
This PR only changes a CPU test that ran on CircleCI, GPU tests cannot be affected. Merging.

PR #303, opened 2018-04-13
MappedScop::mapToThreads: fix depth of threadIdxX in case of reduction

skimo-openhub:
Commit 88764d37 (split out reduction band after thread mapping
to remove nested mapping, Mon Apr 9 17:43:18 2018 +0200) removed
the condition that the reduction member always appears first
in its band at the point of MappedScop::mapToThreads.
The actual mapping was updated to take into account the depth
inside the band, but the threadIdxXScheduleDepthState was
not updated accordingly.


skimo-openhub:
@caffe2bot retest this please

ftynse:
local test passed /@ftynse

PR #304, opened 2018-04-13
fix typos in comments

skimo-openhub:


tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -357,7 +357,7 @@ size_t MappedScop::mapToThreads(detail::ScheduleTree* band) {
 
   // Immediately return if mapping to one thread dimension only was requested
   // and a reduction was already mapped.  (Note that reduction is detected only
-  // if there are not enough outer coincident members, 0 in this case).
+  // if there are enough outer coincident members, 0 in this case).
```
ftynse: I'm not sure about this one. AFAIU, reduction it _not_ detected if there is 1 coincident member followed by 1 reduction member, and mapping to only 1 thread dimension is required.  It _is_ detected if mapping to 2 thread dimensions is required.  That's what I interpreted as having _not enough_ coincident members.  Maybe it's better to reformulate this as "reduction is detected only if there are less coincident dimensions than thread dimensions requested for mapping" to avoid confusion.

skimo-openhub:
On Fri, Apr 13, 2018 at 07:51:11AM -0700, ftynse wrote:
> > @@ -357,7 +357,7 @@ size_t MappedScop::mapToThreads(detail::ScheduleTree* band) {
>  
>    // Immediately return if mapping to one thread dimension only was requested
>    // and a reduction was already mapped.  (Note that reduction is detected only
> -  // if there are not enough outer coincident members, 0 in this case).
> +  // if there are enough outer coincident members, 0 in this case).
> 
> I'm not sure about this one. AFAIU, reduction it _not_ detected if there is 1 coincident member followed by 1 reduction member, and mapping to only 1 thread dimension is required.

Why?  This should be detected as a reduction.

>  It _is_ detected if mapping to 2 thread dimensions is required.  That's what I interpreted as having _not enough_ coincident members.  Maybe it's better to reformulate this as "reduction is detected only if there are less coincident dimensions than thread dimensions requested for mapping" to avoid confusion.

No, it's the other way around.  The number of coincident dimensions
needs to be at least numThreads.view.size() - 1:

  // For now, only support reductions with a sufficient number
  // of coincident outer band members for the remaining thread identifiers.     
  auto nCoincident = band->nOuterCoincident();
  if (nCoincident < numThreads.view.size() - 1) {
    return found; // false here
  }

skimo

skimo-openhub:
@caffe2bot retest this please

ftynse:
>> I'm not sure about this one. AFAIU, reduction it _not_ detected if there is 1 coincident member followed by 1 reduction member, and mapping to only 1 thread dimension is required.

> Why?  This should be detected as a reduction.

Because the code you quoted below will return false.

>> Maybe it's better to reformulate this as "reduction is detected only if there are less coincident dimensions than thread dimensions requested for mapping" to avoid confusion.

> No, it's the other way around.  The number of coincident dimensions
needs to be at least numThreads.view.size() - 1:

I'm not convinced.  If the number of coincident members is _at least_ "number of thread dimensions - 1" for a reduction to be detected, it means that if there are _more_ coincident members than thread dimension, the reduction _will_ be detected.  The code you quoted does the inverse by returning false in this case...

PR #309, opened 2018-04-16
Proposed API changes for ExecutionEngine

nicolasvasilache:
The proposed changes to the API involve the following main things:
1. an explicit Backend type which keeps track of the concrete Executor, MappingOptions, Protobuf, Cache and RTCFunction types. By making types dependent it becomes significantly easier to write a type-safe, backend-agnostic JIT compiler and factor out the common behaviors across the Autotuner, the ExecutionEngine, the Executor and the Mapper and have them interact properly;
2. inferOutputTensorInfo now returns a TensorInfo object. In the current master, this returns a pointer backed by an underlying executor that needs to persist in the executors_ even though it is not yet compiled; clearly not great;
3. run returns a pair<Duration, Duration> which measure the CPU and kernel performance (when profileKernel == true). uncheckedRun returns the kernel time (when profileKernel == true). This will allow more easily to root out CPU overhead issues.
4. uncheckedRun now explicitly takes const void* and void*
5. inputs are now of type DLConstTensor to properly implement const-correctness cc @salexspb (RFC to follow)
6. thanks to 1. MappingOptions can now be properly typed and we don't need to resort to string serialization
7. clearer documentation with an example of usage

There is a currently working implementation here but it still needs work and cleanup. In particular, it also needs slicing up so that people can review more easily.
To facilitate the slicing work on the PR above, let's please have people comment on the proposed API changes and either signoff or proposed changes and iterate until we reach consensus.
Once consensus is reached on a particular piece of the API, independent commits can be created and sent for review.

### DIFF
--- a/tc/aten/aten_compiler_api_rfc.h
+++ b/tc/aten/aten_compiler_api_rfc.h
@@ -1,0 +1,79 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <chrono>
+#include <string>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/tree.h"
+
+namespace tc {
+/// Run inference on the TC tree resulting from parsing a TC function and
+/// applied to the specified input shapes.
+/// This is used for manual processing of outptu metadata on the ATen user
+/// side.
+/// \returns the (contiguous) output tensor shapes as a metadata-owning
+/// DLTensorUPtr.
+std::vector<DLTensorUPtr> inferOutputTensorInfo(
+  lang::TreeRef tcDefinition,
+  const std::vector<at::Tensor>& inputs);
+
+/// Allocate fresh new output tensors.
+/// If one wants inplace/resize behavior, one can implement it using
+/// inferOutputTensorInfo.
+/// \returns the (contiguous) output tensors with properly inferred sizes.
+std::vector<at::Tensor> prepareOutputs(
+  lang::TreeRef tcDefinition,
+  const std::vector<at::Tensor>& inputs);
+
+/// Given a TC tree resulting from parsing a TC function, compile the TC for
+/// the specified input shapes with the prescribed options.
+/// \returns an Executor for the specified backend with an underlying
+/// JIT-compiled callable function.
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    lang::TreeRef tcDefinition,
+    const std::vector<at::Tensor>& inputs,
+    const typename Backend::MappingOptionsType& options);
+
+/// Given an executor resulting from compiling a TC, run the TC and fill the
+/// outputs vector with the results.
+/// \returns ProfilingInfo (cpuOverhead + kernelRuntime) in microseconds.
+/// If profile=false, kernelRuntime is set to Duration::max.
+template <typename Executor>
+ProfilingInfo run(
+  const Executor& executor,
+  const std::vector<at::Tensor>& inputs,
+  std::vector<at::Tensor>& outputs,
+  bool profile = false);
+
+/// This is the "low-latency" mode in which we just propagate ATen tensors
+/// Sizes are not checked and it is the user's responsibility to ensure that
+/// they match. If the user doesn't then segfault will likely occur.
+/// \returns the kernel runtime in microseconds if profile=true.
+/// If profile=false, kernelRuntime is set to Duration::max.
+template <typename Executor>
+Duration uncheckedRun(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs,
+    bool profile = false);
+} // namespace tc
+
+#include "tc/aten/aten_compiler-inl.h"
--- a/tc/core/execution_engine_api_rfc.h
+++ b/tc/core/execution_engine_api_rfc.h
@@ -1,0 +1,147 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "tc/core/compilation_cache.h"
+#include "tc/core/tc_executor.h"
+#include "tc/core/tensor.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/tree.h"
+
+namespace tc {
+
+/**
+ * The ExecutionEngine is the entry point to using TC in C++.
+ * It provides simple backend-agnostic capabilities to:
+ *   1. construct the tree IR for parsed TCs from
+ *      a string representation and store (a shared_ptr to) it
+ *   2. infer actual tmp/output tensor shapes given input tensor shapes;
+ *   3. compile and run TCs on backend-specific TcExecutors;
+ *   4. store the TcExecutors for memoization of compilations.
+ *      The correspondance is:
+ *      1 TcExecutor <-> 1 tuple<TC function, input shapes, MappingOptions>
+ *
+ * ExecutionEngine is templated by the Backend type.
+ * For each backend, the specific Backend type lives in
+ *   backend/backed_tc_executor.h and declares all the required dependent
+ *   **concrete** types.
+ * For example:
+ *   CudaBackend is declared in core/cuda/cuda_tc_executor.h
+ *
+ * struct CudaBackend {
+ *   using ExecutorType = CudaTcExecutor;
+ *   using MappingOptionsType = CudaMappingOptions;
+ *   using ProtobufType = CudaCacheProto;
+ *   using CacheType = CudaCache;
+ *   using CacheEntryType = CudaCacheEntry;
+ *   using CacheKeyType = typename CacheEntryType::Key;
+ *   using CacheValueType = typename CacheEntryType::Values;
+ *   using RTCFunctionType = CudaRTCFunction;
+ * };
+ *
+ * Sketching usage resembles:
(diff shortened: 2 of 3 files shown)


tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,149 @@
...
+class ExecutionEngine {
+ public:
+  using BackendType = Backend;
+  using ExecutorType = typename BackendType::ExecutorType;
+  using MappingOptionsType = typename BackendType::MappingOptionsType;
+
+  ExecutionEngine() = default;
+
+  /// Parse TC definitions provided as string, store parsed trees internally.
+  void define(const std::string& language);
```
ftynse: I would provide some higher-level description of what `define` does.  If we say it parses and stores something, it should be called `parseAndStore`...

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,149 @@
...
+
+  /// Store the provided parsed trees internally.
+  void define(const std::vector<lang::TreeRef>& treeRefs);
+
+  /// For a given TC kernel name, compute the shapes of the output tensors
+  /// provided the shapes of the input TC tensors.  The caller can use the
+  /// computed shapes to allocate memory for outputs.
+  std::vector<TensorInfo> inferOutputTensorInfo(
```
ftynse: `TensorInfo` is never defined.  It is a struct from the autotuner IIRC, not sure we want to carry it here.  Why not `DLTensor` ?

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,149 @@
...
+  /// For thread-safety perform all cheap operations under lock.
+  std::mutex tcExecutorMutex_;
+
+  /// Parsed TC trees.
+  std::map<std::string, lang::TreeRef> tcNameMap_;
+
+  /// List of executors, indexed by handle.
+  std::vector<std::unique_ptr<ExecutorType>> executors_;
+
+  size_t nextHandle = 0;
```
ftynse: Isn't next handle just `executors_.size()` ?

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,149 @@
...
+  /// shapes (inputs same as for copmilation, outputs same as returned by
+  /// inferOutputTensorInfo).
+  /// \returns The pair of (cpu,kernel) runtimes. If profile is set, the
+  /// kernel is profiled otherwise its duration is set to Duration::max().
+  std::pair<Duration, Duration> run(
+      size_t handle,
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs,
+      bool profileKernel = false,
+      std::function<bool(const ExecutorType*)> pruningFunction =
```
ftynse: I never really liked this callback here.  It's only use is in the autotuner deciding whether to run something or not.  It does not really make sense in the context of a single-short `run`, the caller should just decide.  Same for the autotuner by the way.  The logic seems inverted, but the caller just don't have enough information to take the decision.  See my proposal in a separate comment below.

ftynse:
My proposal for the API is to replace the opaque `size_t` compilation handle by a (potentially wrapped for python compatibility) non-owning const pointer to the actual executor.  The caller can thus query the executor directly and decide whether they want to run or not, and do other useful things.  For example, given an appropriate function at the executor level, it becomes possible to save the generated code.  This requires slightly moving the templating logic: instead of a separate `Backend` struct with typedefs, `ExecutionEngine` is parameterized by an `Executor` that includes all relevant typedefs.

Pushing this further, we can separate `ExecutionEngine` from `Executor`, the former essentially becoming `CompilationEngine`.  The engine handles TC parsing and analysis, calls the polyhedral machinery and post-processing up until it obtains a runnable object, e.g. `CudaRTCFunction`.  This object is wrapped together with other useful information into an `Executor` class.  The engine _does not_ have `run` functions at all, only executors do.

There are some thread-safety considerations embedded into the design of `ExecutionEngine`, but they seem to had been driven by how the autotuner works or used to work.  Unique pointers to executors are used, AFAIR, for their single-thread-ownership guarantees.  Practically, I don't see a problem in engine creating _copies_ of the executor _without recompiling_.  So each thread can have its own copy, if necessary, instead of the "steal of possible" logic currently implemented.  There is an overhead of copying the data though.

While this proposal seems to increase the number of objects in the API, I claim that it achieves a nice separation.  In fact, `ExecutorType` is already exposed in the API and as an argument of `run` for pruning, I only make it a first-class citizen. Conceptually, this would separate compilation and caching from running, clearly stating that you can compile once and run multiple times without recompiling.  Practically, it would remove

tc/core/execution_engine_api_rfc.h
ttheodor: One problem with `DLTensor` is memory management: it's using raw pointers to store strides/shapes.

ttheodor:
> Unique pointers to executors are used, AFAIR, for their single-thread-ownership guarantees. 

Actually they were originally introduced as a work-around due to (a misuse of) ISL: `std::vector<T>::push_back(...)` every so often relocates the stored objects. It was these relocation that left the executors in an invalid state.

ftynse:
> Can't we take it a step further and return the actual executor?

Yeah, that's exactly what I proposed in the paragraphs 2-4 :)

nicolasvasilache:
@ftynse @ttheodor the suggested change is more than I had set up to do originally with a first bottom-up refactoring to decouple backends and caches but after playing a bit more, this actually may be a nice opportunity to go even further and drop a bunch of useless classes. 3c2a412 is what going towards a decoupled compilation and executor may look like once percolated up at the ATen level and 819640a is sketching the minimal TC functions required. The implementation itself looks quite reasonable, I still need to do some work to make tests pass but after spending some time exploring this avenue it seems quite doable.

Note that I prefer to keep UPtr rather than by value executor for now until we have something stable we agree on.

If the ATen/TC high-level API looks reasonable as is, I can push more to get the tests passing. In that case what will probably happen is what remains of the execution engine will probably be merged into the autotuner.

ftynse:
> the suggested change is more than I had set up to do originally with a first bottom-up refactoring

That's why I suggested to do two separate changes in the yesterday's call.

> Note that I prefer to keep UPtr rather than by value executor for now until we have something stable we agree on.

It's fine with me given that we document the ownership transfer.

I'm not sure how this would look in the python bindings.  An instantiated `unique_ptr` can be wrapped into a class, but I'm not sure about dereferencing it for `run` that takes a reference.  Let's ask @prigoyal  and @salexspb 

> In that case what will probably happen is what remains of the execution engine will probably be merged into the autotuner.

How about compilation caches and parsed trees?  Don't we want to reuse them in `compile`?  They can be hidden inside with a `static` variable.

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,147 @@
...
+  /// ExecutionEngine, which also interprets the options as it sees fit.
+  /// \returns opaque handle of a compiled kernel.
+  size_t compile(
+      const std::string& name,
+      const std::vector<const DLConstTensor*>& inputs,
+      const MappingOptionsType& options);
+
+  /// Run a compiled TC kernel given its handle, on the given input tensors and
+  /// fill in the outputs.  All tensors must be allocated and have appropriate
+  /// shapes (inputs same as for copmilation, outputs same as returned by
```
skimo-openhub: "compilation"
Also, it's not entirely clear from the description if the caller needs
to allocate the tensor or if this function needs to do it.

tc/aten/aten_compiler_api_rfc.h
```diff
@@ -0,0 +1,79 @@
...
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/tree.h"
+
+namespace tc {
+/// Run inference on the TC tree resulting from parsing a TC function and
+/// applied to the specified input shapes.
+/// This is used for manual processing of outptu metadata on the ATen user
```
skimo-openhub: "output"

salexspb:
Going through a threading discussing above. Just to clarify, is compiled function thread safe or does it use some global states which can not be used in parallel? If the latter, a workflow might be the following (to avoid recompilation)

1. Parse TC lang
2. Compile, get an Executor object which represents something which is executable but is not thread-safe
3. Have an Executor ::Clone() method which returns one more executor. This is useful only if Clone() is much more efficient than recompilation.
4. User starts several threads, each one gets its own Executor


In some sense an Executor here is close to Caffe2 Net. But first you have to compile it. 


@ftynse , regarding passing executor by unique_ptr and then returning it back, this feels a bit like an overkill to me. I.e. usually not thread safe classes don't really require this tricks to enforce it via an API, thread safety is just documented. Even with this trick one can still create fake unique_ptrs and pass them around (with dummy deleter) :)

tc/aten/aten_compiler_api_rfc.h
```diff
@@ -0,0 +1,79 @@
...
+/// side.
+/// \returns the (contiguous) output tensor shapes as a metadata-owning
+/// DLTensorUPtr.
+std::vector<DLTensorUPtr> inferOutputTensorInfo(
```
salexspb: why do we need to return real tensors for output shapes, why not just a vector<size_t> ?

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,147 @@
...
+ *   engine.define(someTc)
+ *   auto handle = engine.compile(someTc, inputs, mappingOptions)
+ *   auto cpuKernelProfilingInfo = engine.run(handle, inputs, outputs, true);
+ *   // alternatively:
+ *   // auto kernelTiming = engine.uncheckedRun(handle, inputs, outputs, true);
+ */
+template <typename Backend>
+class ExecutionEngine {
+ public:
+  using BackendType = Backend;
```
salexspb: why not call it BackendType in the first place? :)

tc/core/execution_engine_api_rfc.h
salexspb: Why not make this a constructor instead, do we really need an empty one?

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,147 @@
...
+      const std::vector<const DLConstTensor*>& inputs,
+      const MappingOptionsType& options);
+
+  /// Run a compiled TC kernel given its handle, on the given input tensors and
+  /// fill in the outputs.  All tensors must be allocated and have appropriate
+  /// shapes (inputs same as for copmilation, outputs same as returned by
+  /// inferOutputTensorInfo).
+  /// \returns The pair of (cpu,kernel) runtimes. If profile is set, the
+  /// kernel is profiled otherwise its duration is set to Duration::max().
+  ProfilingInfo run(
```
salexspb: I think this was mentioned before. Having run() in the same class as compilation complicates things by introducing states to objects. A cleaner API IMO would have constructors defining a TC and compilation returning something runable. Currently we have
1. empty
2. defined
3. compiled states

which is more than 1 state :) Then when you get an object, you don't know which state it is in.

tc/core/execution_engine_api_rfc.h
```diff
@@ -0,0 +1,147 @@
...
+      const std::string& name,
+      const std::vector<const DLConstTensor*>& inputs,
+      const MappingOptionsType& options);
+
+ private:
+  /// For thread-safety perform all cheap operations under lock.
+  std::mutex tcExecutorMutex_;
+
+  /// Parsed TC trees.
+  std::map<std::string, lang::TreeRef> tcNameMap_;
```
salexspb: This can be another class which stores all of the parsed TCs. Users don't have to use it, IMO, it is just a nice feature for some use-cases. In a basic use-case user can figure out caching theirselves, I think.

nicolasvasilache:
@ftynse @ttheodor @salexspb thanks for your comments.
This PR has lived its purpose and ExecutionEngine is now dead.
I'll send replacement PRs with the outcome of this round.

PR #311, opened 2018-04-16
Make Wno-sign-compare flag non-caffe2 specific

prigoyal:
a user reported the build error on slack channel. With the `Werror` enabled, if Caffe2 is not built i.e. `WITH_CAFFE2=OFF` in the build command, then the `sign-compare` errors happen otherwise they don't because caffe2 build adds the flag `Wno-sign-compare`. 

Rather than having the flag as caffe2 specific, making it non-caffe2 specific.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ message(STATUS "CMAKE_CXX_COMPILER_VERSION is ${CMAKE_CXX_COMPILER_VERSION}")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
 if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "5.4")
   # we officially support these compiler flags for compiler version 5.4 or less
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable -Werror")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable -Werror -Wno-sign-compare")
 endif()
 message(STATUS "CMAKE_CXX_FLAGS is ${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ message(STATUS "CMAKE_CXX_COMPILER_VERSION is ${CMAKE_CXX_COMPILER_VERSION}")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
 if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "5.4")
   # we officially support these compiler flags for compiler version 5.4 or less
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable -Werror")
+  # We suppress sign-compare error messages because caffe2 headers have these errors.
```
ftynse: Why would we make something non-caffe2 specific if only caffe2 has the problem?

(I'm sure TC also has sign-compare warnings, but then the comment does not make sense).

prigoyal: thanks. the comment means that WITH_CAFFE2=ON is needed atm so that sign-compare errors are ignored. this is because the compiler flag is added for it only when caffe2 is build and hence Wno-sign-error is caffe2 build specific flag only.

ftynse: If _only_ caffe2 has these problems, there is no need to make the flag independent of `WITH_CAFFE2`.  If other, non-caffe2 files have these problems, the comment above the line should not say "because caffe2 headers have these errors."

prigoyal: okay, thanks. I understand now what you meant to say. I'll remove the comment (it was moved from caffe2 specific section to here)

ftynse: thanks!

ttheodor:
I think the proper solution would be to mark third-party includes as system headers.

(And in this case fix the sign+unsigned comparisons.)

PR #313, opened 2018-04-16
Proposed changes to the tensor API

nicolasvasilache:
Proposed changes to the tensor API:
1. add a DLConstTensor for proper const correctness propagation + proper overloads and specializations
2. uniformize tensor related functions and put them in the same location
3. make TensorInfo more visible as the tensor representation on which compilation and cache interaction occur
4. cleanup unused functions
5. make this file the only place that includes <dlpack/dlpack.h>

### DIFF
--- a/tc/core/tensor_api_rfc.h
+++ b/tc/core/tensor_api_rfc.h
@@ -1,0 +1,156 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <iostream>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <dlpack/dlpack.h>
+
+#include "tc/proto/compcache.pb.h"
+
+/**
+ * Various tensor utilities used in the Tensor Comprehensions compiler.
+ * At a high-level we use the DLTensor type from dlpack and we add a
+ * DLConstTensor to propagate const-correctness.
+ * DLTensor is used as an ML-framework and backend-agnostic representation to
+ * interact with user-level tensors.
+ * DLTensorUPtr and DLConstTensorUPtr are used as metadata-owning pointers.
+ * The underlying concrete device pointer is borrowed when calling
+ * ExecutionEngine::compile and ExecutionEngine::run.
+ *
+ * Internally, TC uses TensorInfo for compiling and interacting with the
+ * compilation caches. TensorInfo is backed by protobuf and stored directly in
+ * the caches.
+ */
+
+/**
+ * DLPack is missing a ConstTensor. Add it here for now.
+ * In the future, this would be contributed back to dlpack/include/dlpack.h.
+ */
+typedef struct {
+  const void* data;
+  DLContext ctx;
+  int ndim;
+  DLDataType dtype;
+  int64_t* shape;
+  int64_t* strides;
+  uint64_t byte_offset;
+} DLConstTensor;
+
+namespace tc {
+/**
+ * C++ unique_ptr abstraction on top of DLTensor/DLConstTensor
+ */
+struct DLTensorDeleter {
+  template <typename DLTensorType>
+  void operator()(const DLTensorType* t);
+};
+using DLTensorUPtr = std::unique_ptr<DLTensor, DLTensorDeleter>;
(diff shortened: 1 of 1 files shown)


tc/core/tensor_api_rfc.h
```diff
@@ -0,0 +1,166 @@
...
+ * compilation caches. TensorInfo is backed by protobuf and stored directly in
+ * the caches.
+ */
+#ifdef __cplusplus
+extern "C" {
+#endif
+/**
+ * DLPack is missing a ConstTensor. Add it here for now
+ */
+typedef struct {
```
ttheodor: I'm nitpicking here but the `typedef struct {...} name` is something I'd expect to see in a C and not in a C++ codebase.

tc/core/tensor_api_rfc.h
```diff
@@ -0,0 +1,166 @@
...
+      delete[] t->shape;
+    }
+    if (t->strides) {
+      delete[] t->strides;
+    }
+    delete t;
+  }
+};
+typedef std::unique_ptr<DLTensor, DLTensorDeleter> DLTensorUPtr;
+typedef std::unique_ptr<DLConstTensor, DLTensorDeleter> DLConstTensorUPtr;
```
ttheodor: Nitpicking again: I find `using DLTensorUPtr =  std::unique_ptr<DLTensor, DLTensorDeleter>;` more readable

tc/core/tensor_api_rfc.h
```diff
@@ -0,0 +1,166 @@
...
+
+namespace tc {
+/**
+ * C++ unique_ptr abstraction on top of DLTensor/DLConstTensor
+ */
+struct DLTensorDeleter {
+  template <typename DLTensorType>
+  void operator()(const DLTensorType* t) {
+    if (t->shape) {
+      delete[] t->shape;
```
ftynse: Do we have a utility for allocating `DLTensors` with `new[]`?  There's a risk of freeing the memory allocated with `malloc` in C by `delete[]` in C++, which has undefined behavior.

tc/core/tensor_api_rfc.h
```diff
@@ -0,0 +1,166 @@
...
+  explicit TensorInfo(const DLConstTensor* t);
+  explicit TensorInfo(const TensorInfoProto& buf);
+
+  bool operator==(const TensorInfo& t) const;
+  bool operator<(const TensorInfo& t) const;
+  TensorInfoProto toProtobuf() const;
+};
+
+template <typename T>
+std::vector<T> makeStridesFromSizes(const std::vector<T>& sizes);
```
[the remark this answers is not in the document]
ftynse: Tensor libraries use "stride" for a different concept.  Multidimensional tensors are allocated as linear arrays, and stride indicates the distance in terms of array elements between two consecutive indexes of some dimension.  For a C-style array `A[N][M][K]`, if stored consecutively in a row-major format, the strides would be `M*K`, `K`, `1`.

tc/core/tensor_api_rfc.h
nicolasvasilache: yes we do and it is the only way to create DLTensorUPtr for exactly that reason.

tc/core/tensor_api_rfc.h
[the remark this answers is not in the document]
nicolasvasilache: Also notice that arbitrary strides are possible in general tensor libraries which is a can of worms.
In particular people have been abusing stride value 0 to traverse the same data multiple times.

tc/core/tensor_api_rfc.h
nicolasvasilache: This is a C extension that ideally would be contributed back to ```third-party/dlpack/include/dlpack/dlpack.h``` once we kicked its tires a bit.

tc/core/tensor_api_rfc.h
```diff
@@ -0,0 +1,163 @@
...
+ * Internally, TC uses TensorInfo for compiling and interacting with the
+ * compilation caches. TensorInfo is backed by protobuf and stored directly in
+ * the caches.
+ */
+
+/**
+ * DLPack is missing a ConstTensor. Add it here for now.
+ * In the future, this would be contributed back to dlpack/include/dlpack.h.
+ */
+typedef struct {
```
salexspb: if we fork this, why not make a better format from the get go, what is the value in using a few dozen of lines in DLPack? Formats are useful for serialization, but here everything is happening in memory.

Storing int64_t*s for the meta info causes all these custom deleters. If you use vectors instead, it would be easier to use. What do you think?

nicolasvasilache:
> how this would work with the rest of TC without changes there

This API RFC is part of a refactoring, there are changes to the rest of TC

tc/core/tensor_api_rfc.h
nicolasvasilache: > all these custom deleters

do you mean the 10 lines of code for ```struct DLTensorDeleter ```?

We can certainly abandon DLPack for a small pure-C++ abstraction. But the rationale here was that a bunch of external people agreed on using DLPack as the simple abstraction to interoperate and people already implement conversions from/to dlpack that we would have to reimplement for various frameworks. 
This is why we started using DLPack ourselves.

I'd like to keep this consideration separate from the API for now if you don't mind, I'm happy to revisit after the refactoring is done.

nicolasvasilache:
served its purpose

PR #314, opened 2018-04-16
Bump isl for fix of isl_multi_*_val and isl_multi_*_multi_val exports

skimo-openhub:


ftynse:
Ideally, we would add an `operator *` overload

skimo-openhub:
@caffe2bot retest this please

PR #316, opened 2018-04-17
Reduce the number of synchronizations produced.

math-fehr:
This PR aims to reduce the number of synchronizations produced by the compiler :

1. Added functions to be able to insert __syncwarp in the schedule tree and to use it with all versions of CUDA  (added a definition of __syncwarp when the compiler version is below 9).
2. Memorize the dependences in scop instead of recomputing it later. Theses dependences are necessary to know when synchronizations are needed.
3. The function findBestSync compute the lowest synchronization necessary between two schedule trees. The function is used by insertBestSyncInSeq to reduce the number of __syncthreads and __syncwarp inserted in a sequence during thread mapping using dynamic programming. As for now, the number of __syncthreads are minimized first, then the number of __syncwarp.
4. When memory reads/writes are inserted, remove the synchronizations between two reads, or two writes.

### DIFF
--- a/tc/core/libraries.h
+++ b/tc/core/libraries.h
@@ -43,6 +43,13 @@ constexpr auto defines = R"C(
 #define inf __longlong_as_double(0x7ff0000000000000LL)
 )C";
 
+constexpr auto warpSyncFunctions = R"C(
+// Before CUDA 9, syncwarp is a noop since warps are always synchronized.
+#if __CUDACC_VER_MAJOR__ < 9
+__device__ void __syncwarp(unsigned mask = 0xFFFFFFFF) {}
+#endif
+)C";
+
 constexpr auto mathFunctionDecl = R"C(
 
 // BEGIN MATH FUNCTIONS FROM CUDA
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -682,10 +682,14 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
     scop->specializeToContext();
   }
 
-  // 5. Map to threads
+  // 5. Insert mapping context
+  mappedScop->insertMappingContext();
+
+  // 6. Map to threads
   if (outerBand->numChildren() > 0) {
     CHECK_EQ(1u, outerBand->numChildren());
-    // 5.1. Optionally detect reductions while mapping to threads
+    // 6.1. Optionally detect reductions while mapping to threads
+
     if (generic.proto.match_library_calls()) {
       mappedScop->detectReductions(outerBand->child({0}));
     }
@@ -698,13 +702,13 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
         << *mappedScop->schedule();
   }
 
-  // 6. Map to blocks
+  // 7. Map to blocks
   mappedScop->mapToBlocksAndScaleBand(
       outerBand, generic.tiling.extractVector());
   LOG_IF(INFO, FLAGS_debug_tc_mapper) << "After mapping to blocks:" << std::endl
                                       << *mappedScop->schedule();
 
-  // 7. Promote to shared memory below the loops mapped to blocks.
+  // 8. Promote to shared memory below the loops mapped to blocks.
   // This may split the outer band, so find the new outer band after promotion.
   if (cudaOptions.proto().use_shared_memory()) {
     size_t sharedMemorySize = cudaOptions.proto().has_max_shared_memory()
@@ -750,13 +754,11 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
     }
   }
 
-  // 8. Promote to registers below the loops mapped to threads.
+  // 9. Promote to registers below the loops mapped to threads.
   if (cudaOptions.proto().use_private_memory()) {
     promoteToRegistersBelowThreads(mappedScop->scop(), -1ull);
   }
 
-  // 9. Insert mapping context
-  mappedScop->insertMappingContext();
   LOG_IF(INFO, FLAGS_debug_tc_mapper)
       << "After outerBlockInnerThread strategy:" << std::endl
       << *mappedScop->schedule();
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -448,7 +448,357 @@ bool hasOuterSequentialMember(
   }
   return false;
 }
+
+// Intersect the union set with all the mapping
+// filters params in the given schedule tree
+isl::union_set intersectMappingFilterParams(
+    detail::ScheduleTree* st,
+    isl::union_set us) {
+  if (auto filter = st->elemAsBase<detail::ScheduleTreeElemFilter>()) {
+    us = us.intersect(filter->filter_);
+  }
+
+  auto children = st->children();
+  auto nChildren = children.size();
+  if (nChildren == 1) {
+    us = intersectMappingFilterParams(children[0], us);
+  } else if (nChildren > 1) {
+    auto usParent = us;
+    us = intersectMappingFilterParams(children[0], us);
+    for (size_t i = 1; i < nChildren; ++i) {
+      us = us.unite(intersectMappingFilterParams(children[i], usParent));
+    }
+  }
+
+  return us;
+}
+
+// Change the name of the isl ids tied to threads and blocks
+// by adding a suffix
+isl::union_set modifyMappingNames(
+    isl::union_set set,
+    const std::string suffix) {
+  USING_MAPPING_SHORT_NAMES(BX, BY, BZ, TX, TY, TZ);
+  std::unordered_set<isl::id, isl::IslIdIslHash> identifiers{
+      BX, BY, BZ, TX, TY, TZ};
+
+  auto space = set.get_space();
+  for (auto id : identifiers) {
+    auto name = id.get_name();
+    auto dim = space.find_dim_by_name(isl::dim_type::param, id.get_name());
+    CHECK_LE(0, dim);
+    space = space.set_dim_name(isl::dim_type::param, dim, name + suffix);
+  }
+  auto newSet = isl::union_set::empty(space);
+  set.foreach_set([&newSet, &identifiers, &suffix](isl::set set) {
+    for (auto id : identifiers) {
+      auto name = id.get_name();
+      auto dim = set.get_space().find_dim_by_name(isl::dim_type::param, name);
+      CHECK_LE(0, dim);
+      set = set.set_dim_name(isl::dim_type::param, dim, name + suffix);
+    }
+    newSet = newSet.unite(set);
+  });
+  return newSet;
+}
+
+// Get the formula computing the linearized index of a thread in a block.
+isl::aff getLinearizedThreadIdxFormula(
+    isl::space space,
+    const Block& block,
+    const std::string& suffix = "") {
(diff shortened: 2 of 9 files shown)


tc/core/libraries.h
```diff
@@ -43,6 +43,13 @@ constexpr auto defines = R"C(
 #define inf __longlong_as_double(0x7ff0000000000000LL)
 )C";
 
+constexpr auto warpSyncFunctions = R"C(
```
ftynse: Let's have a comment above saying that "before cuda 9, warps are implicitly synchronized so syncwarp is a noop"

tc/core/polyhedral/scop.h
```diff
@@ -229,6 +245,20 @@ struct Scop {
     return true;
   }
 
+  static bool isWarpSyncId(isl::id id) {
+    auto name = id.get_name();
```
ftynse: I would abstract this and `isSyncId` away to a function that takes an extra string argument of expected id prefix.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -649,6 +652,9 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
     scop->specializeToContext();
   }
 
+  // 9. Insert mapping context
```
ftynse: This change is independent of dependence computation and can be moved to a separate commit.  That would provide space to motivate the change.

Also, let's please change the numbering in the comment and use 5 here and 6+ below.

tc/core/polyhedral/scop.cc
```diff
@@ -371,30 +371,19 @@ isl::union_map computeDependences(
   return flow.get_may_dependence();
 }
 
-// Do the simplest possible dependence analysis.
 // Live-range reordering needs tagged access relations to be available.
```
ftynse: This comment is also related to the dependence analysis and should be removed

tc/core/polyhedral/scop.h
```diff
@@ -437,6 +438,13 @@ struct Scop {
       isl::schedule_constraints constraints,
       const SchedulerOptionsView& schedulerOptions);
 
+ public:
+  // Do the simplest possible dependence analysis.
+  // Compute all RAW, WAR, and WAW dependences, and save them in dependences.
+  void computeAllDependences();
+  // Save the RAW, WAR, and WAW dependences
```
ftynse: Saying "save" sounds like a function, but it's a member.  Let's just say "RAW, WAR and WAW dependences"

tc/core/polyhedral/scop.h
```diff
@@ -437,6 +438,13 @@ struct Scop {
...
+  isl::union_map dependences;
```
ftynse: The styling policy is to put members below all functions.  In this case, let's put this below read/write map in line 484.

tc/core/polyhedral/cuda/mapped_scop.h
```diff
@@ -155,6 +155,15 @@ class MappedScop {
   // The remaining parts, if any, are no longer considered for replacement
   // by a library call.
   detail::ScheduleTree* separateReduction(detail::ScheduleTree* band);
+
+  // Find best thread sync between st1 and st2 when st2 is scheduled after
+  // st1.
+  // 0 indicates no sync are needed,
```
ftynse: can we have an `enum` with meaningful names instead?

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+    isl::aff aff(isl::aff::param_on_domain_space(space, id));
+    formula = formula * mappingIds[i].second + aff;
+  }
+
+  return formula;
+}
+
+// Return the constraints ensuring that the points with parameters
+// [t0,t1,t2] and [t0',t1',t2'] are in the same warp.
+// (where t0 is "t0" + suffix1 and t0' is "t0" + suffix2)
```
ftynse: Here and in `getThreadIndexFormula`, it would be helpful to describe the shape of the aff/set in the comment

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+  auto contextSt = scop_->scheduleRoot()->children()[0];
+  auto contextElem = contextSt->elemAs<detail::ScheduleTreeElemContext>();
+  CHECK(nullptr != contextElem);
+  domainAndContext = domainAndContext.intersect_params(contextElem->context_);
+
+  // The domain of both schedule trees filtered by mapping filters,
+  // and then modified to have different threads and blocks names.
+  auto domain1 = intersectMappingFilterParams(st1, domainAndContext);
+  auto domain2 = intersectMappingFilterParams(st2, domainAndContext);
+  domain1 = modifyMappingNames(domain1, "_1");
```
ftynse: Let's extract `"_1"` and `"_2"` into local constants with meaningful names

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+    {TZ, TZ.mappingSize(block)}};
+
+  auto formula1 = getThreadIdxFormula(space, block, suffix1);
+  auto formula2 = getThreadIdxFormula(space, block, suffix2);
+
+  return (isl::aff_set((formula1 / warpSize).floor()) == (formula2 / warpSize).floor());
+  }
 } // namespace
 
+int MappedScop::findBestSync(
```
ftynse: This function assumes it is called above thread mapping (and below block mapping).  I would document this assumption.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
   }
   return false;
 }
+
+// Intersect the union set with all the mapping
+// filters params in the given schedule tree
+isl::union_set intersectMappingFilterParams(
```
ftynse: I don't quite understand what this function returns if there are multiple _divergent_ thread mappings in different children, for example one branch is mapped to (32,4) threads and another to (7,1) threads.  Some reasoning in the comment would be helpful.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+      auto ii = i % nChildren;
+      if(ik == ii) {
+        bestSync[i][k] = 0;
+      } else {
+        bestSync[i][k] = findBestSync(children[ii], children[ik]);
+      }
+    }
+  }
+
+  // Get the worse synchronization needed in the sequence
```
ftynse: "worse" -> "strictest" ?

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+
+  // The optimal number of block sync and thread sync needed to
+  // have the sequence children[i], ..., children[j] correctly
+  // synchronized
+  std::vector<std::vector<std::pair<int, int>>> optimalValue(
+      size, std::vector<std::pair<int, int>>(range, {-1, -1}));
+
+  // An optimal position for doing a synchronization between children[i]
+  // and children[j]. The second member indicates which synchronization
+  // should be made. This should be use recursively to get the optimal
```
ftynse: use -> used

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+  }
+
+  // The optimal number of block sync and thread sync needed to
+  // have the sequence children[i], ..., children[j] correctly
+  // synchronized
+  std::vector<std::vector<std::pair<int, int>>> optimalValue(
+      size, std::vector<std::pair<int, int>>(range, {-1, -1}));
+
+  // An optimal position for doing a synchronization between children[i]
+  // and children[j]. The second member indicates which synchronization
```
ftynse: And what is the first member?

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+  // should be made. This should be use recursively to get the optimal
+  // synchronization. If the second member is equal to 0, this means that
+  // no synchronization is needed.
+  std::vector<std::vector<std::pair<int, int>>> optimalSyncPosition(
+      size, std::vector<std::pair<int, int>>(range, {-1, -1}));
+  for (size_t i = 0; i < size; ++i) {
+    optimalValue[i][0] = {0, 0};
+    optimalSyncPosition[i][0] = {0, 0};
+  }
+  for (size_t k = 1; k < range; ++k) {
```
ftynse: I wonder if it would be possible to extract the DP into a separate function (does not seem to need anything other than the input matrix) and write a test for it comparing against a brute-force implementation on a small size?

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+    if(optimalSyncPosition[i][k].second != 0) {
+      auto separation = optimalSyncPosition[i][k].first;
+      stack.push_back({i, separation - i});
+      stack.push_back({separation + 1, i + k - separation - 1});
+      synchronizations.push_back(optimalSyncPosition[i][k]);
+    }
+  }
+
+  // Insert all the synchronizations
+  std::sort(synchronizations.begin(), synchronizations.end());
```
ftynse: You may sort with `std::greater` to avoid iterating in the inverse order.

ftynse:
typo in bb86ece2754a1a91ed738efc678d9d688a00c7b1 comment "choose" -> "chooses"
typos in 199e26cbe1d223d703b26a3b8723289408b9f836 comment "read" -> "reads", "write" -> "writes", "points" -> "point"

7a771d9794a266435d2f229b23a60a586956ed7d can be squashed into ae75cb3c9051a77ec8bbd8c7da2ae99bf6770b05

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -405,8 +405,292 @@ bool hasOuterSequentialMember(
...
+  if (auto mappingFilter = st->elemAs<detail::ScheduleTreeElemMappingFilter>()) {
+    us = us.intersect(mappingFilter->filter_);
+  } else if (auto filter = st->elemAs<detail::ScheduleTreeElemFilter>()) {
+    us = us.intersect(filter->filter_);
+  }
+
+  auto children = st->children();
+  auto nChildren = children.size();
+  if (nChildren == 1) {
+    us = intersectMappingFilterParams(children[0], us);
```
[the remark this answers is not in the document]
math-fehr: That was a mistake, I shouldn't have taken the union with the superset, but only with the subsets.

tc/core/polyhedral/cuda/mapped_scop.cc
math-fehr: This function add the constraints contained in the mapping filters to the active domain points in the mapping filter node.
If we have a domain { S0[i0] : 0 <= i0 < 42;   S1[i1] : 0 <= i1 < 27 },
followed by a sequence splitting S0 and S1 with filters, and a mapping filter in each branch mapping respectively t0 to i0 on 4 threads and t0 to i1 on 3 threads.
The result would be { S0[i0] : 0 <= i0 < 42 and t0 = i0 mod 4 ;  S1[i1] : 0 <= i1 < 27 and t0 = i1 mod 3 }

PR #318, opened 2018-04-17
Enable -Wpedantic and -Wextra

ttheodor:


test/test_harness.h
```diff
@@ -227,7 +227,7 @@ struct TestHarness {
       caffe2::Workspace& expected,
       caffe2::Workspace& actual,
       std::string name,
-      float relativePrecision = 0.0) {
+      float) {
```
ftynse: rather `0.0f`?

ftynse: if it is unused, drop it completely

ttheodor: it's unused

PR #320, opened 2018-04-18
insertCopiesUnder: use insertExtension{Before,After}

skimo-openhub:
This removes some duplication and simplifies the code.

### DIFF
--- a/tc/core/polyhedral/schedule_transforms.cc
+++ b/tc/core/polyhedral/schedule_transforms.cc
@@ -111,9 +111,6 @@ isl::union_set activeDomainPoints(
   CHECK(domainElem) << "root must be a Domain node" << *root;
 
   auto domain = domainElem->domain_;
-  if (root == node) {
-    return domain;
-  }
 
   for (auto anc : node->ancestors(root)) {
     if (auto filterElem = anc->elemAsBase<ScheduleTreeElemFilter>()) {
--- a/tc/core/polyhedral/schedule_transforms.cc
+++ b/tc/core/polyhedral/schedule_transforms.cc
@@ -588,6 +588,10 @@ void insertExtensionLabelBefore(
   size_t pos;
   auto parent = tree->ancestor(root, 1);
   ScheduleTree* seqTree;
+  if (tree->elemAs<detail::ScheduleTreeElemExtension>()) {
+    tree = tree->child({0});
+    parent = tree;
+  }
   if (tree->elemAs<detail::ScheduleTreeElemSequence>()) {
     seqTree = tree;
     pos = 0;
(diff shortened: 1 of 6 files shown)


tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -115,7 +116,11 @@ isl::union_set activeDomainPoints(
     return domain;
   }
 
-  for (auto anc : node->ancestors(root)) {
+  auto ancestors = node->ancestors(root);
+  if (inclusive) {
```
nicolasvasilache: This looks very fishy to me, see [here](https://github.com/facebookresearch/TensorComprehensions/blob/67b50561d913a526701388e8b42b045a2a9545d7/tc/core/polyhedral/schedule_tree.cc#L100), ancestors includes node (and actually checks for it as an invariant)

tc/core/polyhedral/schedule_transforms.h
```diff
@@ -284,11 +284,21 @@ isl::multi_union_pw_aff partialScheduleMupa(
...
-// Get the set of domain points active at the given node.  A domain
+// Get the set of domain points active at the given node, or below
+// the given node if "inclusive" is set.  A domain
 // point is active if it was not filtered away on the path from the
 // root to the node.  The root must be a domain element, otherwise no
 // elements would be considered active.
 isl::union_set activeDomainPoints(
+    const detail::ScheduleTree* root,
+    const detail::ScheduleTree* node,
+    bool inclusive = false);
```
nicolasvasilache: I am definitely not a fan of adding such boolean parameters to modify behavior.
Why not have activeDomainPoints inclusive (as it should already be because of my note above about ancestors). If you want exclusive behavior you can just do:
```
activeDomainPoints(root, node->ancestor(root, 1));
```
instead of the API naming bloat.

tc/core/polyhedral/schedule_transforms.h
```diff
@@ -284,11 +284,21 @@ isl::multi_union_pw_aff partialScheduleMupa(
...
+
+// Get the set of domain points active below the given node.  A domain
+// point is active if it was not filtered away on the path from the
+// root to the node.  The root must be a domain element, otherwise no
+// elements would be considered active.
+isl::union_set activeDomainPointsBelow(
```
nicolasvasilache: This API bloat seems unjustified to me

tc/core/polyhedral/schedule_transforms.h
skimo-openhub: I can rename `activeDomainPoints` to `activeDomainPointsAbove` and implement it in terms of `activeDomainPointsBelow`.  Changing `activeDomainPoints` to mean `activeDomainPointsBelow` does not sound like a good idea at all.  I'm already horribly confused by how the promotion performs the promotion underneath the node it is pointing at.

tc/core/polyhedral/schedule_transforms.cc
skimo-openhub: See https://github.com/facebookresearch/TensorComprehensions/blob/67b50561d913a526701388e8b42b045a2a9545d7/tc/core/polyhedral/schedule_tree.cc#L104

nicolasvasilache: my bad misread the impl, sorry about that

tc/core/polyhedral/schedule_transforms.h
ftynse: We also have this in #316

nicolasvasilache:
ok let's go with this following our sync and let's discuss IR separately

PR #322, opened 2018-04-18
[RFC] High-level compilation API

nicolasvasilache:
Following the comments to #309, here is a more functional-style API for compilation that should be closer to the consensus.
In particular, compile<Backend> returns a unique_ptr<Executor> on which ```run``` can be called.
The ExecutionEngine class disappears which has nice implications on the autotuner (future RFC).

### DIFF
--- a/tc/core/compiler_api_rfc.h
+++ b/tc/core/compiler_api_rfc.h
@@ -1,0 +1,127 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <map>
+#include <string>
+#include <vector>
+
+#include "tc/core/mapping_options.h"
+#include "tc/core/tensor.h"
+#include "tc/lang/tree.h"
+
+/**
+ * This provides a simple functional-style C++ API with multi-backend
+ * capabilities to:
+ *   1. compile a TC function and return an Executor for the specified Backend
+ *      on which the run method can be called;
+ *   2. infer actual tmp/output tensor shapes given input tensor shapes;
+ *   3. parse a TC definition and retrieve the map of TC function to parsed TC
+ *      trees.
+ *
+ * Compilation is backed by a compilation cache, its correspondance is:
+ * 1 TcExecutor <-> 1 compiled tuple<TC function, input shapes, MappingOptions>
+ *
+ * The compile function is templated by the Backend type.
+ * For each backend, the specific Backend type lives in
+ *   backendname/backendname_backend.h and declares all the required dependent
+ *   **derived** types.
+ * For example:
+ *   CudaBackend is declared in core/cuda/cuda_backend.h
+ *
+ * struct CudaBackend {
+ *   using ExecutorType = CudaTcExecutor;
+ *   using MappingOptionsType = CudaMappingOptions;
+ *   using ProtobufType = CudaCacheProto;
+ *   using CacheType = CudaCache;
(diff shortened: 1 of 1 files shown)


tc/core/compiler_api_rfc.h
```diff
@@ -0,0 +1,113 @@
...
+/// compiles a new TcExecutor for the specified Backend.
+/// For now, contiguous output sizes are inferred given input sizes.
+/// If you need another kernel for another TC or another inputs, options then
+/// just compile another TcExecutor; because atm we fully JIT specialize on all
+/// sizes.
+/// \returns a new TcExecutor on which the run method can be called
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    const std::string& tc,
+    const std::string& tcFunctionName,
```
ftynse: Assuming `tc` can contain multiple functions, I'd call this `entryPoint` or `mainFunction`.  If `tc` can only contain one function (because we cannot call other functions atm), this argument is unnecessary.

tc/core/compiler_api_rfc.h
```diff
@@ -0,0 +1,113 @@
...
+namespace tc {
+/// Given a TC representation as a TC + TC function name, this function
+/// compiles a new TcExecutor for the specified Backend.
+/// For now, contiguous output sizes are inferred given input sizes.
+/// If you need another kernel for another TC or another inputs, options then
+/// just compile another TcExecutor; because atm we fully JIT specialize on all
+/// sizes.
+/// \returns a new TcExecutor on which the run method can be called
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
```
ftynse: What's the story here for output tensor info?  Can we query it from the executor as well?

nicolasvasilache: It is present in the executor for runtime check purposes but I have not (yet) exposed to the outside world, maybe we should and then ```inferOutputTensorInfo``` would be less used from the top level (i.e. I imagine people may want to just compile and query the required sizes afterwards, not 100% sure atm).

Executor RFC coming soon, cleaning stuff up.

ftynse: If `inferOutputTensorInfo` is supposed to be called in a regular flow from the outside, I would not put it into `detail::`

nicolasvasilache: The rationale is to put the tree representation in detail.
I could also add
```
std::vector<TensorInfo> inferOutputTensorInfo(
  const std::string& tc,
  const std::string& entryPoint,
  const std::vector<const DLConstTensor*> inputs);
```
in 
```namespace tc```

High-level functionality: operate on strings, parse on the fly, memoizing parsing is premature optimization.
Lower-level functionality: for some reason you care about parsing time (I would claim you shouldn't), use the ```detail```-level API and operate on ```lang::TreeRef``` directly.

Comment (nicolasvasilache) [reply tc/core/compiler_api_rfc.h]: added the string version below. makes sense in the context of #324

File: tc/core/compiler_api_rfc.h [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+std::vector<TensorInfo> inferOutputTensorInfo(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<const DLConstTensor*> inputs);
+
+namespace detail {
+/// Given a TC representation as a TreeRef, this function compiles a new
+/// TcExecutor for the specified Backend.
+/// For now, contiguous output sizes are inferred given input sizes.
+/// If you need another kernel for another TC or another inputs, options then
```
skimo-openhub: Nit: something is wrong in "another inputs, options".
Did you mean "another pair of inputs and options"?
Same phrase also occurs below.

tc/core/compiler_api_rfc.h
```diff
@@ -0,0 +1,127 @@
...
+std::vector<TensorInfo> inferOutputTensorInfo(
+    lang::TreeRef tcDefinition,
+    const std::vector<const DLConstTensor*> inputs);
+
+/// Given a TC representation, this parses the TC functions into a map of
+/// TreeRef indexed by TC function names.
+/// \returns an ordered map of TC function name to parsed TC tree
+std::map<std::string, lang::TreeRef> parse(const std::string& tc);
+} // namespace detail
+} // namespace tc
```
skimo-openhub: Nit: put `parse` declaration before `compile` declaration since the latter
requires the output of the former.

ftynse:
signed off /@ftynse

nicolasvasilache:
served its purpose, implemented in #359

PR #323, opened 2018-04-18
[RFC] API for TcExecutor

nicolasvasilache:
TcExecutor is a backend-agnostic abstraction that is the result of compilation (see #322).
When a concrete executor specializes TcExecutor<Backend> it provides support for running a compiled TC on the particular backend.

Specializing a TcExecutor<Backend> consists in overriding 2 methods:
1. uncheckedRun which provides the low-latency path on void* data
2. setupRuntimeCompiledFunction which takes the results of compileWithTcMapper to:
   a. save backend-specific information required at runtime (e.g. grid and block sizes for the CUDA backend)
   b. create the in-memory object resulting from JIT-compilation that implements the TC

### DIFF
--- a/tc/core/tc_executor_api_rfc.h
+++ b/tc/core/tc_executor_api_rfc.h
@@ -1,0 +1,145 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "tc/core/tc2halide.h"
+#include "tc/core/tensor.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/tree.h"
+
+namespace tc {
+/**
+ * TcExecutor is a backend-agnostic abstraction that is the result of
+ * compilation. When a concrete executor specializes TcExecutor<Backend> it
+ * provides support for running a compiled TC on the particular backend.
+ *
+ * TcExecutor is templated by the Backend type.
+ * For each backend, the specific Backend type lives in
+ *   core/backend/backend_tc_executor.h and declares all the required dependent
+ *   **concrete** types.
+ * For example:
+ *   CudaBackend is declared in core/cuda/cuda_tc_executor.h
+ *
+ * struct CudaBackend {
+ *   using ExecutorType = CudaTcExecutor;
+ *   using MappingOptionsType = CudaMappingOptions;
+ *   using ProtobufType = CudaCacheProto;
+ *   using CacheType = CudaCache;
+ *   using CacheEntryType = CudaCacheEntry;
+ *   using CacheKeyType = typename CacheEntryType::Key;
+ *   using CompilationResultType = typename CacheEntryType::Values;
+ *   using RTCFunctionType = CudaRTCFunction;
+ * };
+ *
+ * The correspondence is 1 TcExecutor for 1
+ *   tuple<TC function, input shapes, Backend::MappingOptions>.
+ *
+ * Backend-specific Executors should specialize TcExecutor<Backend> and
+ * implement the uncheckedRun and setRuntimeCompiledFunction.
+ *
+ * Specializing a TcExecutor<Backend> consists in overriding 2 types of methods:
+ * 1. uncheckedRun/profileUnchecked which provides the low-latency path on
+ *    void* data
+ * 2. setupRuntimeCompiledFunction/clearRuntimeCompiledFunction which takes
+ *    the results of compileWithTcMapper to:
+ *       a. save backend-specific information required at runtime (e.g. grid and
+ *          block sizes for the CUDA backend)
+ *       b. create the in-memory object resulting from JIT-compilation that
+ *          implements the TC
+ */
+template <typename Backend>
+class TcExecutor {
+ protected:
+  TcExecutor(
+      lang::TreeRef tcDefinition,
+      const std::vector<const DLConstTensor*>& inputs);
+
+ public:
+  virtual ~TcExecutor();
+
+  TcExecutor(TcExecutor&&) = delete;
+  TcExecutor& operator=(TcExecutor&&) = delete;
+  TcExecutor(const TcExecutor&) = delete;
+  TcExecutor& operator=(const TcExecutor&) = delete;
+
+  /// Run can be called multiple times given a compilation, inputs are allowed
+  /// to change in that their data pointer is allowed to change.
+  /// Sizes and strides must remain constant otherwise this is an error.
+  /// The only thing that is allowed to change across runs is the input
+  /// and output pointers base address.
+  /// It is the caller's responsibility to ensure proper non-aliasing (or
+  /// advanced aliasing) properties of the input and output tensors.
+  void run(
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs) const;
+
+  /// Calls run and profiles the kernel runtime (microseconds).
+  /// \returns profiling information (see: tc/core/utils/time.h)
+  ProfilingInfo profile(
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs) const;
+
+  /*****************************************************************************
+   * The following pure virtual methods are the minimal set of required
+   * backend-dependent functions.
+   ****************************************************************************/
+  /// This is the "low-latency" mode in which we just propagate raw pointers to
+  /// data in the address space where kernel is executed.
+  /// No tensor-related information can be checked so it is the user's
+  /// responsibility to ensure that shapes and strides match. If the user
+  /// doesn't then segfault will likely occur.
+  virtual void uncheckedRun(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const = 0;
+  virtual ProfilingInfo profileUnchecked(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const = 0;
+
+ protected:
+  /// setupRuntimeCompiledFunction takes a compilationResult (returned by
+  /// the specialized TC mapper) and:
+  ///   a. saves backend-specific information required at runtime
+  ///     (e.g. grid and block sizes for the CUDA backend)
+  ///   b. creates the in-memory object resulting from JIT-compilation that
+  ///      implements the TC
+  virtual void setupRuntimeCompiledFunction(
+      const typename Backend::CompilationResultType& compilationResult) = 0;
+
+  /// It may be necessary to clear the RTC manually because it can throw and
+  /// we can't have that in the RTC destructor.
+  virtual void clearRuntimeCompiledFunction() = 0;
+
+ protected:
+  /// Used to check proper metadata when calling run
+  std::vector<TensorInfo> inputsInfo_;
+  std::vector<TensorInfo> outputsInfo_;
+  tc2halide::HalideComponents halideComponents_;
+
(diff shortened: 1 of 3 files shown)


tc/core/tc_executor_api_rfc.h
```diff
@@ -0,0 +1,138 @@
...
+  /// The only thing that is allowed to change across runs is the input
+  /// and output pointers base address.
+  /// It is the caller's responsibility to ensure proper non-aliasing (or
+  /// advanced aliasing) properties of the input and output tensors.
+  /// if profile is set the kernel runtime (microseconds) is returned.
+  /// \returns profiling information (see: tc/core/utils/time.h)
+  ProfilingInfo run(
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs,
+      bool profileKernel = false) const;
```
ftynse: how about an actual overload:
```c
void run (inputs, outputs) const;
ProfilingInfo run (inputs, outputs, profile) const;
```
?
The third argument can be a dummy `struct` with a good name.  Or even have two separate calls: `run` and `profile`.

tc/core/tc_executor_api_rfc.h
```diff
@@ -0,0 +1,138 @@
...
+ * run on a specified backend. The correspondence is 1 TcExecutor for 1
+ *   tuple<TC function, input shapes, MappingOptions>.
+ *
+ * TcExecutor is an abstract base class which provides capabilities to:
+ *   1. compile a TC function into a CompilationResult and an executable
+ *      RTCFunction;
+ *   2. store an abstraction of ExecutionContext and RTCFunction to check and
+ *      run the result of step 1. multiple times on different tensors of
+ *      compliant sizes;
+ *   4. interact with the compilation CacheType to allow saving and loading
```
ftynse: did you mean `3.`?

tc/core/tc_executor_api_rfc.h
```diff
@@ -0,0 +1,138 @@
...
+      const typename Backend::CompilationResultType& compilationResult) = 0;
+
+ protected:
+  /// Used to check proper metadata when calling run
+  std::vector<TensorInfo> inputsInfo_;
+  std::vector<TensorInfo> outputsInfo_;
+  tc2halide::HalideComponents halideComponents_;
+
+  /// The following are initialized as a result of compilation with the TcMapper
+  std::vector<int> parameters_;
```
ftynse: we moved to using `unordered_map<string, int>` elsewhere for parameters, maybe do this here too?

tc/core/tc_executor_api_rfc.h
```diff
@@ -0,0 +1,138 @@
...
+  /// if profile is set the kernel runtime (microseconds) is returned.
+  /// \returns profiling information (see: tc/core/utils/time.h)
+  ProfilingInfo run(
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs,
+      bool profileKernel = false) const;
+
+  /// It is necessary to clear the RTC manually because it can throw and we
+  /// can't have that in the destructor.
+  inline void clearRuntimeCompiledFunction();
```
ftynse: If `setupRuntimeCompiledFunction` is virtual, so should be `clearRuntimeCompiledFunction`.  It may be concrete and empty though.

tc/core/tc_executor_api_rfc.h
nicolasvasilache: ugh comments are stale, updating

tc/core/tc_executor_api_rfc.h
nicolasvasilache: Position is important (passed in order to rtc function call) and it is not necessarily ordered by name (so it can't be just a map).  And I'd claim we don't want to be in the business of reordering at runtime.

tc/core/tc_executor_api_rfc.h
nicolasvasilache: ```run```/```profile``` sounds nice to me

tc/core/tc_executor_api_rfc.h
```diff
@@ -0,0 +1,145 @@
...
+
+ protected:
+  /// Used to check proper metadata when calling run
+  std::vector<TensorInfo> inputsInfo_;
+  std::vector<TensorInfo> outputsInfo_;
+  tc2halide::HalideComponents halideComponents_;
+
+  /// The following are initialized as a result of compilation with the TcMapper
+  std::vector<int> parameters_;
+  std::shared_ptr<typename Backend::RTCFunctionType> rtcFun_;
```
ttheodor: Why a shared `rtcFun`? Doesn't each executor own it's own?

tc/core/tc_executor_api_rfc.h
```diff
@@ -0,0 +1,145 @@
...
+  /*****************************************************************************
+   * The following pure virtual methods are the minimal set of required
+   * backend-dependent functions.
+   ****************************************************************************/
+  /// This is the "low-latency" mode in which we just propagate raw pointers to
+  /// data in the address space where kernel is executed.
+  /// No tensor-related information can be checked so it is the user's
+  /// responsibility to ensure that shapes and strides match. If the user
+  /// doesn't then segfault will likely occur.
+  virtual void uncheckedRun(
```
ttheodor: Why the virtual interface? Are specialized `TcExecutor`s (e.g., `TcExecutor<CudaBackend>`) supposed to be inherited from?

ftynse: I had had the same question and came up with the following explanation (but it would be much better if it was documented).
We need inheritance to share certain functionality between executors (non-virtual functions, parameters, etc.).  But we want individual executors to be statically type-safe, e.g. take only the mapping options of the specified type, which would not be possible without templates. So we have _both_ templates and inheritance: `class DefaultCudaExecutor : public TcExecutor<Cuda>`.  As a side effect, it provides a possibility for _multiple_ concrete executors for the same backend.  For example, `class DLTensorCudaExecutor : public TcExecutor<Cuda>` and `class ATenCudaExecutor : public TcExecutor<Cuda>`.

 Another solution here is to use CRTP for static polymorphism, but it would require maintaining a list of inheriting classes and manually dispatching polymorphic functions, which is arguably annoying.

ttheodor: I don't think we need virtuals or CRTP. E.g. I'd imagine that `run` does something like:
```
void run(const std::vector<const DLConstTensor*>& inputs,
              const std::vector<const DLTensor*>& outputs) const {
  //check sizes
  uncheckedRun(inputs, outputs)
}
```
`run` is shared among executors and each concrete has to implement `unchechedRun`.
If an executor omits it, then it won't compile

ttheodor: Unless we want to use subtypes of `TcExecutors<ConcreteBackend>` polymorphically then there's no need for the virtuals.

ftynse: How would it call `uncheckedRun` if it is not declared (even as pure virtual)?  If it is declared but not virtual, `run` will call `uncheckedRun` from the base class.

ttheodor: One way to do it:
```
template <typename T> struct A {
    template <typename U> struct URunner { static void uncheckedRun(); };
    void run() { URunner<T>::uncheckedRun(); }
};

template <> template <> struct A<int>::URunner<int> {

    static void uncheckedRun() {}
};

int main() {
    A<int> a;
    a.run();
    A<char> b;
    b.run(); //causes linker error, probably with some TMP magic it can be turned to a compile error
}
```

ttheodor: `static_assert` doesn't work directly in this example, it will trigger regardless of any specializations' existence. 

I don't have  problem with virtual member function but:
A class with a virtual API declares certain invariants/intents one of which would be that it is meant to be used polymorphically. However, according to the documentation we are supposed to specialize the `TcExecutor` and not inherit from it, on the other hand a virtual destructor indicates that we are actually expecting to handle subtypes of `TcExecutor<XBacked>` through pointers to `TcExecutor<XBacked>`.  I find this confusing.

But the use cases you mentioned, e.g. having an `ATen` based executor,  are not handled well with by virtual functions since would have to write wrappers with `at::Tensor` parameters.

ttheodor: `Too much template magic for my taste, and having to run double-template specializations is not much better than maintaining CRTP dispatch.`

`uncheckedRun` and `profileUnchecked` could be factored out to a separate struct that is passed as a template argument, e.g.:

```
template <typename Backend, template <typename> typename TcExecutorUnchecked >
class TcExecutor {
//...
void run(const std::vector<const DLConstTensor*>& inputs,
              const std::vector<const DLTensor*>& outputs) const {
  //check sizes
  TcExecutorUnchecked<Backend>::uncheckedRun(*this, inputs, outputs);
}
//...
};
```
And the user of `TcExecutor` has to just create a type `TcExecutorUnchecked` that implements the two functions (or we provide it and it has to be explicitly specialized).

ftynse: I already commented that `Backend` struct seems artificial.  Couldn't we just have a `CudaTcExecutor` (and `ATenCudaTcExecutor`) with all relevant defines.

> `TcExecutorUnchecked<Backend>::uncheckedRun`

This and previous examples assume `uncheckedRun` is static.  So concrete executors cannot have state.  This looks problematic because they usually need state, like grid configuration.

ftynse: > doesn't work directly in this example,

then just don't declare the function in the non-specialized case

>  However, according to the documentation

This is the root of the problem. *The documentation has not been updated* so I stopped reading it on these PRs.

> A class with a virtual API declares certain invariants/intents one of which would be that it is meant to be used polymorphically.

Dynamic dispatch to specific implementations from a shared generic one seems a legit polymorphism usecase for me.

ttheodor: > This and previous examples assume `uncheckedRun` is static. So concrete executors cannot have state. This looks problematic because they usually need state, like grid configuration.

In the example I'm passing the `*this` pointer to `uncheckedRun` for this exact reason.

Also I think there's a terminology mismatch which sent me down the template rabbit hole:
the documentation mentions specialization but it should have been (in C++ lingo) instantiation. 
Specializing the whole template class makes little sense since we'd be throwing away all default and
shared implementations.

ftynse: > specialization but it should have been (in C++ lingo) instantiation.
> Specializing the whole template class makes little sense since we'd be throwing away all default and
shared implementations.

Indeed.

ftynse:
I would consider this @nicolasvasilache @ttheodor 

```c++
template <typename Concrete>
class AbstractExecutor {
 public:
  Duration run(...) {  // implementation is shared by concrete executors
    checkShapes(...);
    return static_cast<Concrete>(this)->uncheckedRun(...);
  }

 public:
  // data shared by concrete executors
};

class ConcreteExecutorMappingOptions;
class ConcreteRTCFunction;

class ConcreteExecutor : public AbstractExecutor<ConcreteExecutor> {
 public:
  // there is no Backend class anymore
  using MappingOptions = ConcreteExecutorMappingOptions;
  using ExecutableObject = ConcreteRTCFunction;

  static std::unique_ptr<ConcreteExecutor> compile(..., const MappingOptions&);

   // target-specific execution
  Duration uncheckedRun(...);

 public:
  // data specific to this concrete executor
};

template <typename Concrete>
std::unique_ptr<Concrete> compile(..., typename Concrete::MappingOptions& options) {
  return Concrete::compile(..., options);
}
```

ftynse:
or this

```c++
template <typename Backend>
class TcExecutor {
 public:
  Duration run(...) {  // implementation is shared by concrete executors
    checkShapes(...);
    static_cast<Backend::ExecutorType*>(this)->uncheckedRun();
  }

 public:
  // data shared by concrete executors
};


class CudaExecutor : public TcExecutor<CudaBackend> {
 public:
  Duration uncheckedRun();
};
```

nicolasvasilache:
Since people may get confused by mixing CRTP and virtual functions I am fine using the type of static dispatch @ftynse propose in order to remove virtual functions:
```
static_cast<Backend::ExecutorType*>(this)->uncheckedRun();
```
The tradeoff is that one will get error at the call sites with error messages.
I am personally more favorable to virtual functions (original proposal) because error messages will pop up at the definition site in the abstract base class but I can also see that mixing the two makes it more difficult to understand what happens.
I will send a (non-squashed), update so we can keep track of the differences.

nicolasvasilache:
served its purpose

PR #324, opened 2018-04-19
[RFC] ATen API

nicolasvasilache:
This provides a minimal API to wrap tc/core/compiler.h under a tensor library like ATen.
In particular this gets rid of the ATenCompilationUnit class which kept state about parsed TCs.
The proposal is that the executors themselves should be memoized and reused.
Tinkering with the underlying TreeRef memoization is penny-wise and pound foolish: parsing overhead never matters and it adds cognitive overhead to the API.

### DIFF
--- a/tc/aten/aten_compiler_api_rfc.h
+++ b/tc/aten/aten_compiler_api_rfc.h
@@ -1,0 +1,86 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


tc/aten/aten_compiler_api_rfc.h
```diff
@@ -0,0 +1,86 @@
...
+namespace tc {
+/// Given a TC string with multiple functions defined and a TC function name,
+/// this run inference, applied to the specified input shapes.
+/// This is used for manual processing of output metadata on the ATen user
+/// side.
+/// \returns the (contiguous) output tensor shapes as a metadata-owning
+/// DLTensorUPtr.
+std::vector<DLTensorUPtr> inferOutputTensorInfo(
+    const std::string& tc,
+    const std::string& tcFunctionName,
```
skimo-openhub: Shouldn't these (3x) be `entryPoint` as well?

tc/aten/aten_compiler_api_rfc.h
```diff
@@ -0,0 +1,86 @@
...
+#pragma once
+
+#include <chrono>
+#include <string>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/core/utils/time.h"
+
+namespace tc {
```
ftynse: do we want these functions in `tc` namespace or in some `tc::aten` namespace?

nicolasvasilache:
served its purpose

Issue #190, opened 2018-03-21
Warning for uninitialized reductions


ftynse:
Consider the following layer with a reduction:
```
def maxpool3(float(B, C, H, W) input) -> (output) {
  output(b, c, h, w) max= input(b, c, h * 3 + kh, w * 3 + kw) where kh in 0:3, kw in 0:3
}
```

It uses a reduction **without initialization**, which _reads_ from `output` before writing to it.  Such behavior is essentially undefined because we have no control over what is stored in the output tensor given to us by the upstream framework.  Such behavior is dangerous and should be avoided.  We need at least a warning, or even an compilation error.

It is necessary to use `max=!` here.

I see how to extract this information from polyhedral dependence analysis (there are live-in accesses on output tensors), but it may be preferable to do at an earlier stage because we currently don't emit compilation warnings/errors from polyhedral analyses.

Note that #187 fixes some of such problems, and there were early commits dealing with the same effects.

ftynse:
Closed by #325

PR #325, opened 2018-04-19
Warning for uninitialized reductions.

dingobye:
This PR adds warnings for uninitialized reductions, which is described in issue #190 .

Such checking happens in sema stage. When checking a stmt, if it is a reduction, we raise a warning if the following two conditions hold: (1) the reduction operator is not suffixed with !, and (2) the lhs has not been previously defined.

Some test cases are also included.

### DIFF
--- a/tc/lang/sema.h
+++ b/tc/lang/sema.h
@@ -497,10 +497,9 @@ struct Sema {
         nullptr == lookup(stmt.ident(), false)) {
       ErrorReport err(stmt);
       std::string tk = kindToToken(stmt.assignment()->kind());
-      err << "Reduction without initialization. "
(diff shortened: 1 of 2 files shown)


dingobye:
Hi Oleksandr, 

Thanks for the review. As per your feedback, I have rephrased the warning content as below, in case we still want to have such warning:

_WARNING:
Reduction without initialization. If `output` is not pre-initialized before calling the TC function, consider using the !-suffixed reduction operator `op!` instead of `op`:_

skimo-openhub:
@caffe2bot retest this please

dingobye:
rebased

ftynse:
thanks!

PR #326, opened 2018-04-19
TC: Compilation fix for -Werror=shadow-compatible-local

salexspb:
I am hitting these errors internally. But this option doesn't seem to be available in default gcc. Or I might be doing something stupid. What I tried - I added -Werror=shadow-compatible-local to the cmake argument link along with -Werror. Also google shows not a lot of references of this option, first one is from Jim, which makes me thing it is not common. 

So I  figured out I am going to just fix the code for now.

### DIFF
--- a/tc/autotuner/genetic_search.cc
+++ b/tc/autotuner/genetic_search.cc
@@ -230,7 +230,7 @@ TuningConfiguration GeneticSearch::crossover(
     }
   };
 
-  for (size_t i = 0; i < kMutateIterations; ++i) {
+  for (size_t iter = 0; iter < kMutateIterations; ++iter) {
     TuningConfiguration child{a};
     auto params = child.collectParameters();
     for (size_t i = 0; i < params.size(); ++i) {
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -402,12 +402,12 @@ isl::schedule makeScheduleTreeHelper(
     // this loop var at the same index.
     isl::multi_union_pw_aff mupa;
     body.get_domain().foreach_set([&](isl::set s) {
-      isl::aff loopVar(
+      isl::aff newLoopVar(
           isl::local_space(s.get_space()), isl::dim_type::set, thisLoopIdx);
       if (mupa) {
-        mupa = mupa.union_add(isl::union_pw_aff(isl::pw_aff(loopVar)));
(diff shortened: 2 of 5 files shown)


tc/autotuner/genetic_search.cc
```diff
@@ -230,7 +230,7 @@ TuningConfiguration GeneticSearch::crossover(
     }
   };
 
-  for (size_t i = 0; i < kMutateIterations; ++i) {
```
ftynse: hmm, this was a bug!

tc/core/halide2isl.cc
```diff
@@ -402,12 +402,12 @@ isl::schedule makeScheduleTreeHelper(
     // this loop var at the same index.
     isl::multi_union_pw_aff mupa;
     body.get_domain().foreach_set([&](isl::set s) {
-      isl::aff loopVar(
+      isl::aff newLoopVar(
```
ftynse: I wonder if it is not better solved by only capturing what we need in the lambda.

skimo-openhub:
On Thu, Apr 19, 2018 at 07:58:59AM +0000, ftynse wrote:
> > @@ -230,7 +230,7 @@ TuningConfiguration GeneticSearch::crossover(
>      }
>    };
>  
> -  for (size_t i = 0; i < kMutateIterations; ++i) {
> 
> hmm, this was a bug!

How so?

skimo

tc/autotuner/genetic_search.cc
ttheodor: No, the code is actually correct. The outer induction variable is never used inside the body of the loop.

tc/core/halide2isl.cc
salexspb: @ftynse , this won't help, I still get a compilation error if it shadows a name

ftynse: That's a badly implemented warning then :(

Can we at least have something like `loopVariable` instead of `newLoopVar` here?  Otherwise it's misleading in loop creation context....

tc/core/tc2halide.cc
```diff
@@ -724,11 +724,11 @@ void translateComprehension(
 
   if (added_implicit_initialization) {
     // Also reorder reduction initializations to the TC convention
-    vector<Var> lhs = func.args();
+    vector<Var> lhs2 = func.args();
```
ftynse: Not sure why this was called `lhs` to start with...  Can we call it `funcArgs` instead?

Issue #307, opened 2018-04-13
[RFC] C++ API refactor towards true backend-agnostic TC: master task


nicolasvasilache:
Now that the .cc and .h have been merged in the same repositories, I started a bigger effort to refactor and break the cuda dependency end-to-end. Let's use this task to keep track of the proposed APIs.
I'm adding a larger audience in case people are interested to chime in.

The list of issues is:
- [x] Tensor API #313 (impl in #361)
- [x] High-level compilation API #322 (impl in #359)
- [x] TcExecutor API #323 (impl in #363)
- [x] ATen compiler API #324 (impl in #362)
- [x] Options cache API #339 (impl in #365)
- [x] Multi-backend Autotuner API #342 (impl in #366)
- [x] Multi-backend, multi-search user-facing Autotuner API #345 (impl in #367)
- [x] Drop manual and compilation caches (impl in #353)
- [x] Bring it home #379 

Obsolete:
- [x] ~ExecutionEngine API #309, consensus seems to be towards a more functional-style API, as a consequence ExecutionEngine has died~

nicolasvasilache:
cc @mingzhe09088 @mitmul

PR #327, opened 2018-04-19
Remove C2 overhead

nicolasvasilache:
This PR introduces memoization for compilation and data pointers in tc_op.h
This is temporary since the bigger API refactoring will change things here but at least it will allow @salexpb to run benchmarks without ridiculous CPU overhead.

### DIFF
--- a/tc/c2/tc_op.h
+++ b/tc/c2/tc_op.h
@@ -49,7 +49,7 @@ class TcOp : public Operator<Context> {
         OperatorBase::GetSingleArgument<std::string>("tcGradDef", "ERROR");
     gradTcName_ =
         OperatorBase::GetSingleArgument<std::string>("tcGradName", "ERROR");
-    profile_ = OperatorBase::GetSingleArgument<bool>("profile", false);
+    checkSizes_ = OperatorBase::GetSingleArgument<bool>("checkSizes", false);
     compiled_ = false;
     handle_ = 0;
     ArgumentHelper args(operator_def);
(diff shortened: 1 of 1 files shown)


tc/c2/tc_op.h
```diff
@@ -95,38 +97,38 @@ class TcOp : public Operator<Context> {
...
-    // now, given the input tensors, convert them to dlpack tensors so that
-    // we can call the compile command
-    std::vector<::tc::dlutils::DLTensorUPtr> inTensorUPtrs;
-    std::vector<const DLTensor*> inputDLTensors;
-    for (int idx = 0; idx < this->InputSize(); ++idx) {
-      auto dims = this->Input(idx).dims();
-      inTensorUPtrs.emplace_back(
-          dlpack::makeConstDLTensor(this->Input(idx), dims));
-      inputDLTensors.push_back(inTensorUPtrs.back().get());
+    if (!compiled_) {
```
salexspb: What if dimensions changed? You might need to store a cache of compiled kernels and compare dimensions. Or have a flag that tells TcOp that dimensions are fixed. Do you think we can do dimensions checks fast enough so we can actually do them all the time?

nicolasvasilache: This check is provided in ```executionEngine_->run``` which should be run under a ```checkSizes_``` flag instead of ```profile_``` flag. Without it you ```uncheckedRun``` and call the ```void*``` version.
I'd say it is too expensive to run all the time for low latency kernels but YMMV, measure it ?

@abadams was saying that we don't need 2 modes and that Halide + LLVM can compile a check function for you.
I think this would have been tricky to implement with the caching mechanism but in #307 we are killing it so there is an opportunity to stick that piece and drop ```uncheckedRun``` once and for all.

However, note that if your sizes don't match, we will have to raise an error for now: you must fully JIT TCs for now until support for more parametric TCs is implemented. Note that [this example](https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/examples/tensordot.cc#L72) shows how to easily reuse pre-tuned options for different sizes, as-is, with reasonable perf, or as a starting point for a short tuning specialization run.

As @abadams had also suggested at some point we can also use that strategy for tuning for a neighborhood and synthesize parametric kernels. We should definitely explore that route if only to benchmark its feasibility in general and the performance differences we see in parametric vs non-parametric cases.

nicolasvasilache: #352

PR #329, opened 2018-04-20
switch to revisionistic ATen

skimo-openhub:
The ATen repository appears to have been completely rewritten!
Switch to the commit with the same commit message in the new history.
The CMakeLists.txt file now appears in the aten subdirectory.


skimo-openhub:
It seems caffe needs to update its aten submodule first.

PR #332, opened 2018-04-23
drop unused HALIDE_PREFIX cmake variable

skimo-openhub:
The include and library directories used to be derived from
this variable, but they have since been replaced in prehistory
by calls to find_path and find_library, with the first one
not using any path at all and the second one using
CMAKE_INSTALL_PREFIX, which has the same value as HALIDE_PREFIX.


skimo-openhub:
It's not clear what the intention was, but the current situation doesn't make sense.

PR #336, opened 2018-04-24
extract out generic insertNode{Above,Below}

skimo-openhub:
This reduces the number of insert*{Above,Below} functions and
allows the functions to be reused for additional types.


ftynse:
tests pass /@ftynse

PR #337, opened 2018-04-24
build.sh fixes

ftynse:
* only use ccache for nvcc when an alias exists (unbreaks compilation with default ccache on debians)
* avoid calling "find" on an empty `CONDA_PREFIX` (fixes the problem with `./build.sh --all` not recompiling the third-party libraries after changes unless both build and installation paths are cleaned)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -275,7 +275,7 @@ function install_caffe2() {
       CMAKE_ARGS+=("-DCUDA_ARCH_NAME='Maxwell'")
     fi
 
-    if [[ $(which ccache | wc -c) -ne 0 ]]; then
+    if [[ -L ${CCACHE_WRAPPER_DIR}/nvcc && $($(readlink ${CCACHE_WRAPPER_DIR}/nvcc) --version | grep ccache | wc -c) -ne 0 ]]; then
       CMAKE_ARGS+=("-DCUDA_NVCC_EXECUTABLE=${CCACHE_WRAPPER_DIR}/nvcc")
     fi
 
--- a/build.sh
+++ b/build.sh
@@ -467,7 +467,7 @@ function install_halide() {
 }
 
 if ! test -z $gflags || ! test -z $all; then
-  if [[ $(find $CONDA_PREFIX -name libgflags.so) ]]; then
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -275,7 +275,7 @@ function install_caffe2() {
       CMAKE_ARGS+=("-DCUDA_ARCH_NAME='Maxwell'")
     fi
 
-    if [[ $(which ccache | wc -c) -ne 0 ]]; then
+    if [[ $(which ccache | wc -c) -ne 0 && -e ${CCACHE_WRAPPER_DIR}/nvcc ]]; then
```
skimo-openhub: I was going to approve this as obvious, but after a closer look, the original check
should simply be replaced.  The fact that there is some `ccache` is completely irrelevant.

Maybe something like
`-L ${CCACHE_WRAPPER_DIR}/nvcc && $($(readlink ${CCACHE_WRAPPER_DIR}/nvcc) --version | grep ccache | wc -c) -ne 0`

Btw, I fixed a minor typo in the other commit message.

ftynse: I don't have a ccache with nvcc so I canot test if that would work...

skimo-openhub:
On Tue, Apr 24, 2018 at 08:39:39AM -0700, ftynse wrote:
> > @@ -275,7 +275,7 @@ function install_caffe2() {
>        CMAKE_ARGS+=("-DCUDA_ARCH_NAME='Maxwell'")
>      fi
>  
> -    if [[ $(which ccache | wc -c) -ne 0 ]]; then
> +    if [[ $(which ccache | wc -c) -ne 0 && -e ${CCACHE_WRAPPER_DIR}/nvcc ]]; then
> 
> I don't have a ccache with nvcc so I canot test if that would work...

I do:

$ if test $($(readlink /home/skimo/bin/nvcc) --version | grep ccache | wc -c) -ne 0; then echo ok; fi
ok

skimo

ftynse:
cool, thanks, updated

prigoyal:
looks fine to me. thanks.

ftynse:
@prigoyal do you mind pressing "approve" then?

Issue #294, opened 2018-04-11
MappingOptions C++ output broken after splitting off Cuda part


ftynse:
Originally, `MappingOptions` could be printed as compilable C++ code, which we used in tests.  After splitting off `CudaMappingOptions`, the output looks like this

```
W0411 17:14:07.169840  6574 genetic_tuning_harness.cc:420] tc::MappingOptions::makeNaiveMappingOptions()
W0411 17:14:07.169909  6574 genetic_tuning_harness.cc:420]     .outerScheduleFusionStrategy(tc::FusionStrategy::Preserve3Coincident)
W0411 17:14:07.169976  6574 genetic_tuning_harness.cc:420]     .outerScheduleAllowSkewing(false)
W0411 17:14:07.170042  6574 genetic_tuning_harness.cc:420]     .outerSchedulePositiveOrthant(true)
W0411 17:14:07.170109  6574 genetic_tuning_harness.cc:420]     .intraTileScheduleFusionStrategy(tc::FusionStrategy::Min)
W0411 17:14:07.170174  6574 genetic_tuning_harness.cc:420]     .intraTileScheduleAllowSkewing(false)
W0411 17:14:07.170239  6574 genetic_tuning_harness.cc:420]     .intraTileSchedulePositiveOrthant(true)
W0411 17:14:07.170306  6574 genetic_tuning_harness.cc:420]     .tile(8, 16, 1)
W0411 17:14:07.170372  6574 genetic_tuning_harness.cc:420]     .unroll(4)
W0411 17:14:07.170436  6574 genetic_tuning_harness.cc:420]     .tileImperfectlyNested(false)
W0411 17:14:07.170503  6574 genetic_tuning_harness.cc:420]     .matchLibraryCalls(true);
W0411 17:14:07.170569  6574 genetic_tuning_harness.cc:420] 
W0411 17:14:07.170635  6574 genetic_tuning_harness.cc:420]     .mapToThreads(29, 15)
W0411 17:14:07.170701  6574 genetic_tuning_harness.cc:420]     .mapToBlocks(3)
W0411 17:14:07.170769  6574 genetic_tuning_harness.cc:420]     .useSharedMemory(true)
W0411 17:14:07.170836  6574 genetic_tuning_harness.cc:420]     .usePrivateMemory(false)
W0411 17:14:07.170902  6574 genetic_tuning_harness.cc:420]     .unrollCopyShared(true);
```

[...]

PR #340, opened 2018-04-24
Fix MappingOptionsPrinter

nicolasvasilache:
This PR addresses #294 by making the base mapping printer not print a newline
This is safe only if one cannot print the base class so this changeset also:
1. drops the operator<< functions that operate on the base class;
2. adds a pure virtual destructor to the base class;
3. implements CudaMappingOptionsCppPrinter in terms of the parent print function

We now correctly see entries such as:
```
[TUNER][GENERATION LOG] best option so far:
tc::MappingOptions::makeNaiveMappingOptions()
.outerScheduleFusionStrategy(tc::FusionStrategy::Max)
.outerScheduleAllowSkewing(false)
.outerSchedulePositiveOrthant(true)
.intraTileScheduleFusionStrategy(tc::FusionStrategy::Min)
.intraTileScheduleAllowSkewing(false)
.intraTileSchedulePositiveOrthant(true)
.tile(256, 13)
.unroll(64)
.tileImperfectlyNested(false)
.matchLibraryCalls(true)
.mapToThreads(25, 7)
.mapToBlocks(128)
.useSharedMemory(true)
.usePrivateMemory(true)
.unrollCopyShared(true);
```

### DIFF
--- a/tc/core/cuda/cuda_mapping_options_cpp_printer.h
+++ b/tc/core/cuda/cuda_mapping_options_cpp_printer.h
@@ -38,6 +38,8 @@ class CudaMappingOptionsCppPrinter : public MappingOptionsCppPrinter {
   CudaMappingOptionsCppPrinter(std::ostream& out, size_t ws = 0)
       : MappingOptionsCppPrinter(out, ws) {}
 
+  ~CudaMappingOptionsCppPrinter() = default;
+
   friend CudaMappingOptionsCppPrinter& operator<<(
       CudaMappingOptionsCppPrinter& prn,
       const CudaMappingOptions& options);
--- a/tc/core/mapping_options_cpp_printer.cc
+++ b/tc/core/mapping_options_cpp_printer.cc
@@ -19,6 +19,8 @@
 
 namespace tc {
 
+MappingOptionsCppPrinter::~MappingOptionsCppPrinter() = default;
+
 MappingOptionsCppPrinter& MappingOptionsCppPrinter::printSchedulerOptions(
     const SchedulerOptionsView& schedulerOptions,
     const std::string& prefix) {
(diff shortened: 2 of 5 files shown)


tc/core/mapping_options_cpp_printer.cc
```diff
@@ -32,34 +34,25 @@ MappingOptionsCppPrinter& MappingOptionsCppPrinter::printSchedulerOptions(
...
-MappingOptionsCppPrinter& operator<<(
-    MappingOptionsCppPrinter& prn,
-    const std::string& str) {
-  prn.printString(str);
-  return prn;
-}
-
-MappingOptionsCppPrinter& operator<<(
-    MappingOptionsCppPrinter& prn,
+MappingOptionsCppPrinter& MappingOptionsCppPrinter::print(
```
ftynse: do we really need to return `*this` from `print`?

tc/core/mapping_options_cpp_printer.cc
```diff
@@ -32,34 +34,25 @@ MappingOptionsCppPrinter& MappingOptionsCppPrinter::printSchedulerOptions(
...
     const MappingOptions& options) {
-  prn.printString("tc::MappingOptions::makeNaiveMappingOptions()")
+  this->printString("tc::MappingOptions::makeNaiveMappingOptions()")
```
ftynse: I wouldn't use `this->` here, we are in a quite simple member function

tc/core/cuda/cuda_mapping_options_cpp_printer.h
```diff
@@ -38,6 +38,8 @@ class CudaMappingOptionsCppPrinter : public MappingOptionsCppPrinter {
   CudaMappingOptionsCppPrinter(std::ostream& out, size_t ws = 0)
       : MappingOptionsCppPrinter(out, ws) {}
 
+  ~CudaMappingOptionsCppPrinter() {}
```
ftynse: wouldn't `=default` work here?

tc/core/mapping_options_cpp_printer.cc
```diff
@@ -19,6 +19,8 @@
 
 namespace tc {
 
+MappingOptionsCppPrinter::~MappingOptionsCppPrinter() {}
```
ftynse: wouldn't `=default` in the class declaration work?

ftynse:
typo in the commit message "by mapping printer not print a newline" => "not printing"

I would also cut `I0424 14:30:56.935150 41378 genetic_tuning_harness.cc:536] ` part from the sample output because it does not fit into 80 lines and is not relevant for the message.  Also, no need for "```" in the message.

tc/core/mapping_options_cpp_printer.cc
nicolasvasilache: This is for consistency with the other functions.
I am fine not returning but is there something special about ```print``` vs say ```printSchedulerOptions```?

tc/core/mapping_options_cpp_printer.cc
nicolasvasilache: it's pure, how would you write this:
```
~MappingOptionsCppPrinter() = 0, default;
```
?

PR #341, opened 2018-04-24
TC: Sync fbcode to github

mingzhe09088:
This diff syncs changes in Fbcode to Github.

### DIFF
--- a/tensor_comprehensions/pybinds/pybind_options.cc
+++ b/tensor_comprehensions/pybinds/pybind_options.cc
@@ -89,7 +89,7 @@ PYBIND11_MODULE(mapping_options, m) {
           },
           "Require TC to try and execute different TC expressions interleaved (Max), separately (Min)\nor interleaved as long as sufficient parallelism is exploited (Preserve3Coincident) by\nperforming loop fusion and fission. Applies before tiling")
       .def(
-          "serializeToProtobuf",
+          "serialize",
           [](tc::CudaMappingOptions& instance) {
             std::string str = instance.toProtobufSerializedString();
             return py::bytes(str);
--- a/test_python/test_c2.py
+++ b/test_python/test_c2.py
@@ -17,8 +17,9 @@ import unittest, os
 import numpy as np
 import hypothesis.strategies as st
 import caffe2.python.hypothesis_test_util as hu
+import tensor_comprehensions as tc
 
-from hypothesis import given
+from hypothesis import given, settings
 from caffe2.python import core, dyndep
 
 
@@ -28,6 +29,11 @@ if CONDA_PREFIX:
 else:
     dyndep.InitOpsLibrary("@/tc/tc:tc_c2")
 
+MATMUL_LANG = """
+def matmul(float(M,N) A, float(N,K) B) -> (output) {
+output(i, j) +=! A(i, kk) * B(kk, j)
+}
+"""
 
 class TestCaffe2(hu.HypothesisTestCase):
     @given(n=st.integers(1, 128),
@@ -38,14 +44,6 @@ class TestCaffe2(hu.HypothesisTestCase):
     def test_matmul(self, n, m, k, seed, gc, dc):
         np.random.seed(seed)
 
-        tc_forward = """
-        def matmul(float(M,N) A, float(N,K) B) -> (output) {
-          output(i, j) +=! A(i, kk) * B(kk, j)
-        }
-        """
-
-        # TODO: (prigoyal) serialize the options
-        # options = Options("mlp")
         X = np.random.rand(m, k).astype(np.float32)
         W = np.random.rand(k, n).astype(np.float32)
 
@@ -54,7 +52,7 @@ class TestCaffe2(hu.HypothesisTestCase):
 
         op = core.CreateOperator(
             "TcOp", ["X", "Y"], "out",
-            tcDef=tc_forward,
+            tcDef=MATMUL_LANG,
             tcName="matmul",
         )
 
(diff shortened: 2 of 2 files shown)


test/test_lang.cc
```diff
@@ -155,6 +157,8 @@ void testTcFormat() {
   auto def_tree = Parser(source).parseFunction();
   std::ostringstream s;
   tcFormat(s, def_tree);
+  LOG(INFO) << s.str();
```
nicolasvasilache: Tests shouldn't print, just test.
Is this debug information you forgot to clean up?
Looks like this file should be untouched.

test_python/test_c2.py
```diff
@@ -17,8 +17,9 @@
 import numpy as np
 import hypothesis.strategies as st
 import caffe2.python.hypothesis_test_util as hu
+import tensor_comprehensions as tc
```
nicolasvasilache: These changes I have tested within fbcode and things worked there.
Note the perf aspects in this test dependent on #327 being reviewed and landed.

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -89,7 +89,7 @@ PYBIND11_MODULE(mapping_options, m) {
           },
           "Require TC to try and execute different TC expressions interleaved (Max), separately (Min)\nor interleaved as long as sufficient parallelism is exploited (Preserve3Coincident) by\nperforming loop fusion and fission. Applies before tiling")
       .def(
-          "serializeToProtobuf",
+          "serialize",
```
nicolasvasilache: I see this isn't used in any .py file so it's fine to improve the name, thanks!

mingzhe09088:
@nicolasvasilache I've cleaned it up. Please take a look.

nicolasvasilache:
Thanks @mingzhe09088 , could you please squash the 2 commits as per our [coding conventions](https://github.com/facebookresearch/TensorComprehensions/blob/master/CodingConventions.md#commits)?

tensor_comprehensions/pybinds/pybind_options.cc
prigoyal: my suggestion would be to keep `serializeToProtobuf` since its more descriptive. Unless there is strong motivation to use `serialize`, let's stick with `serializeToProtobuf` or something like `serialize_to_proto` etc. to keep it descriptive

test_python/test_c2.py
```diff
@@ -65,6 +63,42 @@ def ref(X, W):
...
         )
 
+    @given(n=st.integers(1, 128),
+           m=st.integers(1, 128),
+           k=st.integers(1, 128),
+           seed=st.integers(min_value=0, max_value=2**32 - 1),
+           **hu.gcs_gpu_only)
+    @settings(max_examples=2)
+    def test_matmul_tune_and_run(self, n, m, k, seed, gc, dc):
+        matmul = tc.define(MATMUL_LANG, name="matmul")
```
prigoyal: this test is bit unclear to me. It looks like both pytorch and C2 pybinds are being used and doing redundant work here. Is the goal to use pytorch autotuner API to get the options and use them? If that's the case, I recommend to add a `genetic_autotuner_caffe2.h` following https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/autotuner/genetic_autotuner_aten.h this should be long term stable solution and doesn't require much coding. Ping me if you want to do this.

test_python/test_c2.py
```diff
@@ -65,6 +63,42 @@ def ref(X, W):
...
+        W = np.random.rand(k, n).astype(np.float32)
+
+        def ref(X, W):
+            return [np.dot(X, W)]
+
+        op = core.CreateOperator(
+            "TcOp", ["X", "Y"], "out",
+            tcDef=MATMUL_LANG,
+            tcName="matmul",
+            mappingOptions=mapping_options.serialize(),
```
prigoyal: you could add a bool parameter here:

`autotune=True` and if its true, then the `genetic_autotuner_caffe2.h` should be called to first tune the kernel on the given tensors. The tuned options can then be cached in memory [this is done for pytorch bindings already in tc_unit.py] and tuning can start from those options to likely yield better kernel OR you can just use those tuned options without re-tuning.

test_python/test_c2.py
```diff
@@ -65,6 +63,42 @@ def ref(X, W):
             reference=ref,
         )
 
+    @given(n=st.integers(1, 128),
```
prigoyal: can you add 1-2 lines for the logic behind how autotuner works atm? I think that the current solution is not stable and a better solution can be coded up easily/quickly.

test/test_lang.cc
```diff
@@ -157,8 +155,6 @@ void testTcFormat() {
   auto def_tree = Parser(source).parseFunction();
   std::ostringstream s;
   tcFormat(s, def_tree);
-  LOG(INFO) << s.str();
```
prigoyal: just curious: what's the reason for removing this logging? does it spew a lot?

prigoyal:
@mingzhe09088 , is there a script in place to sync up fbcode / github? or is this sync manual?

skimo-openhub:
On Tue, Apr 24, 2018 at 07:34:21PM -0700, Nicolas Vasilache wrote:
> Thanks @mingzhe09088 , could you please squash the 2 commits as per our [coding conventions](https://github.com/facebookresearch/TensorComprehensions/blob/master/CodingConventions.md#commits)?

Please also add a commit message (as per the same section).

skimo

mingzhe09088:
@prigoyal There is a sync script to do that. However, current TC in FBcode is 2 weeks behind OSS, so there were some unexpected issues when using the script. To avoid this issue and taking advantage of the script, I will sync more often.

tensor_comprehensions/pybinds/pybind_options.cc
salexspb: @prigoyal , IMO it is still lacking the information about format it is going to be serialized into (i.e. text vs. binary). Also it doesn't serialize to protobuf, it serializes a protobuf to string. Usually in serialize() method I haven't seen people specifying a full description of the serialization format. 

This is why I changed it :) Also it is not used anywhere in any case. 

And regarding the format I think we should also change it to text, so users can print it easily from python side.

test_python/test_c2.py
salexspb: @prigoyal , let me actually respectfully disagree. I think this code is an example how bindings can be reused across frameworks without need to introduce new classes.

test_python/test_c2.py
salexspb: @prigoyal , I like this being explicit step instead of tuning manually when the op creates. At least for now. Because then it is clear for users that they need to tune, without tuning it is arbitrary slow.

salexspb:
@skimo-openhub , next time sync should pull out commit messages, yes. This time it is manual, as Mingzhe pointed out. Here it is just a bunch of random changes to get Caffe2 integration working + tests for this integration.

prigoyal:
thanks @salexspb. we should meet and discuss more on how to make the pybinds shared between two things. 

my recommendation would be to develop C2 integration on github and establish sync scripts. We should figure out things needed for that and find someone to work on those. 

@mingzhe09088 , sounds good about the sync scripts. thanks for the update.

PR #342, opened 2018-04-25
[RFC] API for the autotuner

nicolasvasilache:
This makes numerous changes to the autotuner:
1. since we don't have execution engines anymore, the tuning harness owns
the executors resulting from compilation. This results in significantly
simpler concurrency implementation and allows the pruning function to not
leak to the executors
2. the autotuner and the harness are now templated by the backend type
3. the autotuner is also templated by the search type
4. the autotuner can only be created from a TC string and only exposes the
function ```tune```
5. anything related to multiple starting points was premature optimization and
disappears. We can make baseMapping a vector in the future if we want to
6. backing options caches are loaded and stored automatically if a filename
is passed

### DIFF
--- a/tc/autotuner/autotuner.h
+++ b/tc/autotuner/autotuner.h
@@ -1,0 +1,229 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <atomic>
+#include <csignal>
+#include <memory>
+#include <queue>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "tc/autotuner/genetic_search.h"
+#include "tc/autotuner/options_cache.h"
+#include "tc/autotuner/parameters.h"
+#include "tc/autotuner/utils.h"
+#include "tc/core/tensor.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/parser.h"
+
+namespace tc {
+namespace autotune {
+
+namespace detail {
+/**
+ * Internal harness to support multithreaded compilation and evaluation over
+ * various backends for a particular SearchStrategy.
+ * The SearchStrategy is a template parameter passed to the run method.
+ */
+template <typename Backend>
+class TuningHarness {
+ public:
+  using ExecutorType = typename Backend::ExecutorType;
+  using MappingOptionsType = typename Backend::MappingOptionsType;
+  using OptionsCacheType = OptionsCache<Backend>;
+
+  TuningHarness(
+      size_t maxPopulationSize,
+      lang::TreeRef tcTree,
+      const std::unordered_map<size_t, std::vector<const DLConstTensor*>>&
+          inputs,
+      std::unordered_map<size_t, std::vector<const DLTensor*>>& outputs,
+      const MappingOptionsType& baseMapping,
+      const TuningParameterFixer& fixedParams,
+      std::shared_ptr<OptionsCache<Backend>> optionsCache);
+
+  /// Runs a SearchStrategy
+  template <typename SearchStrategy>
+  void run(SearchStrategy& searchStrategy);
+
+  /// Once a signal has been caught, a flag is set and we terminate after the
+  /// current iteration is finished. This is apparently necessary for proper
+  /// python termination.
+  /// TODO: we should detect when we come from python and exit properly in C++.
+  void stopAfterCurrentIteration();
+
+ private:
+  /// Traverse one iteration of candidates in parallel and evaluate their
+  /// runtimes
+  template <typename SearchStrategy>
+  void runOneIteration(SearchStrategy& searchStrategy, size_t iteration);
+
+  /// Helper function to delegate compiling on the cpu to different threads
+  template <typename SearchStrategy>
+  void doCompile(SearchStrategy& searchStrategy);
+
+  /// Helper function to delegate running the compiled code to different
+  /// threads.
+  /// This function must be specialized per Backend
+  /// TODO: virtual inheritance?
+  void doEvaluate(size_t device, size_t populationSize, Printer& printer);
+
+  const MappingOptionsType& bestMappingOptions() {
+    std::lock_guard<std::mutex> lock(bestTimeMutex_);
+    return bestMappingOptions_;
+  }
+
+ private:
+  /// Synchronization related objects
+  /// The main invariant is that we always try to compile and evaluate
+  /// exactly searchStrategy->population.size() candidates.
+  /// If a candidate fails compilation we still add a null Executor so that
+  /// the invariant holds.
+  /// This way it is easy to implement multi-threaded termination by just
+  /// taking an atomic counter and pushing/popping the queues under lock until
+  /// we have evaluated searchStrategy->population.size() compilation results.
+  std::mutex bestTimeMutex_;
+  std::mutex executorsMutex_;
+  std::atomic_bool stopRequested_;
+  std::atomic_size_t currentCompilationJob_;
+  std::atomic_size_t numEvaluations_;
+  std::queue<std::unique_ptr<ExecutorType>> executors_;
+  std::queue<CandidateConfiguration*> configurations_;
+
+  /// inputs
+  lang::TreeRef kTcTree_;
+  const MappingOptionsType kBaseMapping_;
+  /// maps of inputs and outputs per each device (represented by a size_t)
+  /// involved in autotuning. The client of the autotuner API must allocate
+  /// these properly on each device where autotuning evaluation needs to run.
+  /// In particular all the inputs and outputs must contain the same values
+  /// across devices for the purpose of running correctness checks during
+  /// tuning (future work).
+  const std::unordered_map<size_t, std::vector<const DLConstTensor*>> kInputs_;
+  std::unordered_map<size_t, std::vector<const DLTensor*>> outputs_;
+
+  // results
+  size_t bestTime_;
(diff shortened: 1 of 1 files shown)


tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+  template <typename SearchStrategy>
+  void runOneIteration(SearchStrategy* searchStrategy, size_t iteration);
+
+  /// Helper function to delegate compiling on the cpu to different threads
+  template <typename SearchStrategy>
+  void doCompile(SearchStrategy* searchStrategy);
+
+  /// Helper function to get a kernel into benchmark-able state
+  /// This function must be specialized per Backend
+  /// TODO: virtual inheritance?
```
skimo-openhub: What's the alternative?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+  bool warmupOrPrune(
```
skimo-openhub: What is the meaning of the result?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+
+  const MappingOptionsType& bestMappingOption() {
+    std::lock_guard<std::mutex> lock(bestTimeMtx_);
+    return bestMappingOptions_;
+  }
+
+ private:
+  static constexpr int kReducedWarmupIterations = 2;
+  static constexpr int kReducedBenchmarkIterations = 10;
+  static constexpr int kEarlyPruneFactor = 5;
```
skimo-openhub: Please add some explanation.
The meaning of these constants may be obvious from their names to you,
but not to me.

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+  std::atomic_size_t currentCompilationJob_;
+  std::atomic_size_t numEvaluations_;
+  std::queue<std::unique_ptr<typename Backend::ExecutorType>> executors_;
+  std::queue<CandidateConfiguration*> configurations_;
+
+  /// inputs
+  lang::TreeRef kTcTree_;
+  const MappingOptionsType kBaseMapping_;
+  const std::unordered_map<size_t, std::vector<const DLConstTensor*>> kInputs_;
+  std::unordered_map<size_t, std::vector<const DLTensor*>> outputs_;
```
skimo-openhub: What is the meaning of the key of these two maps?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+#include <atomic>
+#include <csignal>
+#include <deque>
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/aten/aten_compiler.h"
```
skimo-openhub: Why are these includes needed?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+
+  /// Runs autotuning on the TC function tcEntryPoint.
+  /// This assumes input and output tensors of proper sizes have been
+  /// initialized on each GPU. For now this initialization is left to the
+  /// client of the Autotuner.
+  /// Tuning performs a search of template type SearchStrategy starting from
+  /// baseMapping.
+  ///
+  /// Alternatively an OptionsCache cacheFileName serialized path
+  /// can be specified from which the tuner recover multiple starting points
```
skimo-openhub: "recovers"

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+  /// and preempt the baseMapping. This can be useful in a reinforcement
```
skimo-openhub: do you mean ", preempting" instead of "and preempt"?
Does that mean that `baseMapping` gets ignored in that case?
Wouldn't it then make more sense to provide either the baseMapping or
a cacheFileName, but not both?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+  /// situation where short tunings are ran and their result cached
```
skimo-openhub: "are run"
also, "results" instead of "result"?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+  using MappingOptionsType = typename BackendType::MappingOptionsType;
+  using OptionsCacheType = tc::autotune::OptionsCache<Backend>;
+
+  /// Parse and store the underlying tree representation for all the TC
+  /// functions provided.
+  explicit Autotuner(const std::string& tc);
+
+  /// Runs autotuning on the TC function tcEntryPoint.
+  /// This assumes input and output tensors of proper sizes have been
+  /// initialized on each GPU. For now this initialization is left to the
```
skimo-openhub: Do you mean "device" instead of "GPU"?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+ public:
+  using BackendType = Backend;
+  using ExecutorType = typename BackendType::ExecutorType;
+  using MappingOptionsType = typename BackendType::MappingOptionsType;
+  using OptionsCacheType = OptionsCache<Backend>;
+
+  TuningHarness(
+      size_t maxPopulationSize,
+      lang::TreeRef kTcTree,
+      const std::unordered_map<size_t, std::vector<const DLConstTensor*>>&
```
ftynse: Please document that these maps are per-device populated with actual data (same inputs)

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+    const typename Backend::MappingOptionsType& options,
+    const TuningConfiguration& configuration);
+
+template <typename Backend>
+void handleDeviceRuntimeError(
+    size_t device,
+    typename Backend::MappingOptionsType& options);
+} // namespace detail
+
+inline std::vector<size_t> parseDevices();
```
ftynse: `template <typename Backend>` here?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+      const typename Backend::MappingOptionsType& baseMapping,
+      const std::string& cacheFileName = "",
+      const TuningParameterFixer& fixedParams = TuningParameterFixer());
+
+ private:
+  std::map<std::string, lang::TreeRef> tcEntryPointMap_;
+  std::shared_ptr<OptionsCache<Backend>> optionsCache_;
+};
+
+/// Helper functions that need specializing for various backends.
```
ftynse: I'm not quite sure what these functions are supposed to do.  They return the value of the same type as one of their arguments, so looks like some sort of copying but why do we need the second argument then?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+      const std::unordered_map<size_t, std::vector<const DLConstTensor*>>&
+          inputs,
+      std::unordered_map<size_t, std::vector<const DLTensor*>>& outputs,
+      const typename Backend::MappingOptionsType& baseMapping,
+      const TuningParameterFixer& fixedParams,
+      std::shared_ptr<OptionsCache<Backend>>& optionsCache);
+
+  /// Runs a SearchStrategy over numIterations
+  template <typename SearchStrategy>
+  void run(SearchStrategy* searchStrategy);
```
ttheodor: Why a pointer? Do we expect `nullptr`s?

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,214 @@
...
+namespace detail {
+/**
+ * Internal harness to support multithreaded compilation and evaluation over
+ * various backends for a particular SearchStrategy.
+ * The SearchStrategy is a template parameter passed to the run method.
+ */
+template <typename Backend>
+class TuningHarness {
+ public:
+  using BackendType = Backend;
```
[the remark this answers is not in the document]
nicolasvasilache: we don't killing it

tc/autotuner/autotuner.h
nicolasvasilache: Renamed to ```skipExecution``` and documenting it.
In the current impl, the function is not defined in the template case.
It is specialized per backend, like:
```
using CudaTuningHarness = class TuningHarness<tc::CudaBackend>;

template <>
bool CudaTuningHarness::skipExecution(...) {}
```

This is not good because it results in link-time errors when a backend autotuner does not implement it.

This was mostly to avoid .h proliferation but it's too bad. I'll make proper derived CudaTuningHarness and use a pure virtual function in the base instead.

nicolasvasilache: An alternative would be to stick that function as a static member class in the ```Backend``` type which is maybe better. Actually this function is of the same category as the ones at the bottom of the file.

```makeOptions```, ```makeTuningConfiguration```, ```handleDeviceRuntimeError``` and ```skipExecutionOrWarmup``` are all Backend-specific behaviors I haven't really found a good place for.

Note these are Backend-specific but only used in tuning. Probably ```makeOptions``` and ```makeTuningConfiguration``` can go away if the ```TuningConfiguration``` and ```parameters``` are templated properly and made backend-specific too.

In the meantime I'll actually extract ```skipExecutionOrWarmup``` and put it close to the other functions where it belongs.

File: tc/autotuner/autotuner.h (hunk 7 above)
Comment (nicolasvasilache) [reply tc/autotuner/autotuner.h]: impl concats them, fixing the doc

File: tc/autotuner/autotuner.h (hunk 1 above)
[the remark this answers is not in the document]
Comment (nicolasvasilache) [reply tc/autotuner/autotuner.h]: I described my dilemma above, cleaned up this function

File: tc/autotuner/autotuner.h [hunk 15]
```diff
@@ -0,0 +1,214 @@
...
+  /// runtimes
+  template <typename SearchStrategy>
+  void runOneIteration(SearchStrategy* searchStrategy, size_t iteration);
+
+  /// Helper function to delegate compiling on the cpu to different threads
+  template <typename SearchStrategy>
+  void doCompile(SearchStrategy* searchStrategy);
+
+  /// Helper function to get a kernel into benchmark-able state
+  /// This function must be specialized per Backend
```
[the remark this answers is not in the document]
nicolasvasilache: and yet it was, see comments above, moving to a free function, that was indeed a smell check

tc/autotuner/autotuner.h
nicolasvasilache: This is a bottom-up refactoring, this legacy behavior is related to TuningConfigurations and will need to be cleaned up but I am punting atm.

nicolasvasilache:
served its purpose

PR #343, opened 2018-04-25
clean-ups in the mapper

skimo-openhub:
Some of these clean-ups prepare for getting rid of `ThreadIdxXScheduleDepthState`,
but they should also be useful independently.

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -332,45 +332,36 @@ size_t MappedScop::mapToThreads(detail::ScheduleTree* band) {
     return 0;
   }
 
-  size_t nMappedReductionThreads = 0;
-  if (reductionBandUpdates_.count(band) == 1) {
-    // A reduction is assumed to get mapped to threadIdx.x
-    CHECK(reductionBandUpdates_.at(band).separated);
-    auto reductionDim = reductionBandUpdates_.at(band).reductionDim;
-    threadIdxXScheduleDepthState.emplace_back(std::make_pair(
-        activeDomainPoints(schedule(), band),
-        band->scheduleDepth(schedule()) + reductionDim));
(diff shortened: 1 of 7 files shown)


tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -171,34 +171,9 @@ void fixThreadsBelowFilter(
...
-      // Mapping happened below filterTree, so we need points active for its
-      // children.  After insertion, filterTree is guaranteed to have at least
-      // one child.
-      mscop.threadIdxXScheduleDepthState.emplace_back(std::make_pair(
-          activeDomainPoints(mscop.schedule(), filterTree->child({0})),
-          filterTree->scheduleDepth(mscop.schedule())));
-    }
-  }
+  auto band = detail::ScheduleTree::makeEmptyBand(mscop.scop().scheduleRoot());
+  auto bandTree = insertNodeBelow(filterTree, std::move(band));
```
ftynse: do we want to keep this empty band for some further use, or can we just drop it after calling `mapRemaining`?

ftynse: A little bit.  It serves no other purpose than performing one transformation that expects a band to be present.  I hope we won't need special cases to handle zero-dimensional bands elsewhere in the code.

skimo-openhub:
On Thu, Apr 26, 2018 at 12:58:06AM -0700, ftynse wrote:
> A little bit.  It serves no other purpose than performing one transformation that expects a band to be present.  I hope we won't need special cases to handle zero-dimensional bands elsewhere in the code.

I don't see why we would need any special casing
for zero-dimensional bands.  That shoud just work.
(Again, we already keep around zero-dimensional bands
in some cases.)
In fact, being forced to remove the band very much
sounds like special casing to me, but if that's what
you prefer, I'll add the extra code to remove the band.

skimo

ftynse:
let's handle problems when/if they appear

PR #344, opened 2018-04-25
replace ThreadIdxXScheduleDepthState by thread specific markers

skimo-openhub:
ThreadIdxXScheduleDepthState maintains duplicate state and
does so in an inconsistent way.  In particular, when at least
one member is mapped to a thread identifier, it holds
the depth of the member mapped to the x thread identifier
(which is one less that the schedule depth of the newly introduced marker),
but if no member is mapped, then it points to the depth
of where the marker is introduced.

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -705,8 +705,7 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
 
   // 8. Promote to registers below the loops mapped to threads.
   if (cudaOptions.proto().use_private_memory()) {
-    promoteToRegistersBelowThreads(
-        mappedScop->scop(), mappedScop->threadIdxXScheduleDepthState, -1ull);
+    promoteToRegistersBelowThreads(mappedScop->scop(), -1ull);
   }
 
   // 9. Insert mapping context
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -637,7 +637,7 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
     auto child = outerBand->child({0});
     size_t numMappedInnerThreads =
         mappedScop->mapInnermostBandsToThreads(child);
-    mappedScop->mapRemaining<mapping::ThreadId>(child, numMappedInnerThreads);
+    fixThreadsBelow(*mappedScop, outerBand, numMappedInnerThreads);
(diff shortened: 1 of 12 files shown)


skimo-openhub:
On Thu, Apr 26, 2018 at 01:23:11AM -0700, ftynse wrote:
> >   * accessed in a coalesced way if all references in this group are accessed in
>   * a coalesced way.
> + * If a branch is mapped to a single thread, then all accesses
> + * in that branch are also considered to be coalesced.
> 
> Can we then rename `isCoalesced` into something like `isAccessCoalescedOrSingleThread` ?  I understand the motivation for considering accesses from a single thread as coalesced for the sake of promotion, but they are not really coalesced.  So if somebody decides to reuse that function, it may come as a surprise.

If we are going to rename it, then let's use a name that reflects
what the caller actually want to know.
How about `wantsPromotionForCoalescing`? (equal to !isCoalesced)

skimo

skimo-openhub:
On Thu, Apr 26, 2018 at 08:27:14AM +0000, ftynse wrote:
> 
> Given that markers are now inserted below the shared memory promotion copies mapped to threads, it looks like register promotion will kick in for those.  I suppose there will be no actual promotion due to the lack of reuse, but I wonder if we could spare the call to the entire promotion code there.

Do you mean that you want me to drop the call to ThreadIdxXScheduleDepthState
from mapCopiesToThreads for this PR?

skimo

ftynse:
> How about `wantsPromotionForCoalescing`? (equal to !isCoalesced)

OK. Or `promotionImprovesCoalescing`.

> Do you mean that you want me to drop the call to ThreadIdxXScheduleDepthState from mapCopiesToThreads for this PR?

Discussed separately. Conclusion: keep as is.  If register promotion checks (that should never promote copies due to the lack of reuse) take significant time, do the additional check if we are in the "main" tree or in an extension before running those checks.

skimo-openhub:
On Thu, Apr 26, 2018 at 06:03:41AM -0700, ftynse wrote:
> > How about `wantsPromotionForCoalescing`? (equal to !isCoalesced)
> 
> OK. Or `promotionImprovesCoalescing`.
> 
> > Do you mean that you want me to drop the call to ThreadIdxXScheduleDepthState from mapCopiesToThreads for this PR?
> 
> Discussed separately. Conclusion: keep as is.  If register promotion checks (that should never promote copies due to the lack of reuse) take significant time, do the additional check if we are in the "main" tree or in an extension before running those checks.

OK, I updated the commit message of
e11308da5cf23a31a0047abf8424a09bbf2dd8c5,
introduced a new commit
a28c9e9d1318fe8ca600ddf2fc8db8a282d3f87d
and updated 71103100b8d9c5b28161456e0f552020666c0ca5
accordingly.

skimo

PR #345, opened 2018-04-25
[RFC] Proposed API for user-facing autotuner

nicolasvasilache:
After iterating on a bottom up refactoring and rounds of API RFCs I am now ready
to propose this as the model for the user-facing C++ API for autotuning.

This proposed API essentially:
1. makes the tuner future-proof by templating by backend type and search strategy
2. drops the user-exposed load function which retrieves options from a
serialized proto cache: the tuner has no business providing this functionality.
If one wants to query the options cache one can do so explicitly
3. drops the owning pointer to a genetic autotuner: an ATenAutotuner **is a**
Autotuner<Backend, SearchStrategy> that operates on ATen tensors.
4. adds documentation and example of usage

Compare to the current API [here](https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/autotuner/genetic_autotuner_aten.h).


nicolasvasilache:
also cc @mingzhe09088

nicolasvasilache:
@ftynse the tuner should not provide the functionality of loading options from caches and returning them (this is the Options Cache job).
But it can certainly store the results it found to caches and load a preexisting cache file to resume autotuning.

ftynse:
Let's document that then.  Existing documentation does not specify that the given file will also be used for _storing_ the options that were tested, and what happens if there are already some options from which the tuner is populated.

LGTM otherwise.

mingzhe09088:
One question I have is do we allow user giving multiple previously saved files for autotuner? If a user has been using TC for a long time, he or she might have several saved tuning files.  Does the user have to pick the right tuning file for autotuner to start with?

nicolasvasilache:
@mingzhe09088 this sounds like a good candidate for the options cache API to merge multiple saved proto files in a single one and save as a file or return an in-memory representation. 

To avoid forcing the user to write that to disk, maybe a good API would be to pass in an options cache to the tune function instead of a file name.

Does that sound reasonable?

If so, I'd propose to provide this in a follow-up PR/issue since it touches multiple parts of the codebase and is more easily added once refactoring has landed.

mingzhe09088:
Yes. Passing options cache instead of a file name sounds good to me.

ttheodor:
> this sounds like a good candidate for the options cache API to merge multiple saved proto files in a single one and save as a file or return an in-memory representation.

I already have a [branch](https://github.com/facebookresearch/TensorComprehensions/tree/cache_merger) with that.  It's rebased on top of the CUPTI stuff, but if you think it's useful I can clean it up and open a PR.

tc/aten/aten_autotuner.h
```diff
@@ -0,0 +1,94 @@
...
+  /// protobuf file.
+  ///
+  /// Lastly a TuningParameterFixer function can be specified to limit the
+  /// search space (i.e. when certain parameters are known to be good/bad
+  /// independently on a particular TC).
+  ///
+  /// \return a vector MappingOptions, if it is empty then tuning did not find
+  /// a single good configuration. This should be a very rare occurrence but
+  /// it is possible in particular if the skipExecutionOrWarmup function is too
+  /// aggressive and the problem size is too small.
```
skimo-openhub: What kind of MappingOptions does it return if the vector is not empty
and in what order?

tc/aten/aten_autotuner.h
```diff
@@ -0,0 +1,94 @@
...
+      const std::string& tcEntryPoint,
+      const std::vector<at::Tensor>& inputs,
+      const MappingOptionsType& baseMapping,
+      const std::string& cacheFileName = "",
+      const tc::autotune::TuningParameterFixer& fixedParams = {});
+
+ private:
+  /// The TC string is stored internally so we can tune independent TC
+  /// functions on demand.
+  const std::string kTc_;
```
skimo-openhub: I think @ftynse asked about this before, but isn't the "k" prefix used for (compile time) constants?

tc/aten/aten_autotuner.h
```diff
@@ -0,0 +1,94 @@
...
+};
+
+} // namespace aten
+} // namespace tc
+
+#include "tc/aten/aten_autotuner-inl.h"
```
skimo-openhub: Where is `tc/aten/aten_autotuner-inl.h`?

PR #348, opened 2018-04-25
Add % operator and propagate it from parser -> halide -> isl

### LINKED ISSUE #290: The % operator does not seem to work
Hello!
```python
CONV_LANG = """
def local_sparse_convolution(float(N,C,H,W) I, float(M,KC,KH,KW) W1) -> (O) {{
	O(n, m, h, w) +=! I(n, (c + kc) % c, {sh} * h + kh, {sw} * w + kw) * W1(m, kc, kh, kw)
}}
"""
```
Gives the following error 
`RuntimeError: expected a valid token but found %`

prigoyal:
adding % operator support

closes #290

### DIFF
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -203,7 +203,7 @@ std::vector<isl::aff> makeIslAffBoundsFromExpr(
     std::vector<isl::aff> result;
     // We cannot span multiple constraints if a modulo operation is involved.
     // x > max(a,b) % C is not equivalent to (x > a % C && x > b % C).
-    auto lhs = makeIslAffBoundsFromExpr(space, e, false, false);
+    auto lhs = makeIslAffBoundsFromExpr(space, op->a, false, false);
     CHECK_EQ(lhs.size(), 1u);
     if (const int64_t* b = as_const_int(op->b)) {
       return {lhs[0].mod(isl::val(space.get_ctx(), *b))};
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -34,7 +34,8 @@ using namespace tc::polyhedral::detail;
 
 SymbolTable makeSymbolTable(const tc2halide::HalideComponents& components) {
   // const Stmt& s) {
-  // Collect and categorize all the Variable symbols
+  // Collect and categorize all the Halide Variable symbols as reduction
+  // or index variables
   class BuildSymbolTable : public IRVisitor {
     using IRVisitor::visit;
     std::set<std::string> included;
@@ -59,7 +60,7 @@ SymbolTable makeSymbolTable(const tc2halide::HalideComponents& components) {
 
   components.stmt.accept(&builder);
   // Get params from components.params which contain everything declared in
-  // tcdef. However, the 0-D tensors are registered as both params and inputs,
+  // TC Def. However, the 0-D tensors are registered as both params and inputs,
   // filter those out.
   for (auto kvp : components.params) {
     bool skip = false;
--- a/tc/core/tc2halide.cc
+++ b/tc/core/tc2halide.cc
@@ -158,6 +158,8 @@ Expr translateExpr(
       return t(0) * t(1);
     case '/':
       return t(0) / t(1);
+    case '%':
+      return t(0) % t(1);
     case lang::TK_MIN:
       return min(t(0), t(1));
     case lang::TK_MAX:
--- a/tc/core/tc2halide.cc
+++ b/tc/core/tc2halide.cc
@@ -497,18 +497,18 @@ Expr reductionUpdate(Expr e) {
 // contain kReductionUpdate intrinsics.  These may have to be removed
 // in order to be able to apply internal Halide analysis passes on them.
 void translateComprehension(
-    const lang::Comprehension& c,
+    const lang::Comprehension& comprehension,
     const map<string, Parameter>& params,
     bool throwWarnings,
     map<string, Function>* funcs,
     FunctionBounds* bounds) {
   Function f;
-  auto it = funcs->find(c.ident().name());
+  auto it = funcs->find(comprehension.ident().name());
   if (it != funcs->end()) {
     f = it->second;
   } else {
-    f = Function(c.ident().name());
-    (*funcs)[c.ident().name()] = f;
+    f = Function(comprehension.ident().name());
+    (*funcs)[comprehension.ident().name()] = f;
   }
   // Function is the internal Halide IR type for a pipeline
   // stage. Func is the front-end class that wraps it. Here it's
@@ -517,7 +517,7 @@ void translateComprehension(
 
   vector<Var> lhs;
   vector<Expr> lhs_as_exprs;
-  for (lang::Ident id : c.indices()) {
+  for (lang::Ident id : comprehension.indices()) {
     lhs.push_back(Var(id.name()));
     lhs_as_exprs.push_back(lhs.back());
   }
@@ -526,17 +526,17 @@ void translateComprehension(
   // in the future we may consider using Halide Let bindings when they
   // are supported later
   map<string, Expr> lets;
-  for (auto wc : c.whereClauses()) {
+  for (auto wc : comprehension.whereClauses()) {
     if (wc->kind() == lang::TK_LET) {
       auto let = lang::Let(wc);
       lets[let.name().name()] = translateExpr(let.rhs(), params, *funcs, lets);
     }
   }
 
-  Expr rhs = translateExpr(c.rhs(), params, *funcs, lets);
+  Expr rhs = translateExpr(comprehension.rhs(), params, *funcs, lets);
 
   std::vector<Expr> all_exprs;
-  for (auto wc : c.whereClauses()) {
+  for (auto wc : comprehension.whereClauses()) {
     if (wc->kind() == lang::TK_EXISTS) {
       all_exprs.push_back(
           translateExpr(lang::Exists(wc).exp(), params, *funcs, lets));
@@ -560,7 +560,7 @@ void translateComprehension(
   // values (2) +=!, TK_PLUS_EQ_B which first sets the tensor to the identity
   // for the reduction and then applies the reduction.
   bool should_zero = false;
-  switch (c.assignment()->kind()) {
+  switch (comprehension.assignment()->kind()) {
     case lang::TK_PLUS_EQ_B:
       should_zero = true; // fallthrough
     case lang::TK_PLUS_EQ:
@@ -592,12 +592,13 @@ void translateComprehension(
     case '=':
       break;
     default:
-      throw lang::ErrorReport(c) << "Unimplemented reduction "
-                                 << c.assignment()->range().text() << "\n";
+      throw lang::ErrorReport(comprehension)
+          << "Unimplemented reduction "
+          << comprehension.assignment()->range().text() << "\n";
   }
 
   // Tag reductions as such
-  if (c.assignment()->kind() != '=') {
+  if (comprehension.assignment()->kind() != '=') {
     rhs = reductionUpdate(rhs);
   }
 
@@ -637,7 +638,7 @@ void translateComprehension(
   Scope<Interval> solution;
 
   // Put anything explicitly specified with a 'where' class in the solution
-  for (auto constraint_ : c.whereClauses()) {
+  for (auto constraint_ : comprehension.whereClauses()) {
     if (constraint_->kind() != lang::TK_RANGE_CONSTRAINT)
       continue;
     auto constraint = lang::RangeConstraint(constraint_);
@@ -658,7 +659,8 @@ void translateComprehension(
 
   // Infer the rest
   all_exprs.push_back(rhs);
-  forwardBoundsInference(all_exprs, *bounds, c, throwWarnings, &solution);
+  forwardBoundsInference(
+      all_exprs, *bounds, comprehension, throwWarnings, &solution);
 
   // TODO: What if subsequent updates have incompatible bounds
   // (e.g. an in-place stencil)?. The .bound directive will use the
@@ -669,7 +671,7 @@ void translateComprehension(
 
   for (Var v : lhs) {
     if (!solution.contains(v.name())) {
-      throw lang::ErrorReport(c)
+      throw lang::ErrorReport(comprehension)
           << "Free variable " << v
           << " was not solved in range inference. May not be used right-hand side";
     }
@@ -693,7 +695,7 @@ void translateComprehension(
     for (size_t i = 0; i < unbound.size(); i++) {
       auto v = unbound[unbound.size() - 1 - i];
       if (!solution.contains(v->name)) {
-        throw lang::ErrorReport(c)
+        throw lang::ErrorReport(comprehension)
             << "Free variable " << v << " is unconstrained. "
             << "Use a 'where' clause to set its range.";
       }
--- a/tc/core/tc2halide.cc
+++ b/tc/core/tc2halide.cc
@@ -71,37 +71,36 @@ void translateParam(
   // Check if the param has already been converted to halide components.
   if (params->find(p.ident().name()) != params->end()) {
     return;
-  } else {
-    lang::TensorType type = p.tensorType();
-    int dimensions = (int)type.dims().size();
-    ImageParam imageParam(
-        translateScalarType(type.scalarType()), dimensions, p.ident().name());
-    inputs->push_back(imageParam);
-    vector<Expr> dims;
-    for (auto d_ : type.dims()) {
-      if (d_->kind() == lang::TK_IDENT) {
-        auto d = lang::Ident(d_);
-        auto it = params->find(d.name());
-        Parameter p;
-        if (it != params->end()) {
-          p = it->second;
-        } else {
-          p = Parameter(Int(32), false, 0, d.name(), true);
-          (*params)[d.name()] = p;
-        }
-        dims.push_back(Variable::make(Int(32), p.name(), p));
-      } else {
-        CHECK(d_->kind() == lang::TK_CONST);
-        int32_t value = lang::Const(d_).value();
-        dims.push_back(Expr(value));
-      }
-    }
-
-    for (int i = 0; i < imageParam.dimensions(); i++) {
-      imageParam.dim(i).set_bounds(0, dims[i]);
-    }
-    (*params)[imageParam.name()] = imageParam.parameter();
   }
+  lang::TensorType type = p.tensorType();
+  int dimensions = (int)type.dims().size();
+  ImageParam imageParam(
+      translateScalarType(type.scalarType()), dimensions, p.ident().name());
+  inputs->push_back(imageParam);
+  vector<Expr> dims;
+  for (auto d_ : type.dims()) {
+    if (d_->kind() == lang::TK_IDENT) {
+      auto d = lang::Ident(d_);
+      auto it = params->find(d.name());
+      Parameter p;
+      if (it != params->end()) {
+        p = it->second;
+      } else {
+        p = Parameter(Int(32), false, 0, d.name(), true);
+        (*params)[d.name()] = p;
+      }
+      dims.push_back(Variable::make(Int(32), p.name(), p));
+    } else {
+      CHECK(d_->kind() == lang::TK_CONST);
+      int32_t value = lang::Const(d_).value();
+      dims.push_back(Expr(value));
+    }
+  }
+
+  for (int i = 0; i < imageParam.dimensions(); i++) {
+    imageParam.dim(i).set_bounds(0, dims[i]);
+  }
+  (*params)[imageParam.name()] = imageParam.parameter();
 }
 
 void translateOutput(
--- a/tc/core/tc2halide.cc
+++ b/tc/core/tc2halide.cc
@@ -62,10 +62,13 @@ Type translateScalarType(int tcType) {
   }
 }
 
+// Translate the TC def input params to corresponding Halide components.
+// params, inputs will be populated here.
 void translateParam(
     const lang::Param& p,
     map<string, Parameter>* params,
     vector<ImageParam>* inputs) {
+  // Check if the param has already been converted to halide components.
   if (params->find(p.ident().name()) != params->end()) {
     return;
   } else {
@@ -488,6 +491,9 @@ Expr reductionUpdate(Expr e) {
   return Call::make(e.type(), kReductionUpdate, {e}, Call::Intrinsic);
 }
 
+// Translate a single TC comprehension/statement to Halide components: funcs,
+// bounds, reductions.
+//
 // Note that the function definitions created by translateComprehension may
 // contain kReductionUpdate intrinsics.  These may have to be removed
 // in order to be able to apply internal Halide analysis passes on them.
@@ -736,6 +742,7 @@ void translateComprehension(
   stage.reorder(loop_nest);
 }
 
+// Translate a semantically checked TC def to HalideComponents struct.
 HalideComponents translateDef(const lang::Def& def, bool throwWarnings) {
   map<string, Function> funcs;
   HalideComponents components;
@@ -895,6 +902,8 @@ translate(isl::ctx ctx, const lang::TreeRef& treeRef, bool throwWarnings) {
       lang::Def(lang::Sema().checkFunction(treeRef)), throwWarnings);
 }
 
+// NOTE: there is no guarantee here that the tc string has only one def. It
+// could have many defs. Only first def will be converted in that case.
 HalideComponents
 translate(isl::ctx ctx, const std::string& tc, bool throwWarnings) {
   LOG_IF(INFO, tc::FLAGS_debug_halide) << tc;
--- a/tc/lang/sema.h
+++ b/tc/lang/sema.h
@@ -339,6 +339,21 @@ struct Sema {
         throw ErrorReport(exp) << "NYI - semantic checking for " << exp;
     }
   }
+  // This is the entry function for semantic analysis. It is called by
+  // tc2halide to associate type with each node of the tree and to also make
+  // sure that the tree is sematically correct. For example: a variable
+  // may not be input two times. Parser only verifies for the syntax but does
+  // not check the semantics.
+  //
+  // It converts the TK_APPLY nodes to TK_ACCESS or TK_BUILT_IN
+  //
+  // The reduction variables are deduced and the objects are created for them
+  // and they are appended to the tree
+  //
+  // Type checking is also done by small amount of code
+  //
+  // The method 'withType' can be used to associate the type with a given node
+  //
   TreeRef checkFunction(TreeRef func_) {
     auto func = Def(func_);
     auto params_ =
@@ -350,6 +365,10 @@ struct Sema {
       }
     }
 
+    // Everything has to be input or output. Keep track of the variables that
+    // are either input/output. We will check that the statements have variables
+    // from this list. If not, then throw error that temporaries are not yet
+    // implemented.
     for (auto p : func.params()) {
       nonTemporaries.insert(p.ident().name());
       inputParameters.insert(p.ident().name());
@@ -437,6 +456,7 @@ struct Sema {
       return checkRangeConstraint(RangeConstraint(ref));
     }
   }
+  // Semantic checking for the statements/comprehensions in a TC Def.
   TreeRef checkStmt(TreeRef stmt_) {
     auto stmt = Comprehension(stmt_);
 
--- a/tc/lang/sema.h
+++ b/tc/lang/sema.h
@@ -293,6 +293,7 @@ struct Sema {
       case '-':
       case '*':
       case '/':
+      case '%':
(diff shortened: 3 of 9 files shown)


test/cuda/test_execution_engine.cc
```diff
@@ -235,6 +235,24 @@ def convolution(float(N,C,H,W) I, float(O,C,KH,KW) W1, float(O) B)
...
+TEST_F(ATenCompilationUnitTest, LocalSparseConvolution) {
+  at::Tensor I = at::CUDA(at::kFloat).rand({N, C, H, W});
+  at::Tensor W1 = at::CUDA(at::kFloat).rand({O, KC, KH, KW});
+  std::vector<at::Tensor> inputs = {I, W1};
+  std::vector<at::Tensor> outputs;
+
+  Check(
+      R"(
+def local_sparse_convolution(float(N, C, H, W) I, float(O, KC, KH, KW) W1) -> (O) {
+    O(n, o, h, w) +=! I(n, (c + kc) % c, h + kh, w + kw) * W1(o, kc, kh, kw) where c in 0:C
```
abadams: Isn't (c + kc) % c  equal to (kc % c) ?

abadams:
This all makes sense to me so far.

ftynse:
When you get to Halide -> isl conversion, there may be a need to differentiate between right hand sides of `%` that appear in index expressions.  If those are compile-time constants, we can express them as quasi-affine expressions and have exact polyhedral analyses.  Otherwise, we need to overapproximate like we do for indirections.

prigoyal:
thanks @ftynse . I looked further and tried to understand the `tc2halide` and `halide2isl` better. It looks like the segfault happens because of infinite recursion in https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/core/halide2isl.cc#L203 . I fixed that bug and test seems to pass now :)

tc/core/halide2isl.cc
```diff
@@ -35,13 +35,15 @@ using namespace tc::polyhedral::detail;
...
-  // Collect and categorize all the Variable symbols
+  // Collect and categorize all the Halide Variable symbols as reduction
+  // or index variables
   class BuildSymbolTable : public IRVisitor {
     using IRVisitor::visit;
     std::set<std::string> included;
     void visit(const Variable* op) {
       if (!included.count(op->name)) {
         if (op->param.defined()) {
+          // why is this empty? - what's the point of this?
```
prigoyal: I didn't quite understand why this is empty and added a comment. if one of the reviewers can explain it, that'd be great and I can remove this comment as well

tc/core/halide2isl.cc
```diff
@@ -197,10 +199,9 @@ std::vector<isl::aff> makeIslAffBoundsFromExpr(
   } else if (const Div* op = e.as<Div>()) {
     return combineSingleAffs(space, op, &isl::aff::div);
   } else if (const Mod* op = e.as<Mod>()) {
-    std::vector<isl::aff> result;
     // We cannot span multiple constraints if a modulo operation is involved.
     // x > max(a,b) % C is not equivalent to (x > a % C && x > b % C).
-    auto lhs = makeIslAffBoundsFromExpr(space, e, false, false);
+    auto lhs = makeIslAffBoundsFromExpr(space, op->a, false, false);
```
prigoyal: using `e` would lead to infinite recursion and hence segfault

tc/core/tc2halide.cc
```diff
@@ -62,43 +62,45 @@ Type translateScalarType(int tcType) {
...
-        Parameter p;
-        if (it != params->end()) {
-          p = it->second;
-        } else {
-          p = Parameter(Int(32), false, 0, d.name(), true);
-          (*params)[d.name()] = p;
-        }
-        dims.push_back(Variable::make(Int(32), p.name(), p));
+  }
+  lang::TensorType type = p.tensorType();
```
prigoyal: there is no need for `else` here since we return from the `if` condition if satisfied.

tc/core/tc2halide.cc
```diff
@@ -492,20 +496,25 @@ Expr reductionUpdate(Expr e) {
   return Call::make(e.type(), kReductionUpdate, {e}, Call::Intrinsic);
 }
 
+// translate a single TC comprehension/statement to Halide component.
+// funcs, bounds, reductions will be populated
 void translateComprehension(
-    const lang::Comprehension& c,
+    const lang::Comprehension& comprehension,
```
prigoyal: while reading the code, it was confusing to me to read `c` so I made it more descriptive to facilitate code readability. Hope this is okay, otherwise I can revert too.

tc/lang/sema.h
```diff
@@ -188,23 +191,27 @@ struct Sema {
     }
     return matched_type;
   }
+
```
prigoyal: I was repeatedly getting confused about where a function started and ended. So I added these white lines to help. is this okay @zdevito ? :)

tc/lang/sema.h
```diff
@@ -366,21 +395,27 @@ struct Sema {
...
+    return createCompound(TK_INT32, anchor->range(), {});
   }
+
   TreeRef dimType(TreeRef anchor) {
     return indexType(anchor);
   }
+
   TreeRef floatType(TreeRef anchor) {
-    return c(TK_FLOAT, anchor->range(), {});
+    return createCompound(TK_FLOAT, anchor->range(), {});
```
prigoyal: simple `c` was confusing to me as a function name.

test/cuda/test_execution_engine.cc
```diff
@@ -235,6 +235,24 @@ def convolution(float(N,C,H,W) I, float(O,C,KH,KW) W1, float(O) B)
...
+TEST_F(ATenCompilationUnitTest, LocalSparseConvolution) {
+  at::Tensor I = at::CUDA(at::kFloat).rand({N, C, H, W});
+  at::Tensor W1 = at::CUDA(at::kFloat).rand({O, KC, KH, KW});
+  std::vector<at::Tensor> inputs = {I, W1};
+  std::vector<at::Tensor> outputs;
+
+  Check(
+      R"(
+def local_sparse_convolution(float(N, C, H, W) I, float(O, KC, KH, KW) W1) -> (O1) {
+    O1(n, o, h, w) +=! I(n, kc % c, h + kh, w + kw) * W1(o, kc, kh, kw) where c in 0:C
```
nicolasvasilache: This cannot be correct, it will introduce ```kc % 0``` operations.

test/cuda/test_execution_engine.cc
```diff
@@ -235,6 +235,24 @@ def convolution(float(N,C,H,W) I, float(O,C,KH,KW) W1, float(O) B)
...
+}
+    )",
+      "local_sparse_convolution",
+      tc::CudaMappingOptions::makeConvolutionCudaMappingOptions(),
+      inputs,
+      outputs);
+}
```
nicolasvasilache: Please add an actual check function (even if you have to write it using plain C loops).

tc/lang/sema.h
```diff
@@ -627,10 +677,12 @@ struct Sema {
     }
     return it == the_env.end() ? nullptr : it->second;
   }
-  TreeRef c(int kind, const SourceRange& range, TreeList&& trees) {
+
+  TreeRef createCompound(int kind, const SourceRange& range, TreeList&& trees) {
```
nicolasvasilache: We have been using ```makeXYZ``` syntax throughout the codebase for functions that return an owning pointer, let's try to be consistent.

tc/lang/sema.h
```diff
@@ -627,10 +677,12 @@ struct Sema {
...
     return Compound::create(kind, range, std::move(trees));
   }
-  TreeRef s(const std::string& s) {
+
+  TreeRef createString(const std::string& s) {
```
nicolasvasilache: We have been using ```makeXYZ``` syntax throughout the codebase for functions that return an owning pointer, let's try to be consistent.

tc/lang/sema.h
prigoyal: it was named `create...` because that's how the functions are named in lexer/sema/parser codebade to create a new node like: https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/lang/tree_views.h#L204. The coding conventions recommend to have consistency locally within lexer/sema/parser (for example) code base here https://github.com/facebookresearch/TensorComprehensions/blob/master/CodingConventions.md#naming

if the naming should be `make...`, 1. can you kindly create a rule in `CodingConventions.md` so that people can follow that? 2. it would be also nice to define when should global/local rules be followed? thanks :)

tc/lang/sema.h
prigoyal: [copy-paste from above]:

it was named `create...` because that's how the functions are named in lexer/sema/parser codebade to create a new node like: https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/lang/tree_views.h#L204. The coding conventions recommend to have consistency locally within lexer/sema/parser (for example) code base here https://github.com/facebookresearch/TensorComprehensions/blob/master/CodingConventions.md#naming

if the naming should be `make...`, 1. can you kindly create a rule in `CodingConventions.md` so that people can follow that? 2. it would be also nice to define when should global/local rules be followed? thanks :)

tc/lang/sema.h
nicolasvasilache: sounds good, local trump global :)

tc/lang/sema.h
nicolasvasilache: sounds good, local trumps global :)

tc/core/tc2halide.cc
```diff
@@ -71,37 +71,36 @@ void translateParam(
...
+  vector<Expr> dims;
+  for (auto d_ : type.dims()) {
+    if (d_->kind() == lang::TK_IDENT) {
+      auto d = lang::Ident(d_);
+      auto it = params->find(d.name());
+      Parameter p;
+      if (it != params->end()) {
+        p = it->second;
       } else {
-        CHECK(d_->kind() == lang::TK_CONST);
```
nicolasvasilache: This hoisting looks like a behavior change to me, could you please explain why it is necessary or correlated with this commit message?

prigoyal: thanks for the review. I am not sure I understand. There is no code change to this section except removing unnecessary else after line 74. 

line 88 - 98 in the green section matches line 85 - 95 in red section. I think that github alignment for code changes is not exactly good and hence causing confusion.

tc/core/halide2isl.cc
```diff
@@ -93,7 +94,7 @@ namespace {
  * Convert Halide binary expression "op" into a list of isl affine functions by
  * converting its LHS and RHS into lists of affs and concatenating those lists.
  * This is intended to be used with Min/Max operations in upper/lower bound
- * computations, respectively.  Essentially, this allows for replacements
```
ftynse: I consistently use double spaces between sentences.  Makes it easy to read longer text fragments typeset in monospace fonts.  You may have noticed in my commit messages, or other comments, including this one.  I intend to keep doing so unless the code guidelines require the opposite.

nicolasvasilache:
So it seems the unit test change requested was addressed by simply deleting the unit test. 
As it stands, this PR adds % to parsing and checks that the operation is indeed performed on an int32 value which is good.

However as far as I see, it does not address #290 which requires support for ***modulus array indexing***. So I would remove the  ```halide -> isl``` part from your commit message until this is addressed and a proper test is added.

Could you elaborate on what happens if you try to compile a TC with a modulus array indexing? I assume the parsing now passes but I would expect some undefined behavior when reaching ISL. 

Does the TC mapper complain in an actionable way, just segfault/crash, produce some wrong code or something else?

tc/core/tc2halide.cc
ftynse: diff computation alogrithm is not perfect; reviewing changes per-commit helps

tc/core/tc2halide.cc
ftynse: +1, one character names are for iteration variables of short loops

tc/core/tc2halide.cc
nicolasvasilache: Ah yes, you're right, I see there are 2 columns of numbers. I learned something today! Sorry for the noise here

tc/lang/sema.h
```diff
@@ -447,11 +489,13 @@ struct Sema {
       insert(index_env, index, typ, true);
     }
 
-    // make dimension variables for each dimension of the output tensor
+    // check that the input is not used for output - input are immutable
```
ftynse: "inputs are" or "input is"

ftynse:
After a brief discussion, it looks like we all want three distinct test cases:
- modulus with constant RHS used in the subscript (`A(i) = B(i % 42)`), make sure _exact_ access relations are computed;
- modulus with variable RHS used in the subscript (`A(i) +=! B(i % j) where j in 1:3`), make sure access relations are non-empty;
- modulus with variable RHS used in the outermost expression (`A(i) = B(i) % C(0)`), check that the result is correct.
The difference in only important when modulus appears in the subscript, where we have a special case for constant RHS.  In the outermost expression, we just check that the most general case works.

Bonus points for adding a warning in the two latter cases that the precondtion (`j != 0` and `C(0) != 0` will not be checked in the runtime)

prigoyal:
thanks all for the review. The unit test in this PR so far is meant to address the "correctness" that % works with a simplest example. The breadth of cases wasn't explored in that unit test. Modulus array indexing also works as it was in earlier version of commit but I don't have a way to write correctness when involving convolution type operations. I'll think of a simple test case to demonstrate that and I welcome any ideas there.

thanks @ftynse for a clear description :) I'll add those unit tests as well.

nicolasvasilache:
> correctness when involving convolution type operations
@prigoyal you could have a simple nested loop that operates on the CPU. It is probably overkill for this PR though our friend in #290 may use it.

> I'll think of a simple test case to demonstrate that and I welcome any ideas there.
You could for instance try an integer array rotation. i.e. [1, 2, 3] -> [3, 1, 2] that should be enough to exercise the ```A(i) = B(i % 42)```, you already have the ```A(i) = B(i) % C(i)``` case. For the ```A(i) +=! B(i % j) where j in 1:3``` case it's a bit more tricky, what @ftynse  suggested (test non-emptyness of access relations) sounds reasonable.

ftynse:
For the first two cases, you could do something similar to what I did for non-affine subscripts [here](https://github.com/facebookresearch/TensorComprehensions/pull/378/files#diff-30ac823bfce2e27b8c04e70f8dba0047R717).  That lets you run do the TC->Halide->Scop conversion with all the parsing and inference involved, but does not require anything runnable.  Given a Scop, you can look at the read/write relations (public member variables).  In the first case, I expect the range of the read relation to have a stride.  It is possible to extract the stride of the isl set using an appropriate isl function (`isl_set_get_stride`). In the second case with the snippet I gave, I expect the range of the read relation to be either a universe or restricted to the extent of the tensor `B`.

For the third case, if that can be expressed as an ATen expression, just have it in `test/cuda/test_tc_mapper`.  Tests do not have to correspond to actual use cases.

tc/core/halide2isl.cc
skimo-openhub: Me too.  In any case, this change does not belong to a commit that adds documentation.

tc/core/tc2halide.cc
```diff
@@ -62,10 +62,13 @@ Type translateScalarType(int tcType) {
   }
 }
 
+// translate the TC def input params to corresponding Halide components.
+// params, inputs will be populated here
```
skimo-openhub: Please start sentences with a capital and close them with a period.
This also applies to other comments added below.

tc/core/tc2halide.cc
```diff
@@ -62,10 +62,13 @@ Type translateScalarType(int tcType) {
...
 void translateParam(
     const lang::Param& p,
     map<string, Parameter>* params,
     vector<ImageParam>* inputs) {
+  // check if the param is already converted to halide components
```
skimo-openhub: Hmm... do you mean "has already been converted"?

tc/core/tc2halide.cc
```diff
@@ -492,6 +495,8 @@ Expr reductionUpdate(Expr e) {
   return Call::make(e.type(), kReductionUpdate, {e}, Call::Intrinsic);
 }
 
+// translate a single TC comprehension/statement to Halide component.
```
skimo-openhub: _a_ Halide component.

tc/core/tc2halide.cc
```diff
@@ -492,6 +495,8 @@ Expr reductionUpdate(Expr e) {
...
+// funcs, bounds, reductions will be populated
```
skimo-openhub: Since you are documenting this function, you might as well explain what
those are and how they relate to the "Halide component".

tc/lang/sema.h
```diff
@@ -339,6 +339,21 @@ struct Sema {
...
+  // sure that the tree is sematically correct. For example: a variable
+  // may not be input two times. Parser only verifies for the syntax but does
+  // not check the semantics.
+  //
+  // It converts the TK_APPLY nodes to TK_ACCESS or TK_BUILT_IN
+  //
+  // The reduction variables are deduced and the objects are created for them
+  // and they are appended to the tree
+  //
+  // Type checking is also done but by a very small codebase
```
skimo-openhub: What does "but by a very small codebase" mean?
Does it mean that the type checking is done using a small amount of code?
If so, then why the "but"?

tc/lang/sema.h
```diff
@@ -339,6 +339,21 @@ struct Sema {
...
+  //
+  // withType - associates the type with a given node
```
skimo-openhub: What's the connection with `checkFunction`?
That is, why did you add this here?

tc/lang/sema.h
skimo-openhub: `createString` doesn't appear to be used anywhere.
It may be better to simply remove it (in a separate commit).

tc/core/halide2isl.cc
skimo-openhub: This comment should be in the commit message.

test/cuda/test_corner_cases.cc
```diff
@@ -285,6 +285,17 @@ TEST(TestCornerCases, E23) {
...
 
+TEST(TestCornerCases, E24) {
+  auto a = I();
+  auto b = I();
+  auto r = I(0);
+  Succeed(
+      "def f(int32 a, int32 b) -> (c) { c(i) = int32(a % b) where i in 0:1 }",
+      {a, b},
+      {r});
+  CHECK_EQ(at::Scalar(r[0]).toInt(), 0);
```
skimo-openhub: I understand that this file is only meant to check compilation, but shouldn't also add a test
somewhere that checks that the correct operation is being performed?

tc/lang/sema.h
prigoyal: checkFunction is the main entry point for Sema codebase called by `tc2halide`

I added `withType` because this is important component of sema where each node gets it's proper type. I wanted to highlight that it's an important functionality as part of what `checkFunction` will result it.

test/cuda/test_corner_cases.cc
prigoyal: this is the correctness check as well. I can modify it a bit so it actually is visible as correctness test. Currently, I did mental calculation on inputs and added the expected output as 0 but that's not the safe way, I'll change this :)

tc/lang/sema.h
prigoyal: yep, there's no need for "but". I'll remove it :)

test/test_cuda_mapper.cc
```diff
@@ -753,6 +753,26 @@ def perforatedConvolution(float(N, C, H, W) input, float(M, C, KH, KW) weights,
   Prepare(tc);
 }
 
+TEST_F(PolyhedralMapperTest, ModulusConstantRHS) {
+  string tc = R"TC(
+def fun(float(N) a) -> (b) { b(i) = a(i % 2) where i in 0:N }
+)TC";
+  // This triggers tc2halide conversion and should not throw.
+  auto scop = Prepare(tc);
+  auto reads = scop->reads;
```
prigoyal: hey @ftynse , I have extracted out the reads here but I don't know how to get the isl stride information. I tried to search in the codebase but couldn't find much help. Can you give some pointers? thanks.

ftynse: Yeah, you probably want to search in the third-party/islpp as well :)  There is a greppable documentation file in `doc/`.

Something along the lines of
```c++
for (auto r : scop->reads.range()) {
  if (r.get_tuple_name() != std::string("B")) { // skip irrelevant reads, if any
    continue;
  }
  EXPECT_EQ(r.get_stride(), putExpectedStrideValueHere);
  // EXPECT_TRUE(r.plain_is_universe()); // for the the set being universe
}
```
should work

prigoyal:
the % operator existed in halide2isl but it was segfaulting. 52989ea69bdb93fca1813e5d885a8d9977637eda commit fixed that :)

test/test_cuda_mapper.cc
prigoyal: thanks. I tried the above snippet but getting compilation failure on `for (auto r : scop->reads.range())` it seems like it's not iterable. 

is the https://github.com/simbuerg/isl/blob/master/doc/user.pod correct file to consult?

ftynse: Sorry, my bad, try `scop->reads.range().get_set_list()` instead.

This one https://github.com/nicolasvasilache/isl/blob/ntv_dev/doc/user.pod comes from what we submodule (our copy of isl is ahead even of isl's master)

skimo-openhub:
On Wed, May 02, 2018 at 09:07:31AM -0700, Priya Goyal wrote:
> prigoyal commented on this pull request.
> > @@ -339,6 +339,21 @@ struct Sema {
>          throw ErrorReport(exp) << "NYI - semantic checking for " << exp;
>      }
>    }
> +  // This is the entry function for semantic analysis. It is called by
> +  // tc2halide to associate type with each node of the tree and to also make
> +  // sure that the tree is sematically correct. For example: a variable
> +  // may not be input two times. Parser only verifies for the syntax but does
> +  // not check the semantics.
> +  //
> +  // It converts the TK_APPLY nodes to TK_ACCESS or TK_BUILT_IN
> +  //
> +  // The reduction variables are deduced and the objects are created for them
> +  // and they are appended to the tree
> +  //
> +  // Type checking is also done but by a very small codebase
> +  //
> +  // withType - associates the type with a given node
> 
> checkFunction is the main entry point for Sema codebase called by `tc2halide`
> 
> I added `withType` because this is important component of sema where each node gets it's proper type. I wanted to highlight that it's an important functionality as part of what `checkFunction` will result it.

Hmm... now it just says "Associates the type with a given node".
Is this what checkFunction does or what withType does?
You added that to the documentation of checkFunction.

If you want to refer to some method from inside the documenation
of another method, then please be a bit more explicit about it.
"The method withType can be used to ..."
Just reading this "withType -", I really had no idea what it was about.

skimo

test/test_cuda_mapper.cc
```diff
@@ -753,6 +753,38 @@ def perforatedConvolution(float(N, C, H, W) input, float(M, C, KH, KW) weights,
...
+def fun(float(N) a) -> (b) { b(i) = a(i % 2) where i in 0:N }
+)TC";
+  // This triggers tc2halide conversion and should not throw.
+  auto scop = Prepare(tc);
+  for (auto r : scop->reads.range().get_set_list()) {
+    // skip irrelevant reads, if any
+    if (r.get_tuple_name() != std::string("a")) {
+      continue;
+    }
+    // EXPECT_EQ(r.get_stride(?), ?);
```
prigoyal: I am not quite sure what should be the `pos` for `get_stride()`.

ftynse: index of the dimension in the set (position of the subscript that should involve the stride), `0` in this case a believe

test/test_cuda_mapper.cc
```diff
@@ -753,6 +753,38 @@ def perforatedConvolution(float(N, C, H, W) input, float(M, C, KH, KW) weights,
...
+def fun(float(N) a) -> (b) { b(i) = a(i % 3) where i in 0:N }
+)TC";
+  // This triggers tc2halide conversion and should not throw.
+  auto scop = Prepare(tc);
+  for (auto r : scop->reads.range().get_set_list()) {
+    // skip irrelevant reads, if any
+    if (r.get_tuple_name() != std::string("a")) {
+      continue;
+    }
+    // EXPECT_EQ(r.get_stride(0), ?);
```
ftynse: this test is still commented out...

skimo-openhub:
On Fri, May 04, 2018 at 10:22:09PM +0000, ftynse wrote:
> > @@ -753,6 +753,38 @@ def perforatedConvolution(float(N, C, H, W) input, float(M, C, KH, KW) weights,
>    Prepare(tc);
>  }
>  
> +TEST_F(PolyhedralMapperTest, ModulusConstantRHS) {
> +  string tc = R"TC(
> +def fun(float(N) a) -> (b) { b(i) = a(i % 3) where i in 0:N }
> +)TC";
> +  // This triggers tc2halide conversion and should not throw.
> +  auto scop = Prepare(tc);
> +  for (auto r : scop->reads.range().get_set_list()) {
> 
> Sorry, I misled you here; taking the range of the access relation here will actually drop the stride because the stride is on the input dimension.  Instead, let's use _wrapping_.  It is a transformation that reshapes the space from a relation (`[i]->a[a0]`) to a set _wrapping a relation_ (`[[i]->a[a0]]`, note the extra surrounding `[]` that mean it's a set now).  In that set, we can take the stride an compare it with the expected value.  However, we cannot directly check the ex-range's name, so we need to unwrap the set and take its range before using `get_tuple_name()`.

Why do you see this as a stride check?
Why don't you simply compare the access relation to the expected
access relation (perhaps after projecting out the reference tag)?

skimo

test/test_cuda_mapper.cc
```diff
@@ -753,6 +753,38 @@ def perforatedConvolution(float(N, C, H, W) input, float(M, C, KH, KW) weights,
   Prepare(tc);
 }
 
+TEST_F(PolyhedralMapperTest, ModulusConstantRHS) {
+  string tc = R"TC(
+def fun(float(N) a) -> (b) { b(i) = a(i % 3) where i in 0:N }
+)TC";
+  // This triggers tc2halide conversion and should not throw.
+  auto scop = Prepare(tc);
+  for (auto r : scop->reads.range().get_set_list()) {
```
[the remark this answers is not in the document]
prigoyal: thanks @ftynse . can you provide guidance on skimo's comment ? I am not sure myself which is the right approach. Thanks a lot :)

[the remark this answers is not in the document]
ftynse: It's a matter of preference, I think.  @skimo-openhub suggests to construct a new relation from its textual description, basically `isl::map(ctx, "<paste here the textual representation of the map litearlly>")` (taking `ctx` from whatever isl object available around) and comparing that to the result.  It would work, but I prefer testing only the relevant properties of the relation, hence the stride check.  Checking only relevant properties removes the need to modify the test when irrelevant properties change.  I will accept both ways of checking.

More specifically, current access relations do not involve bounds imposed by tensor sizes.  I'm quite annoyed at this during memory promotion and was planning to add those bounds in access relation construction.  That would change these relations, requiring to update this test that is affected only indirectly.

skimo-openhub:
On Tue, May 08, 2018 at 01:30:16PM -0700, ftynse wrote:
> > @@ -753,6 +753,38 @@ def perforatedConvolution(float(N, C, H, W) input, float(M, C, KH, KW) weights,
>    Prepare(tc);
>  }
>  
> +TEST_F(PolyhedralMapperTest, ModulusConstantRHS) {
> +  string tc = R"TC(
> +def fun(float(N) a) -> (b) { b(i) = a(i % 3) where i in 0:N }
> +)TC";
> +  // This triggers tc2halide conversion and should not throw.
> +  auto scop = Prepare(tc);
> +  for (auto r : scop->reads.range().get_set_list()) {
> 
> It's a matter of preference, I think.  @skimo-openhub suggests to construct a new relation from its textual description, basically `isl::map(ctx, "<paste here the textual representation of the map litearlly>")` (taking `ctx` from whatever isl object available around) and comparing that to the result.  It would work, but I prefer testing only the relevant properties of the relation, hence the stride check.  Checking only relevant properties removes the need to modify the test when irrelevant properties change.  I will accept both ways of checking.

Of course you should only check for relevant properties,
but the stride is only a part of the property you want to check.
In any case, don't do any wrapping.  We can easily
add a isl_map_get_range_stride based on isl_map_get_range_stride_info
in the same way isl_set_get_stride is based on
isl_set_get_stride_info and then use isl_map_get_range_stride.

> More specifically, current access relations do not involve bounds imposed by tensor sizes.  I'm quite annoyed at this during memory promotion and was planning to add those bounds in access relation construction.  That would change these relations, requiring to update this test that is affected only indirectly.

You shouldn't check that is _equal_ to some map,
but a _subset_ of the expected property
(and perhaps also that it is non-empty).

skimo

ftynse:
rebased and revived

Issue #338, opened 2018-04-24
temporarily remove GPU CI results from PRs


ftynse:
It has been some time since GPU CI does not actually run tests.  As a result, we cannot easily differentiate the PRs that are failing both CPU and GPU and those only failing GPU.  It also makes the PR history on Github look bad.

Could we prevent failing GPU CI reports from showing up in the PRs?  It may be switched back once it is operational again.

prigoyal:
thanks for opening up the issue. We are actively working on resolving this and currently configuring dedicated gpu machines for TC. Jenkins CI has also been unstable these days which is adding to the delay.

There is no good way per-se to disable the CI at the same time trying to debug it. Please use CircleCI tests only for the moment for PRs. I will update this issue once the CI is functional. Thanks.

skimo-openhub:
GPU CIs appear to be working again.

PR #349, opened 2018-04-25
Activate conda env which has TC build for test time

prigoyal:
we have dedicated 4 gpu machines finally configured and running now. In attempt to revive jenkins CI, found this fix. The reason is in commit message. 

partially helps #338


prigoyal:
thanks @ftynse. There is a docker trigger phase on Jenkins CI which firsts builds the docker images for all the 4 build env -> build TC in all 4 env -> test TC in 4 env -> if all is good, the image is finalized and usable for rest of the builds. 

Current status is that 3 build/test env should work. One env is failing to build aten which I am debugging right now. We are getting closer. :)

PR #353, opened 2018-04-25
[C++ API] Step 1: drop manual and compilation caches

nicolasvasilache:
This is the first PR towards implementing the API proposed in #307.
As part of refactoring the OptionsCache we realized that compilation caches may
have been quite useful in the beginning but are a premature optimization in the
new API.

Removing them has multiple benefits such as:
1. simplify the codebase CRTP + inheritance + custom traversals and
serialization are not fun to maintain
2. simplify porting to the new API where ExecutionEngine is dead
3. caches seem to add overhead in the context of cooperative groups
4. in the new API we can cache executors which removes the need to rely on
compilation caches
5. tuning and options caches is the expensive bit that needs to be memoized

This seems corroborated by the [discussion](https://github.com/facebookresearch/TensorComprehensions/pull/339#discussion_r183937583) in #339.

Note that some tests in test_tc_mapper need to be temporarily degraded because
we do not have a simple way to extract generated code before the new API.

cc @prigoyal for the python-specific things, I grepped but have not yet run the python tests myself


ftynse:
Stylistic comment: can we use plain test in git commit messages instead of markdown?

test/cuda/test_tc_mapper.cc
```diff
@@ -122,19 +117,7 @@ struct TcCudaMapper2DReductionStressTest : public TcCudaMapper2DReductionTest {
...
-          std::to_string(i) + std::string(",1,1,__tc::ReductionOp::Sum>");
-      ASSERT_NE(std::string::npos, res.second.find(expected))
-          << "In resulting code:\n"
-          << res.second << "\ncould not find: " << expected;
-    } else {
-      std::string expected = "__tc::CubReduceAlongX<";
-      ASSERT_EQ(std::string::npos, res.second.find(expected))
-          << "In resulting code:\n"
-          << res.second << "\nfound unexpected: " << expected;
-    }
```
skimo-openhub: I like how this commit removes a lot of code, but this bit seems like
it is not directly related to the caching and looks like it would be nice to keep.
Any chance of giving callers of Check access to the ExecutorType such that
this caller can extract the `cudaSource` from the `CudaTcExecutor`?

nicolasvasilache: Yes but it is happening in the larger refactoring in branch [pr/big-refactor](https://github.com/nicolasvasilache/TensorComprehensions/blob/pr/big-refactor/test/test_tc_mapper_harness-inl.h#L54)

The code says:
```
    return std::make_pair(
        std::move(outputs), std::move(pExecutor->compiledSource));
```

It seems like wasteful extra work to locally keep this behavior?
@skimo-openhub if you feel strongly that this temporary state of affairs is not good I can put in the effort.

OTOH, let this be an opportunity to freeze the codebase for pushing the refactoring?

nicolasvasilache:
@ftynse are there particular features that are annoying?
The backticks to delimit and highlight code don't feel like they would be a problem, the enumeration either.

I would think maybe the hyperlinks would be but then how do you cross reference code? 
Copy-paste it directly?

skimo-openhub:
On Thu, Apr 26, 2018 at 02:13:00PM +0000, Nicolas Vasilache wrote:
> Yes but it is happening in the larger refactoring in branch [pr/big-refactor](https://github.com/nicolasvasilache/TensorComprehensions/blob/pr/big-refactor/test/test_tc_mapper_harness-inl.h#L54)
> 
> The code says:
> ```
>     return std::make_pair(
>         std::move(outputs), std::move(pExecutor->compiledSource));
> ```
> 
> It seems like wasteful extra work to locally keep this behavior?
> @skimo-openhub if you feel strongly that this temporary state of affairs is not good I can put in the effort.

I'll leave it to you to judge.
Since it only affects the test and you explain
that it's cominb back in the commit message,
I guess it shouldn't be too bad.

skimo

nicolasvasilache:
Cool thanks for your reviews, let's roll with this then, I'll look at python in the context of #307 once I am able to build from source. Until then CI is still down.

ftynse:
> I would think maybe the hyperlinks would be but then how do you cross reference code?
Copy-paste it directly?

I don't see a point in cross-referencing code.  You can mention a specific function by name, but anything else is too volatile to stay in history.  Especially links like `github/repo/master/blob/file.cc#L312` because `master` will change and even individual hashes may change due to rebasing.

If we want all history and motivation so stay in git logs, it should be just pasted there.

PR #358, opened 2018-04-26
Use gcc 5.4 on xenial non-conda + use pytorch 0.4 official + fix broken python API

prigoyal:
Fix docker image to use gcc 5.4 on xenial non-conda env to fix aten build which fails with avx related issues on gcc5.5. Ubuntu toolchain ships gcc 5.5 be default since April 24, 2018 hence not using the toolchain for gcc 5.4

issue filed on pytorch repo https://github.com/pytorch/pytorch/issues/6992 
i'll propagate this to gcc as well

partially helps #338

### DIFF
--- a/docker/common/install_base.sh
+++ b/docker/common/install_base.sh
@@ -34,11 +34,15 @@ apt-get install -y --no-install-recommends \
 apt-get clean
 rm -rf /var/lib/apt/lists/*
 # setup gcc
-add-apt-repository ppa:ubuntu-toolchain-r/test
-apt-get update
-apt-get install -y --no-install-recommends libcilkrts5 gcc-$GCC_VERSION g++-$GCC_VERSION
-update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-$GCC_VERSION 50
-update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-$GCC_VERSION 50
+if [[ "$GCC_VERSION" == 4.9 ]]; then
+  add-apt-repository ppa:ubuntu-toolchain-r/test
+  apt-get update
+  apt-get install -y --no-install-recommends libcilkrts5 gcc-$GCC_VERSION g++-$GCC_VERSION
+  update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-$GCC_VERSION 50
+  update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-$GCC_VERSION 50
+else
+  apt-get install -y --no-install-recommends libcilkrts5 gcc g++
+fi
 
 # Install ccache from source. Needs 3.4 or later for ccbin support
 # Needs specific branch to work with nvcc (ccache/ccache#145)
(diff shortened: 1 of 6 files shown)


docker/common/install_base.sh
```diff
@@ -34,11 +34,15 @@ apt-get install -y --no-install-recommends \
 apt-get clean
 rm -rf /var/lib/apt/lists/*
 # setup gcc
-add-apt-repository ppa:ubuntu-toolchain-r/test
-apt-get update
-apt-get install -y --no-install-recommends libcilkrts5 gcc-$GCC_VERSION g++-$GCC_VERSION
-update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-$GCC_VERSION 50
-update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-$GCC_VERSION 50
+if [[ "$GCC_VERSION" == 4.9 ]]; then
```
ftynse: SO this installs from PPA when asked for GCC 4.9, but what is the version shipped with the system?

skimo-openhub:
On Thu, Apr 26, 2018 at 03:13:33PM +0000, ftynse wrote:
> 
> Could you please add an empty line in your commit message between "Fix docker image to use gcc 5.4 on xenial non-conda env" and the rest?  In git, the first "paragraph" (before the empty line) is the short summary and the rest is the longer commit message that are treated differently.  We generally prefer short summaries to fit into 80 characters.

Not just the one-line summary, but the entire commit message
should fit in 80 characters.
See
https://github.com/facebookresearch/TensorComprehensions/blob/master/CodingConventions.md#commits

skimo

docker/common/install_base.sh
prigoyal: system version 4.8.4

ftynse: then I don't really understand how is this supposed to make it use gcc-5.4 as the commit message says

prigoyal: for the trusty distrib, we use gcc4.9 but default shipped is 4.8.4
for xenial, we use gcc 5.4 but default shipped by ubuntu toolchain is 5.5 as of April 24

ftynse: so how do we get gcc 5.4 then?  we only install from ppa if 4.9 is requested.  so when 5.4 is requested -- nothing happens.  I assume than the default 5.5 is used, which sort of contradicts your commit message.

prigoyal: if gcc version 4.9 is requested, we get it from ppa. otherwise, don't use the toolchain and apt install it. Earlier, gcc5 was coming from ppa as well which used to be gcc5.4 before April 24

ftynse: I guess the part that is missing from the description is that xenial has 5.4 _by default_

prigoyal: got it. I understand now what you mean. I'll edit the description. :)

ftynse: thanks! sorry it took so long

prigoyal:
thanks @ftynse for review, I just edited the description and also fixed the python API broken by a recent change. this PR should be good to go now and ready for review again :)

prigoyal:
merging this to help facilitate fixing the Jenkins CI

@mingzhe09088 can you re-try the python tests again with this PR? I noticed the similar issue as you reported. It was broken in a recent change in PR 353.

Issue #351, opened 2018-04-25
Implement CHECK macros as exceptions


nicolasvasilache:
At the moment we have too many places where we call one of the glog CHECK macros and terminate when it fails.
Let's implement our own CHECK macros and recover when it makes sense.
@salexspb @Yangqing I know C2 has its way of dealing with this, could you offer some guidance on what would make the most sense wrt FB infra?

Assigning to @ttheodor since we discussed about this last month.

nicolasvasilache:
Related to [this](https://github.com/facebookresearch/TensorComprehensions/pull/327#discussion_r184203685).

ftynse:
Related to #193

PR #359, opened 2018-04-26
[C++ API] Step 2: add the new compiler files

nicolasvasilache:
This PR adds the files implementing the new compiler API #322. This allows
retiring the ExecutionEngine #309 which has been a significant source of
unnecessary complexity.

More specifically, this implements a more functional-style API for
compilation. ```compile``` returns an owning pointer to a TcExecutor
on which ```run``` and ```uncheckedRun``` can be called.

This also implements support for ```parse``` and ```inferOutputTensorInfo```
as well as operations on the parsed ```TreeRef``` in the ```detail`` namespace.

Higher-level layers are responsible for caching compilation results.

the global refactoring.

### DIFF
--- a/tc/core/compiler-inl.h
+++ b/tc/core/compiler-inl.h
@@ -1,0 +1,71 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "tc/core/compiler.h"
+
+#include <sstream>
+#include <string>
+
+#include "tc/core/flags.h"
+#include "tc/core/halide_utils.h"
+#include "tc/core/tensor.h"
+#include "tc/lang/canonicalize.h"
+
+namespace tc {
+namespace detail {
+void checkInputsCompliant(
+    const tc2halide::HalideComponents& halideComponents,
+    const std::vector<const DLConstTensor*>& inputsInfo);
+} // namespace detail
+
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const typename Backend::MappingOptionsType& options) {
+  auto parsedTcs = detail::parse(tc);
+  CHECK_EQ(parsedTcs.count(entryPoint), 1u)
+      << "attempting to access undefined function " << entryPoint;
+  return compile<Backend>(parsedTcs[entryPoint], inputs, options);
+}
+
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    lang::TreeRef tcDefinition,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const typename Backend::MappingOptionsType& options) {
+  using CompilationResultType = typename Backend::CompilationResultType;
+
+  auto inputsInfo = makeTensorInfoVector(inputs);
+  auto outputsInfo = detail::inferOutputTensorInfo(tcDefinition, inputs);
+  auto halideComponents =
+      tc2halide::translate(isl::with_exceptions::globalIslCtx(), tcDefinition);
+  detail::checkInputsCompliant(halideComponents, inputs);
+
+  auto tcName = lang::Def(tcDefinition).name().name();
+  CompilationResultType compilationResult = Backend::compileWithTcMapper(
+      tcName,
+      halideComponents,
+      inputs,
+      /* TODO outputs, */
+      options);
+  return std::unique_ptr<typename Backend::ExecutorType>(
+      new typename Backend::ExecutorType(
+          inputsInfo, outputsInfo, halideComponents, compilationResult));
+}
+} // namespace tc
--- a/tc/core/compiler.h
+++ b/tc/core/compiler.h
@@ -1,0 +1,123 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <map>
+#include <string>
+#include <vector>
+
+#include "tc/core/mapping_options.h"
+#include "tc/core/tensor.h"
+#include "tc/lang/tree.h"
+
+/**
+ * This provides a simple functional-style C++ API with multi-backend
+ * capabilities to:
+ *   1. compile a TC function and return an Executor for the specified Backend
+ *      on which the run method can be called;
+ *   2. infer actual tmp/output tensor shapes given input tensor shapes;
+ *   3. parse a TC definition and retrieve the map of TC function to parsed TC
+ *      trees.
+ *
+ * Compilation is backed by a compilation cache, its correspondance is:
+ * 1 TcExecutor <-> 1 compiled tuple<TC function, input shapes, MappingOptions>
+ *
+ * The compile function is templated by the Backend type.
+ * For each backend, the specific Backend type lives in
+ *   backendname/backendname_backend.h and declares all the required dependent
+ *   **derived** types.
+ * For example:
+ *   CudaBackend is declared in core/cuda/cuda_backend.h
+ *
+ * struct CudaBackend {
+ *   using ExecutorType = CudaTcExecutor;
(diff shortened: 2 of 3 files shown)


tc/core/compiler.h
```diff
@@ -0,0 +1,123 @@
...
+ *   using MappingOptionsType = CudaMappingOptions;
+ *   using CompilationResultType = CudaCompilationResult;
+ *   using RTCFunctionType = CudaRTCFunction;
+ * };
+ *
+ * Sketching usage resembles:
+ *   std::string someTc = "...";
+ *   auto pExecutor = tc::compile<CudaBackend>(
+ *     someTc, tcFunctionName, inputs, mappingOptions);
+ *   auto profilingInfo = pExecutor->run(handle, inputs, outputs, true);
```
ftynse: didn't we split this into `void run()` and `someType profile()`?

tc/core/compiler.h
```diff
@@ -0,0 +1,123 @@
...
+ *   // alternatively:
+ *   // auto kernelTiming = pExecutor->uncheckedRun(inputs, outputs, true);
+ */
+namespace tc {
+/// Given a TC representation as a TC + TC function name entryPoint, this
```
ftynse: "TC representation as a TC" does not really make sense.  Given a TC as a string maybe?

tc/core/compiler-inl.h
```diff
@@ -0,0 +1,101 @@
...
+void checkInputsCompliant(
+    const tc2halide::HalideComponents& halideComponents,
+    const std::vector<const DLConstTensor*>& inputsInfo) {
+  if (inputsInfo.size() != halideComponents.inputs.size()) {
+    throw lang::ErrorReport(halideComponents.getDef())
+        << "expected " << halideComponents.inputs.size() << " inputs but found "
+        << inputsInfo.size();
+  }
+  for (size_t i = 0; i < inputsInfo.size(); ++i) {
+    auto dltype_ = inputsInfo[i]->dtype;
```
ftynse: there is a slight inconsistency in naming here:
- dltype and htype imply "dltensor" and "halide", but later we also use adim and edim for expected/actual;  It is not immediately clear what is what in general
- names suffixed with `_` are used for private members;  we are in a free function, so this is not really a big deal

tc/core/compiler-inl.h
```diff
@@ -0,0 +1,101 @@
...
+#include <string>
+
+#include "tc/core/flags.h"
+#include "tc/core/halide_utils.h"
+#include "tc/core/tensor.h"
+#include "tc/lang/canonicalize.h"
+
+namespace tc {
+namespace {
+void checkInputsCompliant(
```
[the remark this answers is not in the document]
ttheodor: It can't. But the anonymous namespace is not useful: anyone who includes this header (indirectly by including `compiler.h`) can call `checkInputsCompliant` (and each translation unit will have a separate version of this function which, I guess, is not what's intended here).

nicolasvasilache:
There are a few of those anon namespaces throughout. It is more for
documentation purposes than to enforce at compile time. The  letter
alternative is to put in a .cc but then we're in the business of
instantiating templates manually, which is terrible

I could alternately make it into a private namespace this way one really
needs to go out of one's way to use the function.

Other suggestions?

ttheodor:
> alternative is to put in a .cc but then we're in the business of
> instantiating templates manually, which is terrible

`checkInputsCompliant` is not a function template.

nicolasvasilache:
This is transitive if a function template calls it even indirectly then you
have some issue: even if you put it in a .cc then you need to expose the
function  somehow so that it is accessible in the template.

On my phone atm so I can't browse code.

Even if this is particular one can be put in a .cc, my question is still
open for the other use cases and there are multiple occurrences.

What do you guys prefer?

I

ftynse:
> need to expose the function somehow so that it is accessible in the template.

If there is a declaration before the template, and the objects where the template is instantiated are linked to the object where the function is defined, it's okay.

I was fine with the original version.

skimo-openhub:
> (or even compiled) therefore the commit is guaranteed correct.
> This new implementation will be switched in as part of the last commit of the
> the global refactoring.

The more I see this, the more I dislike it.  I understand your reasons for taking
this route, but you shouldn't make it sound like something positive ("guaranteed correct").
It should be more like "this commit unfortunately cannot be tested on its own".

skimo-openhub:
On Fri, Apr 27, 2018 at 01:01:46PM +0000, Nicolas Vasilache wrote:
> This is transitive if a function template calls it even indirectly then you
> have some issue: even if you put it in a .cc then you need to expose the
> function  somehow so that it is accessible in the template.
> 
> On my phone atm so I can't browse code.
> 
> Even if this is particular one can be put in a .cc, my question is still
> open for the other use cases and there are multiple occurrences.
> 
> What do you guys prefer?

Declare checkInputsCompliant in namespace detail and
implement it in .cc.

If you really prefer keeping it in .h, then it should
be declared "inline", no?

skimo

tc/core/compiler-inl.h
```diff
@@ -0,0 +1,101 @@
...
+} // namespace
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const typename Backend::MappingOptionsType& options) {
+  auto parsedTcs = detail::parse(tc);
+  CHECK_EQ(1, parsedTcs.count(entryPoint))
```
skimo-openhub: Isn't this generating a warning for you?
I remember having to change all these constants to be unsigned.
Also, I find it much easier to read if you put the target of the check
before what you are checking it against.

Same for the other `CHECK_EQ` introduced in this commit.

tc/core/compiler.h
```diff
@@ -0,0 +1,123 @@
...
+ *   auto profilingInfo = pExecutor->profile(handle, inputs, outputs, true);
+ *   // alternatively:
+ *   // auto kernelTiming = pExecutor->uncheckedRun(inputs, outputs, true);
+ */
+namespace tc {
+/// Given a TC string containing multiple functions and a TC function name
+/// "entryPoint", this function compiles a new TcExecutor for the specified
+/// Backend. For now, contiguous output sizes are inferred given input sizes.
+/// If you need another kernel for another entryPoint or another inputs,
+/// options then just compile another TcExecutor; because atm we fully JIT
```
skimo-openhub: "another inputs, options" still looks wrong

tc/core/compiler.h
```diff
@@ -0,0 +1,123 @@
...
+namespace detail {
+/// Given a TC representation, this parses the TC functions into a map of
+/// TreeRef indexed by TC function names.
+/// \returns an ordered map of TC function name to parsed TC tree
+std::map<std::string, lang::TreeRef> parse(const std::string& tc);
+
+/// Given a TC representation as a TreeRef, this function compiles a new
+/// TcExecutor for the specified Backend.
+/// For now, contiguous output sizes are inferred given input sizes.
+/// If you need another kernel for another TC or another inputs or options then
```
skimo-openhub: Same here

salexspb:
regarding .cc files - everything which can be there, should be. Then we reduce compilation time for changes which go only inside those .cc files.

tc/core/compiler-inl.h
salexspb: do you mind to start using ENFORCE instead of a check here and in other places? Check crashes the process. And this can be a problem going forward with productionization (if we put this code in inference). 
Feel free to borrow CAFFE_ENFORCE code :) 

Regarding CHECK_EQ it should be preferred over CHECK, agree with Nicolas. As it automatically prints values. For this reason we have ENFORCE_EQ :)

tc/core/compiler-inl.h
```diff
@@ -0,0 +1,101 @@
...
+  checkInputsCompliant(halideComponents, inputs);
+
+  auto tcName = lang::Def(tcDefinition).name().name();
+  CompilationResultType compilationResult = Backend::compileWithTcMapper(
+      tcName,
+      halideComponents,
+      inputs,
+      /* TODO outputs, */
+      options);
+  return std::unique_ptr<typename Backend::ExecutorType>(
```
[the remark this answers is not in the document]
salexspb: when using this make_unique, please use tc:: namespace, btw. Otherwise it causes compilation errors in a project which uses c++14 and includes <memory>

tc/core/compiler-inl.h
```diff
@@ -0,0 +1,101 @@
...
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    lang::TreeRef tcDefinition,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const typename Backend::MappingOptionsType& options) {
+  using CompilationResultType = typename Backend::CompilationResultType;
+
+  auto inputsInfo = makeTensorInfoVector(inputs);
+  auto outputsInfo = detail::inferOutputTensorInfo(tcDefinition, inputs);
```
salexspb: so we are going to infer it two times. Once here and once in places like TcOp. Do you think we can have an interface which allows to get this from the get go and avoid output shape inference twice?

tc/core/compiler-inl.h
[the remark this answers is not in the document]
nicolasvasilache: not using something custom here, we'll use the C++14 variant when we decide to pull the trigger on C++11-only

nicolasvasilache:
Reg decl. none of the proposals I have seen result in completely hiding the helper function from the user.
I'll just make the namespace detail rather than anonymous and make the function inline; I don't see value in file proliferation for a single function that can easily be inlined.

tc/core/compiler-inl.h
nicolasvasilache: Old, habit, it used to be (expected, actual).

Reg warnings .. no it's explicitly disabled on my machine due to [this](https://github.com/facebookresearch/TensorComprehensions/blob/master/CMakeLists.txt#L23)
```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-variable -Werror -Wno-sign-compare")
```

I disabled locally, I checked I indeed compile with -Wall -Werror but I still do not see anything.
I'll still make it 1u

nicolasvasilache: Please see #351, the idea we floated was to use exceptions everywhere, will happen globally over the whole codebase when it is addressed.
Do you see a value in maintaining 2 sets of macros?
Care to comment and add suggestions to #351?

Thanks!

PR #360, opened 2018-04-26
create a single block mapping filter and a single thread mapping filter per branch

skimo-openhub:
After these changes, there is only a single call left to `makeMappingFilter`
(in `MappedScop::map`).  This will make it easy to change the representation
of the mapping filter to contain extra information available inside `MappedScop::map`.
The change of the representation itself is left for a later PR.

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -108,24 +108,6 @@ const CudaDim& mappingSize<mapping::ThreadId>(const MappedScop* mscop) {
 }
 } // namespace
 
-template <typename MappingTypeId>
-void MappedScop::mapRemaining(detail::ScheduleTree* tree, size_t nMapped) {
-  size_t nToMap = mappingSize<MappingTypeId>(this).view.size();
-  if (nMapped >= nToMap) {
-    return;
-  }
-
-  std::unordered_set<MappingTypeId, typename MappingTypeId::Hash> ids;
-  for (size_t i = nMapped; i < nToMap; ++i) {
-    ids.insert(MappingTypeId::makeId(i));
-  }
-  auto root = scop_->scheduleRoot();
-  auto domain = activeDomainPoints(root, tree);
-  auto filter = makeFixRemainingZeroFilter(domain, ids);
-  auto mapping = detail::ScheduleTree::makeMappingFilter(filter, ids);
-  insertNodeAbove(root, tree, std::move(mapping));
-}
-
 // Map the elements in "list" to successive blocks or thread identifiers,
 // with the first element mapped to identifier X.  The extents are obtained
 // from the initial elements of numBlocks or numThreads.  The identifiers
@@ -166,10 +148,17 @@ detail::ScheduleTree* MappedScop::map(
     idSet.emplace(id);
   }
 
+  std::unordered_set<MappingTypeId, typename MappingTypeId::Hash> unmapped;
+  for (size_t i = nToMap; i < extent.size(); ++i) {
+    auto id = MappingTypeId::makeId(i);
+    unmapped.emplace(id);
+    idSet.emplace(id);
+  }
+  filter = filter.intersect(makeFixRemainingZeroFilter(domain, unmapped));
+
   auto mapping = detail::ScheduleTree::makeMappingFilter(filter, idSet);
   tree = insertNodeAbove(root, tree, std::move(mapping))->child({0});
 
-  mapRemaining<MappingTypeId>(tree, nToMap);
   return tree;
 }
 
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -175,10 +175,7 @@ void fixThreadsBelow(
 
   auto band = detail::ScheduleTree::makeEmptyBand(mscop.scop().scheduleRoot());
   auto bandTree = insertNodeBelow(tree, std::move(band));
-  auto ctx = tree->ctx_;
-  insertNodeBelow(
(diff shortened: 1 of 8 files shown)


tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -107,42 +107,47 @@ const CudaDim& mappingSize<mapping::ThreadId>(const MappedScop* mscop) {
...
-  auto band = tree->elemAs<detail::ScheduleTreeElemBand>();
-  CHECK(band) << "expected a band, got " << *tree;
-  CHECK_GE(pos, 0u) << "dimension underflow";
-  CHECK_LT(pos, band->nMember()) << "dimension overflow";
-  CHECK_NE(extent, 0u) << "NYI: mapping to 0";
+    isl::union_pw_aff_list list,
+    const std::vector<MappingIdType>& ids,
+    const CudaDimView& extent) {
+  CHECK_EQ(ids.size(), list.n());
+  CHECK_LE(list.n(), extent.size()) << "dimension overflow";
```
ftynse: I would add a function-level comment mentioning this assumption: if "extent" has more values than "ids", only the first ids.size() values of "extent" will be used.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -107,42 +107,47 @@ const CudaDim& mappingSize<mapping::ThreadId>(const MappedScop* mscop) {
...
-  // Create mapping filter by equating the newly introduced
-  // parameter "id" to the "pos"-th schedule dimension modulo its extent.
-  auto upa =
-      band->mupa_.get_union_pw_aff(pos).mod_val(isl::val(tree->ctx_, extent));
-  upa = upa.sub(isl::union_pw_aff::param_on_domain(domain, id));
-  auto filter = upa.zero_union_set();
-  auto mapping =
-      detail::ScheduleTree::makeMappingFilter<MappingIdType>(filter, {id});
+    // Create mapping filter by equating the newly introduced
+    // parameter "id" to the "pos"-th schedule dimension modulo its extent.
```
ftynse: this comment is outdated

tc/core/polyhedral/cuda/mapped_scop.h
```diff
@@ -134,6 +134,14 @@ class MappedScop {
   }
 
  private:
+  // Map the elements in "list" to successive blocks or thread identifiers,
+  // with the first element mapped identifier X.
```
ftynse: mapped _to_ identifier X

skimo-openhub:
On Fri, Apr 27, 2018 at 08:53:50AM +0000, ftynse wrote:
> 
> shouldn't 384ccc4f9d170224c3b7f049f8ef9825db7b0428 (Bump isl for isl::union_pw_aff_list::reverse and isl:union_pw_aff_involves_param) go before b937b5b211849c1b4b8bd42b0188761418327246 that uses these functions?

It does come before (in history, at least; perhaps you are confused
by the fact that github ignores history)

skimo

ftynse:
> It does come before (in history, at least; perhaps you are confused by the fact that github ignores history)

Yes, I'm looking at commits in the order of github... Nevermind then.

skimo-openhub:
f84119cc74ab8ec09bab2186d491faf763b57326: fixed typo in commit message
51229f5c9996c5ba57cbe828f306beaa269c9fc2: updated comments
bff75f5e5ef71acf850a7715f07dc0b92a78a7c5: updated comments

diff:
```
diff --git a/tc/core/polyhedral/cuda/mapped_scop.cc b/tc/core/polyhedral/cuda/mapped_scop.cc
index 627299f5..a44abffb 100644
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -109,8 +109,8 @@ const CudaDim& mappingSize<mapping::ThreadId>(const MappedScop* mscop) {
 } // namespace
 
 // Map the elements in "list" to successive blocks or thread identifiers,
-// with the first element mapped to identifier X.
-// The extents are obtained from numBlocks or numThreads.  The identifiers
+// with the first element mapped to identifier X.  The extents are obtained
+// from the initial elements of numBlocks or numThreads.  The identifiers
 // must not be present in the space of the partial schedules in "list" and
 // extents must be non-zero.  The mapping corresponds to inserting a filter
 // node with condition 'list % extent = ids'.
@@ -140,7 +140,7 @@ detail::ScheduleTree* MappedScop::map(
     CHECK_NE(extent[i], 0u) << "NYI: mapping to 0";
 
     // Create mapping filter by equating the newly introduced
-    // parameter "id" to the "pos"-th schedule dimension modulo its extent.
+    // parameter ids[i] to the "i"-th affine function modulo its extent.
     upa = upa.mod_val(isl::val(tree->ctx_, extent[i]));
     upa = upa.sub(isl::union_pw_aff::param_on_domain(domain, id));
     filter = filter.intersect(upa.zero_union_set());
diff --git a/tc/core/polyhedral/cuda/mapped_scop.h b/tc/core/polyhedral/cuda/mapped_scop.h
index e9c08581..0ebd5d4c 100644
--- a/tc/core/polyhedral/cuda/mapped_scop.h
+++ b/tc/core/polyhedral/cuda/mapped_scop.h
@@ -129,7 +129,7 @@ class MappedScop {
 
  private:
   // Map the elements in "list" to successive blocks or thread identifiers,
-  // with the first element mapped identifier X.
+  // with the first element mapped to identifier X.
   // Return a pointer to the updated node (below the inserted filter)
   // for call chaining purposes.
   template <typename MappingTypeId>
```

ftynse:
Running tests....

ftynse:
tests pass locally /@ftynse

PR #361, opened 2018-04-26
[C++ API] Step 3: add refactored tensor API files

nicolasvasilache:
This PR adds the files refactoring the tensor API #313.

In particular this:
1. adds a DLConstTensor for proper const correctness propagation + proper
overloads and specializations
2. uniformizes tensor related functions and put them in the same location
3. makes TensorInfo more visible as the tensor representation on which
compilation and cache interaction occur cleanup unused functions
4. makes this file the only place that includes <dlpack/dlpack.h>

the global refactoring.

### DIFF
--- a/tc/core/tensor-inl.h
+++ b/tc/core/tensor-inl.h
@@ -1,0 +1,263 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <algorithm>
+#include <sstream>
+
+#include <glog/logging.h>
+#include <google/protobuf/text_format.h>
+
+#include "tc/proto/compcache.pb.h"
+
+namespace tc {
+template <typename T>
+std::vector<T> makeStridesFromSizes(const std::vector<T>& sizes) {
+  auto ndim = sizes.size();
+  if (ndim == 0) {
+    return std::vector<T>();
+  }
+  std::vector<T> strides(sizes.size(), 0);
+  strides[ndim - 1] = 1;
+  for (int i = ndim - 2; i >= 0; --i) {
+    strides[i] = strides[i + 1] * sizes[i + 1];
+  }
+  return strides;
+}
+
+inline DLContext getCPUDLContext() {
+  DLContext res;
+  res.device_id = 0;
+  res.device_type = DLDeviceType::kDLCPU;
+  return res;
+}
+
+inline DLContext getGPUDLContext(int device_id) {
+  DLContext res;
+  res.device_id = device_id;
+  res.device_type = DLDeviceType::kDLGPU;
+  return res;
+}
+
+template <typename DLTensorType>
+void DLTensorDeleter::operator()(const DLTensorType* t) {
+  if (t->shape) {
+    delete[] t->shape;
+  }
+  if (t->strides) {
+    delete[] t->strides;
+  }
+  delete t;
+};
+
+namespace detail {
+template <typename DLTensorType, typename T>
+inline std::unique_ptr<DLTensorType, DLTensorDeleter> makeDLTensor(
+    DLContext ctx,
+    DLDataType dtype,
+    const std::vector<T>& sizes,
+    const std::vector<T>& strides = std::vector<T>(),
+    decltype(DLTensorType().data) data = nullptr,
+    uint64_t byteOffset = 0) {
+  static_assert(
+      std::is_convertible<T, int64_t>::value,
+      "Template type not convertible to int64_t");
+  std::unique_ptr<DLTensorType, DLTensorDeleter> res(new DLTensorType);
+  res->data = data;
+  res->ctx = ctx;
+  auto ndim = sizes.size();
+  res->ndim = ndim;
+  res->dtype = dtype;
+  res->shape = new int64_t[ndim];
+  for (size_t i = 0; i < ndim; ++i) {
+    res->shape[i] = sizes[i];
+  }
+  res->strides = new int64_t[ndim];
+  std::vector<T> st(strides);
+  if (st.size() == 0) {
+    st = makeStridesFromSizes(sizes);
+  }
+  for (size_t i = 0; i < ndim; ++i) {
+    res->strides[i] = st[i];
+  }
+  res->byte_offset = byteOffset;
+  return res;
+}
+
+template <typename DLTensorType>
+inline std::unique_ptr<DLTensorType, DLTensorDeleter> makeDLTensorHelper(
+    const DLTensor* ptr) {
+  std::vector<int64_t> sizes(ptr->ndim, 0);
+  std::copy(ptr->shape, ptr->shape + ptr->ndim, sizes.begin());
+  std::vector<int64_t> strides(ptr->ndim, 0);
+  std::copy(ptr->strides, ptr->strides + ptr->ndim, strides.begin());
+  return makeDLTensor<DLTensorType>(
+      ptr->ctx, ptr->dtype, sizes, strides, ptr->data, ptr->byte_offset);
+}
+} // namespace detail
+
+inline DLTensorUPtr makeDLTensor(const DLTensor* ptr) {
+  return detail::makeDLTensorHelper<DLTensor>(ptr);
+}
+
+inline DLTensorUPtr makeDLTensor(const TensorInfo& tensor) {
+  return detail::makeDLTensor<DLTensor>(
+      DLContext{kDLCPU, 0},
+      tensor.dtype,
+      tensor.shape,
+      tensor.strides,
+      nullptr,
+      tensor.alignment);
+}
+
+template <typename T>
+inline DLTensorUPtr makeDLTensor(
+    DLContext ctx,
+    DLDataType dtype,
+    const std::vector<T>& sizes,
+    const std::vector<T>& strides,
+    void* data,
+    uint64_t byteOffset) {
+  return detail::makeDLTensor<DLTensor>(
+      ctx, dtype, sizes, strides, data, byteOffset);
+}
+
+template <typename DLTensorType>
+inline DLConstTensorUPtr makeDLConstTensor(const DLTensorType* ptr) {
+  return detail::makeDLTensorHelper<DLConstTensor>(ptr);
+}
+
+inline DLConstTensorUPtr makeDLConstTensor(const TensorInfo& tensor) {
+  return detail::makeDLTensor<DLConstTensor>(
+      DLContext{kDLCPU, 0},
+      tensor.dtype,
+      tensor.shape,
+      tensor.strides,
+      nullptr,
+      tensor.alignment);
+}
+
+template <typename T>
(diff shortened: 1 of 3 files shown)


tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+  res.device_id = device_id;
+  res.device_type = DLDeviceType::kDLGPU;
+  return res;
+}
+
+template <typename T>
+DLDataType getDLDataType();
+
+template <>
+inline DLDataType getDLDataType<float>() {
```
ftynse: Don't we want other types (`double`, other `(u)int` types) ? 
I don't see this function used though.  Add a TODO maybe...

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+};
+
+template <typename T, typename DLTensorType, typename VoidPtrType>
+inline void fillDLTensor(
+    DLTensorType* res,
+    DLContext ctx,
+    DLDataType dtype,
+    const std::vector<T>& sizes,
+    const std::vector<T>& strides = std::vector<T>(),
+    VoidPtrType* data = nullptr,
```
ftynse: I wouldn't call this `VoidPtrType`, it's just the type of data stored.  Actually, there's no need to instantiate this template for different sizes because the underlying dltensor uses an actual `void*`.  This will only bloat the code, slowing down compilation, with no additional type safety.

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+    uint64_t byte_offset = 0) {
+  res->data = data;
+  res->ctx = ctx;
+  auto ndim = sizes.size();
+  res->ndim = ndim;
+  res->dtype = dtype;
+  res->shape = new int64_t[ndim];
```
ftynse: maybe add a static assert that `std::is_convertible<T, int64_t>::value`?

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+
+template <typename T, typename DLTensorType, typename VoidPtrType>
+inline void fillDLTensor(
+    DLTensorType* res,
+    DLContext ctx,
+    DLDataType dtype,
+    const std::vector<T>& sizes,
+    const std::vector<T>& strides = std::vector<T>(),
+    VoidPtrType* data = nullptr,
+    uint64_t byte_offset = 0) {
```
ftynse: camelCase?

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+
+inline DLTensorUPtr makeDLTensor(const TensorInfo& tensor) {
+  DLTensorUPtr res(new DLTensor);
+  fillDLTensor(
+      res.get(),
+      DLContext{kDLCPU, 0},
+      tensor.dtype,
+      tensor.shape,
+      tensor.strides,
+      static_cast<void*>(nullptr),
```
ftynse: `nullptr_t` (type of `nullptr`) is implicitly-convertible into any pointer

tc/core/tensor.cc
```diff
@@ -0,0 +1,123 @@
...
+#include <numeric>
+#include <tuple>
+
+#include "tc/core/tensor.h"
+#include "tc/core/utils/math.h"
+
+namespace tc {
+
+namespace {
+uint64_t GetDLTensorAlignment(const DLTensor* t) {
```
ftynse: lowerCamelCase for functions

tc/core/tensor.cc
```diff
@@ -0,0 +1,123 @@
...
+  if (not t->strides) {
+    return;
+  }
+  strides.reserve(t->ndim);
+  std::copy(t->strides, t->strides + t->ndim, std::back_inserter(strides));
+}
+
+TensorInfo::TensorInfo(const DLConstTensor* t)
+    : dtype(t->dtype), alignment{GetDLTensorAlignment(t)} {
+  shape.reserve(t->ndim);
```
ftynse: This is a copy-paste from `TensorInfo::TensorInfo(const DLTensor* t)`.   Can't we use templates, delegating constructors or some other mechanism to avoid copy-pasting?

tc/core/tensor.cc
```diff
@@ -0,0 +1,123 @@
...
+  buf.mutable_dtype()->set_bits(dtype.bits);
+  buf.mutable_dtype()->set_lanes(dtype.lanes);
+  return buf;
+}
+
+bool operator==(const DLDataType& a, const DLDataType& b) {
+  return a.code == b.code and a.bits == b.bits and a.lanes == b.lanes;
+}
+
+bool operator<(const DLDataType& a, const DLDataType& b) {
```
ftynse: Not sure default less-comparison operators make sense for types.  Why do we need them, to use types in ordered containers?  I would prefer function objects `DLDataTypeSimpleCompare`.

tc/core/tensor-inl.h
nicolasvasilache: We have both ```void*``` and ```const void*``` for DLTensor and ```DLConstTensor``` (coming from inputs and outputs of networks as per caffe2 so there is additional type safety).

skimo-openhub:
Minor comments on the commit message:

>     This PR adds the files refactoring the tensor API #313.

nit: the commit message will go down in history, so you should formulate it
in terms as if it had already been accepted.  So, in the commit message,
you should refer to "commit" instead of "PR".

>     In particular this:
>     1. adds a DLConstTensor for proper const correctness propagation + proper
>     overloads and specializations
>     2. uniformizes tensor related functions and put them in the same location

typo: "puts"

>     3. makes TensorInfo more visible as the tensor representation on which
>     compilation and cache interaction occur cleanup unused functions

Is "cleanup unused functions" meant to be a separate item?

>     4. makes this file the only place that includes <dlpack/dlpack.h>
>     
>     As part of the bigger refactoring, these files are not yet activated
>     (or even compiled) therefore the commit is guaranteed correct.
>     This new implementation will be switched in as part of the last commit of the
>     the global refactoring.
>     
>     The larger branch pr/big-refactor maintains an up-to-date and tested
>     implementation of the new API to keep us honest.

This sentence should be removed from the commit message.

tc/core/tensor.cc
```diff
@@ -0,0 +1,123 @@
...
+
+TensorInfo::TensorInfo(
+    DLDataType t,
+    uint64_t align,
+    const std::vector<int64_t>& sz,
+    const std::vector<int64_t>& st)
+    : dtype(t), alignment(align), shape(sz), strides(st) {}
+
+TensorInfo::TensorInfo(const DLTensor* t)
+    : dtype(t->dtype), alignment{GetDLTensorAlignment(t)} {
```
skimo-openhub: Just curious, but why uniform initialization {} on `alignment`?

tc/core/tensor.h
```diff
@@ -0,0 +1,156 @@
...
+template <typename T>
+inline DLConstTensorUPtr makeDLConstTensor(
+    DLContext ctx,
+    DLDataType dtype,
+    const std::vector<T>& sizes,
+    const std::vector<T>& strides = std::vector<T>(),
+    const void* data = nullptr,
+    uint64_t byte_offset = 0);
+// A metadata-owning DLTensor reconstructed from TensorInfo does not have a
+// meaningful pointer or DLContext: pointer is  nullptr and ctx is kDLCPU.
```
skimo-openhub: nit: extra space before "nullptr"

Also, if there is no meaningful DLContext, then why not set it to zero instead of `kDLCPU`?

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+
+#include <algorithm>
+#include <sstream>
+
+#include <glog/logging.h>
+
+namespace tc {
+
+template <typename DLTensorType>
+inline void setSizes(DLTensorType& t, const std::vector<int64_t>& sizes) {
```
salexspb: which code style are you using here? Normally we use Google code style which uses pointers for mutable arguments. And they go in the end of the list of function inputs. This is also the most common style at Facebook AFAIK

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+  auto ndim = sizes.size();
+  for (size_t i = 0; i < ndim; ++i) {
+    t.shape[i] = sizes[i];
+  }
+}
+
+template <typename DLTensorType>
+inline void setStrides(DLTensorType& t, const std::vector<int64_t>& strides) {
```
salexspb: ditto

tc/core/tensor.h
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * DLPack is missing a ConstTensor. Add it here for now.
+ * In the future, this would be contributed back to dlpack/include/dlpack.h.
+ */
+typedef struct {
+  const void* data;
+  DLContext ctx;
+  int ndim;
+  DLDataType dtype;
+  int64_t* shape;
```
salexspb: sorry, I might've missed your earlier reply. Why not just get rid of DLPack all together and have own struct which is RAII complaint? I.e. uses vectors instead of raw pointers. Then you don't need to do all the custom deleter business :)

tc/core/tensor.h
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * TensorInfo wraps the necessary tensor information to compile TCs and
+ * interact with the various caches.
+ * Notably, it contains alignment information but no underlying data pointer.
+ * It is serializable to protobuf and stored directly in the cache.
+ */
+struct TensorInfo {
+  DLDataType dtype;
+  uint64_t alignment;
```
salexspb: how do you use this alignment if you don't control the pointer allocation?

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+        case 8:
+          return "uint8_t";
+      }
+      break;
+  }
+  CHECK(false) << "NYI: toString for type: " << t.code << ", bits: " << t.bits;
+  return "";
+}
+
+inline std::string toString(const DLTensor& t) {
```
salexspb: did you consider using protobufs and TextFormat?

tc/core/tensor-inl.h
```diff
@@ -0,0 +1,322 @@
...
+      }
+      break;
+    case DLDataTypeCode::kDLUInt:
+      switch (t.bits) {
+        case 8:
+          return "uint8_t";
+      }
+      break;
+  }
+  CHECK(false) << "NYI: toString for type: " << t.code << ", bits: " << t.bits;
```
salexspb: could you use exceptions instead, please?

tc/core/tensor-inl.h
ttheodor: <rant>
I would strongly argue against passing pointers due to  "historical reasons", especially in a new project. A pointer can be null, but in this particular we never want it to be null. So we either introduce a check and possibly throw at runtime, or we use a reference which makes explicit the fact that we expect a fully constructed object. 

Yes, a user can go out of her/his way to construct a reference out of a `nullptr` but then it's their fault.
</rant>

tc/core/tensor-inl.h
nicolasvasilache: fixing the ones I can, note some are coming from DLPack

tc/core/tensor-inl.h
nicolasvasilache: well dead code should die, done

tc/core/tensor-inl.h
nicolasvasilache: I always hate that discussion, both sides have reasonable arguments.
So I'll just be removing that useless code. 
In general and as much as possible we favor return types.

tc/core/tensor.h
nicolasvasilache: we are not using it yet; in the future it will be used to switch between different vectorized versions of a same kernel

tc/core/tensor-inl.h
nicolasvasilache: Let's do it once and for all in #351 for the whole codebase, this PR will not start spinning off a solution for #351

nicolasvasilache:

PR #362, opened 2018-04-26
[C++ API] Step 4: add the new ATen compiler files

nicolasvasilache:
This PR adds the files implementing the ATen user-facing part (#324) on
top of the new compiler API (#322). The ExecutionEngine has been retired
and the user-facing API reflects this.

This provides a minimal API to wrap ```tc/core/compiler.h``` under a tensor
library like ATen. In particular this gets rid of the ```ATenCompilationUnit```
class which kept state about parsed TCs.

Tinkering with the underlying TreeRef memoization is penny-wise and pound
foolish: parsing overhead never matters and it adds cognitive overhead to the
API. Therefore this API only provides functionality operating on TC strings.

To guard against overhead, the executors themselves should be memoized and
reused.

the global refactoring.

### DIFF
--- a/tc/aten/aten_compiler_new_api-inl.h
+++ b/tc/aten/aten_compiler_new_api-inl.h
@@ -1,0 +1,87 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "tc/aten/aten_compiler.h"
+
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/core/compiler.h"
+#include "tc/core/tc_executor.h"
+#include "tc/core/tensor.h"
+
+namespace tc {
+namespace aten {
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs,
+    const typename Backend::MappingOptionsType& options) {
+  auto inputDLTensors = makeDLConstTensors(inputs);
+  return tc::compile<Backend>(
+      tc, entryPoint, extractRawPtrs(inputDLTensors), options);
+}
+
+template <typename Executor>
+void run(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs) {
+  auto inputDLTensors = makeDLConstTensors(inputs);
+  auto outputDLTensors = makeDLTensors(outputs);
+  return executor.run(
+      extractRawPtrs(inputDLTensors), extractRawPtrs(outputDLTensors));
+}
+
+template <typename Executor>
+ProfilingInfo profile(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs) {
+  auto start = std::chrono::system_clock::now();
+  auto inputDLTensors = makeDLConstTensors(inputs);
+  auto outputDLTensors = makeDLTensors(outputs);
+  ProfilingInfo pi(executor.profile(
+      extractRawPtrs(inputDLTensors), extractRawPtrs(outputDLTensors)));
+
+  // The total CPU overhead is the total time minus the (synchronized) kernel
+  // runtime
+  auto end = std::chrono::system_clock::now();
+  Duration cpuOverhead(end - start);
+  cpuOverhead = cpuOverhead - pi.kernelRuntime;
+  return ProfilingInfo{cpuOverhead, pi.kernelRuntime};
+}
+
+template <typename Executor>
+void uncheckedRun(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs) {
+  CHECK_GE(outputs.size(), 1u);
+  std::vector<const void*> rawInputs(inputs.size(), nullptr);
+  std::vector<void*> rawOutputs(outputs.size(), nullptr);
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    rawInputs[i] = inputs[i].data_ptr();
+  }
+  for (size_t i = 0; i < outputs.size(); ++i) {
+    rawOutputs[i] = outputs[i].data_ptr();
+  }
+  return executor.uncheckedRun(rawInputs, rawOutputs);
+}
+} // namespace aten
+} // namespace tc
--- a/tc/aten/aten_compiler_new_api.h
+++ b/tc/aten/aten_compiler_new_api.h
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <chrono>
+#include <string>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/core/tensor.h"
+#include "tc/core/utils/time.h"
+
+namespace tc {
+namespace aten {
+/// Given a TC string with multiple functions defined and a TC function name
+/// entryPoint, this runs inference, applied to the specified input shapes.
+/// This is used for manual processing of output metadata on the ATen user
+/// side.
+/// \returns the (contiguous) output tensor shapes as a metadata-owning
+/// DLTensorUPtr.
+std::vector<tc::DLTensorUPtr> inferOutputTensorInfo(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs);
+
+/// Given a TC string with multiple functions defined and a TC function name
+/// entryPoint, this runs inference, applied to the specified input shapes and
+/// allocates fresh new output tensors.
+/// If one wants inplace/resize behavior, one can implement it using
+/// inferOutputTensorInfo.
+/// \returns the (contiguous) output tensors with properly inferred sizes.
+std::vector<at::Tensor> prepareOutputs(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs);
+
+/// Given a TC string with multiple functions defined and a TC function name
+/// entryPoint, compile the TC for the specified input shapes with the
+/// prescribed options.
+/// \returns an Executor for the specified backend with an underlying
+/// JIT-compiled callable function.
+template <typename Backend>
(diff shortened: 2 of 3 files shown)


tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,145 @@
...
+using tc::DLTensorUPtr;
+
+namespace tc {
+namespace aten {
+inline std::vector<DLTensorUPtr> inferOutputTensorInfo(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs) {
+  auto parsedTcs = tc::detail::parse(tc);
+  CHECK_EQ(1, parsedTcs.count(entryPoint))
```
ftynse: shouldn't this be rather a lang ErrorReport?

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,145 @@
...
+  }
+  CHECK_LT(0, inputs.size())
+      << "Need >= 1 input tensors to determine backend and prepare ATen outputs";
+  auto atenBackend = inputs[0].type().backend();
+  for (size_t i = 0; i < outTensorInfo.size(); ++i) {
+    auto info = outTensorInfo[i];
+    auto stype = at::toScalarType(info.dtype);
+    outputs.push_back(
+        at::getType(atenBackend, stype).tensor(at::IntList(info.shape)));
+    // TODO: we just malloc'ed I guess we can pay an inplace memset
```
ftynse: Why do we want to zero-initialize outputs?  We now throw a warning if they are not written to in the code.

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,145 @@
...
+}
+
+template <typename Executor>
+void uncheckedRun(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs) {
+  CHECK_LT(0, outputs.size());
+
+  constexpr auto kReservedSize = 8;
```
ftynse: Can't we use an actual `.reserve()` instead and remove the extra branching logic below?  I'm not sure the compiler is smart enough to allocate small constant-sized vectors on stack.  If you want that, you need something like `SmallVector` from LLVM.

ftynse: (not saying we should depend on LLVM here though)

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,145 @@
+/**
```
ftynse: do we really want to have a file called `*_new_api.h`?

tc/aten/aten_compiler_new_api.h
```diff
@@ -0,0 +1,88 @@
...
+#include <string>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/core/utils/time.h"
+
+namespace tc {
+namespace aten {
+/// Given a TC string with multiple functions defined and a TC function name,
+/// this run inference, applied to the specified input shapes.
```
ftynse: runs

tc/aten/aten_compiler_new_api.h
```diff
@@ -0,0 +1,88 @@
...
+/// If one wants inplace/resize behavior, one can implement it using
+/// inferOutputTensorInfo.
+/// \returns the (contiguous) output tensors with properly inferred sizes.
+std::vector<at::Tensor> prepareOutputs(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs);
+
+/// Given a TC string with multiple functions defined and a TC function name,
+/// compile the TC for the specified input shapes with the prescribed
```
ftynse: an explanation of what `entryPoint` means is missing

tc/aten/aten_compiler_new_api-inl.h
nicolasvasilache: No we shouldn't but this is the collision-less way of making progress that makes things reviewable.
I wouldn't want to dump the big PR on you ;)
I can rebase and move in the final PR that will switch everything.

salexspb:
@nicolasvasilache , high level question about these files not being compiled in. Do we need to manually block them from synchronization to fbcode? Currently would automatically import them. Also if this is not going to be compiled, is there a point to merge this till the full stack is ready?

skimo-openhub:
As @ftynse mentioned on another PR, don't use triple backticks to quote inline code, especially in the commit message.

tc/aten/aten_compiler_new_api.h
```diff
@@ -0,0 +1,88 @@
...
+template <typename Executor>
+void run(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs);
+
+/// Given an executor resulting from compiling a TC, run the TC and fill the
+/// outputs vector with the results.
+/// \returns ProfilingInfo (cpuOverhead + kernelRuntime) in microseconds.
+/// If profile=false, kernelRuntime is set to Duration::max.
```
skimo-openhub: Stale comment?

tc/aten/aten_compiler_new_api-inl.h
skimo-openhub: I also think this looks like a premature optimization.

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,145 @@
...
+  std::vector<void*> O(kReservedSize, nullptr);
+  size_t i;
+  for (i = 0; i < inputs.size(); ++i) {
+    if (i < kReservedSize) {
+      I[i] = inputs[i].data_ptr();
+    } else {
+      I.push_back(inputs[i].data_ptr());
+    }
+  }
+  I.resize(i);
```
skimo-openhub: Please use `I.resize(inputs.size())` instead.
If you are really worried about recomputing this value, then compute it up front and
store it in a variable that is also used in the for loop condition.

tc/aten/aten_compiler_new_api-inl.h
nicolasvasilache: ugh absolutely, I ended up just copying this, cleaning up, thanks!

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,137 @@
...
+namespace aten {
+inline std::vector<DLTensorUPtr> inferOutputTensorInfo(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs) {
+  auto parsedTcs = tc::detail::parse(tc);
+  if (parsedTcs.count(entryPoint) != 1u) {
+    CHECK_GE(parsedTcs.size(), 1u)
+        << "No TC was parsed, should have thrown earlier";
+    throw lang::ErrorReport(parsedTcs.begin()->second)
```
ftynse: not sure where this report will going to point in the code

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,137 @@
...
+        << "\nattempting to access undefined entryPoint: " << entryPoint;
+  }
+  auto inputDLTensors = toDLConstTensors(inputs);
+  return makeDLTensorVector(detail::inferOutputTensorInfo(
+      parsedTcs.at(entryPoint), extractRawPtrs(inputDLTensors)));
+}
+
+inline std::vector<at::Tensor> prepareOutputs(
```
ftynse: why do we want to inline these relatively long functions?

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,137 @@
...
+  auto parsedTcs = tc::detail::parse(tc);
+  if (parsedTcs.count(entryPoint) != 1u) {
+    CHECK_GE(parsedTcs.size(), 1u)
+        << "No TC was parsed, should have thrown earlier";
+    throw lang::ErrorReport(parsedTcs.begin()->second)
+        << "\nattempting to access undefined entryPoint: " << entryPoint;
+  }
+
+  std::vector<at::Tensor> outputs;
+  auto inputDLTensors = toDLConstTensors(inputs);
```
ftynse: until this point, this is a copy-paste from `inferOutputTensorInfo`, extract some internal function to avoid code duplication

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,137 @@
...
+  auto outTensorInfo = detail::inferOutputTensorInfo(
+      parsedTcs.at(entryPoint), extractRawPtrs(inputDLTensors));
+  if (outTensorInfo.size() == 0) {
+    return outputs;
+  }
+  CHECK_GE(inputs.size(), 1u)
```
ftynse: feels like an unhandled edge case; add "NYI" for now, and later an overload taking an ATen backed instead of a list of tensors

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,137 @@
...
+      tc, entryPoint, extractRawPtrs(inputDLTensors), options);
+}
+
+template <typename Executor>
+void run(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs) {
+  auto inputDLTensors = toDLConstTensors(inputs);
+  auto outputDLTensors = makeDLTensors(outputs);
```
ftynse: I just noticed this name duality between toDLConstTensors and makeDLTensors; looks confusing when used together

tc/aten/aten_compiler_new_api-inl.h
```diff
@@ -0,0 +1,137 @@
...
+}
+
+template <typename Executor>
+void uncheckedRun(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs) {
+  CHECK_GE(outputs.size(), 1u);
+  std::vector<const void*> I(inputs.size(), nullptr);
+  std::vector<void*> O(outputs.size(), nullptr);
```
ftynse: not really a fan of these variable names

tc/aten/aten_compiler_new_api.h
```diff
@@ -0,0 +1,89 @@
...
+/// JIT-compiled callable function.
+template <typename Backend>
+std::unique_ptr<typename Backend::ExecutorType> compile(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::vector<at::Tensor>& inputs,
+    const typename Backend::MappingOptionsType& options);
+
+/// Given an executor resulting from compiling a TC, run the TC and fill the
+/// outputs vector with the results.
```
ftynse: maybe mention the output vector must have as many elements as output tensors in the given TC and that those tensors must be allocated;  "fill the outputs vector" is ambiguous enough for someone to assume it will do the allocation

tc/aten/aten_compiler_new_api.h
```diff
@@ -0,0 +1,89 @@
...
+template <typename Executor>
+void run(
+    const Executor& executor,
+    const std::vector<at::Tensor>& inputs,
+    std::vector<at::Tensor>& outputs);
+
+/// Given an executor resulting from compiling a TC, run the TC and fill the
+/// outputs vector with the results.
+/// \returns ProfilingInfo (cpuOverhead + kernelRuntime) in microseconds.
```
ftynse: let's not use "+" here, confusing given that fields of ProfilingInfo are numbers totalTime = cpuOverhead + kernelRuntime

nicolasvasilache:
> this PR seems to update isl for no reason

hmm, good catch, thanks!

tc/aten/aten_compiler_new_api-inl.h
nicolasvasilache: or just use the function above instead of the `detail::` version

tc/aten/aten_compiler_new_api-inl.h
nicolasvasilache: yeah I have that on the stacked PR, backporting

tc/aten/aten_compiler_new_api-inl.h
nicolasvasilache: was trying to avoid creating yet another library (tc_aten) that needs to be linked in proper places (tests etc). Now that I stuck my finger in this, it involves extra cmake genuflexions, oh well..

tc/aten/aten_compiler_new_api.h
nicolasvasilache: done

nicolasvasilache:

PR #363, opened 2018-04-26
[C++ API] Step 5: add the new TcExecutor implementations

nicolasvasilache:
This PR adds the files implementing the TcExecutor API (#323) on
top of the new compiler API (#322). The ExecutionEngine has been retired
and the TcExecutor reflects this.

TcExecutor is a backend-agnostic abstraction that is the result of compilation
(see #322). When a concrete executor specializes TcExecutor it provides support
for running a compiled TC on the particular backend.

Specializing a TcExecutor consists in implementing (not overriding) 2 methods:
1. uncheckedRun provides the low-latency path on void* data
2. setupRuntimeCompiledFunction takes the result of compileWithTcMapper to:
  a. save backend-specific information required at runtime (e.g. grid and block
     sizes for the CUDA backend)
  b. create the in-memory object resulting from JIT-compilation that implements
     the TC

Note that the CPU implementation is currently just a stub that only forcibly
prints to screen. This allows a simple sanity check in the branch
`pr/big-refactor` that compiles without CUDA and prints the proper steps.
Implementation for the CPU will come once the refactoring has landed.

The reason the consensus went for implementing (and not overriding) in #323 is
that mixing CRTP at a distance via the Backend type with inheritance was
confusing. More specifically:

```
Since people may get confused by mixing CRTP and virtual functions I am fine
using the type of static dispatch @ftynse propose in order to remove virtual
functions:

static_cast<Backend::ExecutorType*>(this)->uncheckedRun();

The tradeoff is that one will get error at the call sites with much more
complex error messages than with virtual functions.
I am personally more favorable to virtual functions (original proposal)
because error messages will pop up at the definition site in the abstract
base class but I can also see that mixing the two makes it more difficult
to understand what happens.
```

This new implementation will be switched in as part of the last commit of the global refactoring.

### DIFF
--- a/tc/core/cpu/cpu_tc_executor_new_api.cc
+++ b/tc/core/cpu/cpu_tc_executor_new_api.cc
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "tc/core/cpu/cpu_tc_executor.h"
+
+#include "tc/core/cpu/cpu_mapping_options.h"
+#include "tc/core/cpu/cpu_mapping_options_cpp_printer.h"
+#include "tc/core/halide_utils.h"
+#include "tc/core/tc2halide.h"
+#include "tc/core/tensor.h"
+#include "tc/lang/parser.h"
+#include "tc/lang/sema.h"
+#include "version.h"
+
+namespace tc {
+CpuTcExecutor::CpuTcExecutor(
+    const std::vector<TensorInfo>& inputsInfo,
+    const std::vector<TensorInfo>& outputsInfo,
+    const tc2halide::HalideComponents& halideComponents,
+    const typename CpuBackend::CompilationResultType& compilationResult)
+    : TcExecutor<CpuBackend>(
+          inputsInfo,
+          outputsInfo,
+          halideComponents,
+          compilationResult) {
+  LOG(ERROR) << "NYI: CpuTcExecutor::CpuTcExecutor setup RTC";
+}
+
+CpuCompilationResult CpuBackend::compileWithTcMapper(
+    const std::string& tcName,
+    tc2halide::HalideComponents halideComponents,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const CpuMappingOptions& options) {
+  LOG(ERROR) << "NYI: CpuBackend::compileWithTcMapper";
+  return CpuCompilationResult{std::string("source"),
+                              std::string("specializedName"),
+                              std::vector<long>()};
+}
+
+void CpuTcExecutor::uncheckedRun(
+    const std::vector<const void*>& inputs,
+    const std::vector<void*>& outputs) const {
+  LOG(ERROR) << "NYI: CpuTcExecutor::uncheckedRun";
+}
+
+ProfilingInfo CpuTcExecutor::profileUnchecked(
+    const std::vector<const void*>& inputs,
+    const std::vector<void*>& outputs) const {
+  LOG(ERROR) << "NYI: CpuTcExecutor::profileUnchecked";
+  return ProfilingInfo{
+      Duration(std::chrono::microseconds(static_cast<int64_t>(999999999999))),
+      Duration(std::chrono::microseconds(static_cast<int64_t>(999999999999)))};
+}
+} // namespace tc
--- a/tc/core/cpu/cpu_tc_executor_new_api.h
+++ b/tc/core/cpu/cpu_tc_executor_new_api.h
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "tc/core/cpu/cpu_backend.h"
+#include "tc/core/cpu/cpu_mapping_options.h"
+#include "tc/core/tc_executor.h"
+#include "tc/core/tensor.h"
+
+namespace tc {
+class CpuTcExecutor : public TcExecutor<CpuBackend> {
+ public:
+  CpuTcExecutor(
+      const std::vector<TensorInfo>& inputsInfo,
+      const std::vector<TensorInfo>& outputsInfo,
+      const tc2halide::HalideComponents& halideComponents,
+      const typename CpuBackend::CompilationResultType& compilationResult);
+
+  /// This is the "low-latency" mode in which we just propagate raw pointers to
+  /// data in the address space where kernel is executed.
+  /// No tensor-related information can be checked so it is the user's
+  /// responsibility to ensure that shapes and strides match. If the user
+  /// doesn't then segfault will likely occur.
+  void uncheckedRun(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const;
+
+  /// Calls uncheckedRun and profiles the cpu overhead and kernel runtime
+  /// (microseconds).
+  /// \returns profiling information
+  ProfilingInfo profileUnchecked(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const;
+};
+} // namespace tc
--- a/tc/core/cuda/cuda_tc_executor_new_api.cc
+++ b/tc/core/cuda/cuda_tc_executor_new_api.cc
@@ -1,0 +1,152 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "tc/core/cuda/cuda_tc_executor.h"
+
+#include "tc/core/cuda/cuda_mapping_options_cpp_printer.h"
+#include "tc/core/halide_utils.h"
+#include "tc/core/polyhedral/cuda/mapped_scop.h"
+#include "tc/core/tc2halide.h"
+#include "tc/core/tensor.h"
+
+#include "tc/lang/parser.h"
+#include "tc/lang/sema.h"
+
+#include <utility>
+
+namespace tc {
+namespace {
+// Append ordered values to the kernel name, separated by "_".
+template <typename T>
+std::string specializeKernelName(
+    const std::string& tcName,
+    std::vector<T> params) {
+  std::stringstream ss;
+  ss << tcName;
+  for (auto i : params) {
+    ss << "_" << i;
+  }
+  return ss.str();
+}
+} // namespace
+
+CudaTcExecutor::CudaTcExecutor(
+    const std::vector<TensorInfo>& inputsInfo,
+    const std::vector<TensorInfo>& outputsInfo,
+    const tc2halide::HalideComponents& halideComponents,
+    const typename CudaBackend::CompilationResultType& compilationResult)
+    : TcExecutor<CudaBackend>(
+          inputsInfo,
+          outputsInfo,
+          halideComponents,
+          compilationResult) {
+  auto t0 = std::chrono::high_resolution_clock::now();
+  // force unloading in case we JIT with the same name/input/outputs with
+  // different options.
+  this->clearRuntimeCompiledFunction();
+  rtcFun_ = CudaRTCFunction::Compile(
+      compilationResult.specializedName, compilationResult.source);
+  grid_ = compilationResult.grid;
+  block_ = compilationResult.block;
+  auto t1 = std::chrono::high_resolution_clock::now();
+  LOG_IF(INFO, FLAGS_debug_tc_mapper)
+      << "[COMPILE] Compiling with host JIT compiler took: "
+      << std::chrono::duration_cast<std::chrono::milliseconds>(t1 - t0).count()
+      << "ms" << std::endl;
+}
+
+CudaCompilationResult CudaBackend::compileWithTcMapper(
+    const std::string& tcName,
+    tc2halide::HalideComponents halideComponents,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const CudaMappingOptions& options) {
+  // A bit chicken-and-eggy, need scop from TC to have the space to build the
+  // context to specialize the scop..
+  auto scop = polyhedral::Scop::makeScop(
+      isl::with_exceptions::globalIslCtx(), halideComponents);
+  auto globalParameterContext = scop->makeContextFromInputs(inputs);
+  scop = polyhedral::Scop::makeSpecializedScop(
+      *scop, globalParameterContext.intersect(scop->globalParameterContext));
+  LOG_IF(INFO, FLAGS_debug_tc_mapper) << options;
+  LOG_IF(INFO, FLAGS_debug_tc_mapper) << "original schedule:\n"
+                                      << *(scop->scheduleRoot());
+
+  // Now we can build stuff
+  auto mappedScop =
+      polyhedral::MappedScop::makeWithOuterBlockInnerThreadStrategy(
+          std::move(scop), options);
+  LOG_IF(INFO, FLAGS_debug_tc_mapper) << "Mapped schedule:" << std::endl
+                                      << *(mappedScop->schedule());
+
+  auto parameters =
+      mappedScop->scop().getParameterValues(globalParameterContext);
+  auto specializedName = specializeKernelName(tcName, parameters);
+
+  // This updates the launch bounds with the actual result from compilation
+  // with tightening of launch_bounds. What you get is not necessarily what
+  // you asked for, the autotuner should adapt to that.
+  std::string source;
+  Grid grid;
+  Block block;
+  std::tie(source, grid, block) = mappedScop->codegen(specializedName);
+  LOG_IF(INFO, FLAGS_dump_cuda) << "generatedCuda: " << source << "\n"
+                                << "grid: " << grid << " block: " << block;
+
+  return CudaCompilationResult{
+      source, specializedName, parameters, grid, block};
+}
+
+void CudaTcExecutor::uncheckedRun(
+    const std::vector<const void*>& inputs,
+    const std::vector<void*>& outputs) const {
+  CHECK(rtcFun_) << "No rtcFun_ attached, cannot launch";
+  cudaStream_t stream = 0;
+  CHECK_NE(grid_.view[0], 0u) << "Grid dims are not set up";
+  CHECK_NE(block_.view[0], 0u) << "Block dims are not set up";
+  rtcFun_->Launch(
+      grid_.view.extractDefaultedArray(),
+      block_.view.extractDefaultedArray(),
+      0,
+      stream,
+      parameters_,
+      outputs,
+      inputs);
+}
+
+ProfilingInfo CudaTcExecutor::profileUnchecked(
+    const std::vector<const void*>& inputs,
+    const std::vector<void*>& outputs) const {
+  auto start = std::chrono::system_clock::now();
+  CHECK(rtcFun_) << "No rtcFun_ attached, cannot launch";
+  cudaStream_t stream = 0;
+  CHECK_NE(grid_.view[0], 0u) << "Grid dims are not set up";
+  CHECK_NE(block_.view[0], 0u) << "Block dims are not set up";
+  Duration kernelRuntime(rtcFun_->Launch(
+      grid_.view.extractDefaultedArray(),
+      block_.view.extractDefaultedArray(),
+      0,
+      stream,
+      parameters_,
+      outputs,
+      inputs,
+      true));
+  // The CPU overhead is the total time minus the (synchronized) kernel runtime
+  auto end = std::chrono::system_clock::now();
+  Duration cpuOverhead(end - start);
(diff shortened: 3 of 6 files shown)


tc/core/cpu/cpu_tc_executor_new_api.h
```diff
@@ -0,0 +1,51 @@
...
+  /// This is the "low-latency" mode in which we just propagate raw pointers to
+  /// data in the address space where kernel is executed.
+  /// No tensor-related information can be checked so it is the user's
+  /// responsibility to ensure that shapes and strides match. If the user
+  /// doesn't then segfault will likely occur.
+  void uncheckedRun(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const;
+
+  /// Calls uncheckedRun and profiles the cpu overheal and kernel runtime
```
skimo-openhub: overhead

tc/core/cuda/cuda_tc_executor_new_api.h
```diff
@@ -0,0 +1,66 @@
...
+  /// This is the "low-latency" mode in which we just propagate raw pointers to
+  /// data in the address space where kernel is executed.
+  /// No tensor-related information can be checked so it is the user's
+  /// responsibility to ensure that shapes and strides match. If the user
+  /// doesn't then segfault will likely occur.
+  void uncheckedRun(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const;
+
+  /// Calls uncheckedRun and profiles the cpu overheal and kernel runtime
```
skimo-openhub: overhead

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * /// This is the "low-latency" mode in which we just propagate raw pointers to
+ * /// data in the address space where kernel is executed.
+ * /// No tensor-related information can be checked so it is the user's
+ * /// responsibility to ensure that shapes and strides match. If the user
+ * /// doesn't then segfault will likely occur.
+ * void uncheckedRun(
+ *     const std::vector<const void*>& inputs,
+ *     const std::vector<void*>& outputs) const;
+ * /// Calls uncheckedRun and profiles the cpu overheal and kernel runtime
```
skimo-openhub: overhead

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+  /// Sizes and strides must remain constant otherwise this is an error.
+  /// The only thing that is allowed to change across runs is the input
+  /// and output pointers base address.
+  /// It is the caller's responsibility to ensure proper non-aliasing (or
+  /// advanced aliasing) properties of the input and output tensors.
+  void run(
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs) const;
+
+  /// Calls run and profiles the cpu overheal and kernel runtime (microseconds).
```
skimo-openhub: overhead

tc/core/tc_executor_new_api-inl.h
```diff
@@ -0,0 +1,148 @@
+/**
```
skimo-openhub: This file doesn't appear to be included from anywhere.

tc/core/tc_executor_new_api-inl.h
```diff
@@ -0,0 +1,148 @@
...
+                                   << actual->ndim << " dimensions";
+    }
+    auto atype = toTypeToken(actual->dtype);
+    auto etype = toTypeToken(expected.dtype);
+    if (atype != etype) {
+      throw lang::ErrorReport(dbg)
+          << "expected " << lang::kindToString(etype) << " but found "
+          << lang::kindToString(atype);
+    }
+    std::vector<int64_t> shapeA(actual->shape, actual->shape + actual->ndim);
```
skimo-openhub: why do you need to copy the data into a vector?

tc/core/tc_executor_new_api-inl.h
```diff
@@ -0,0 +1,148 @@
...
+  ProfilingInfo pi(
+      // Static dispatch instead of virtual functions requires this cast.
+      static_cast<const typename Backend::ExecutorType&>(*this)
+          .profileUnchecked(I, O));
+
+  // The total CPU overhead is the total time minus the (synchronized) kernel
+  // runtime
+  auto end = std::chrono::system_clock::now();
+  Duration cpuOverhead(end - start);
+  cpuOverhead = cpuOverhead - pi.kernelRuntime;
```
skimo-openhub: Isn't `cpuOverhead` already computed by `CudaTcExecutor::profileUnchecked`?

skimo-openhub:
The commit message says that people may find inheritance confusing.
Could you add a note to explain why people might find that confusing?

nicolasvasilache:
> The commit message says that people may find inheritance confusing.
> Could you add a note to explain why people might find that confusing?

People have found it confusing, I am myself not completely sure why but mixing templates + inheritance wasn't welcome. @ttheodor care to propose a sentence to explain why that I can put in the commit message as @skimo-openhub  suggests?

tc/core/cpu/cpu_tc_executor_new_api.cc
```diff
@@ -0,0 +1,69 @@
...
+  LOG(ERROR) << "NYI: CpuTcExecutor::uncheckedRun";
+}
+
+ProfilingInfo CpuTcExecutor::profileUnchecked(
+    const std::vector<const void*>& inputs,
+    const std::vector<void*>& outputs) const {
+  LOG(ERROR) << "NYI: CpuTcExecutor::profileUnchecked";
+  return ProfilingInfo{
+      Duration(std::chrono::microseconds(static_cast<int64_t>(1000))),
+      Duration(std::chrono::microseconds(static_cast<int64_t>(1000)))};
```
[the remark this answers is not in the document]
nicolasvasilache: 0 wouldn't pass some checks, I find 1000 visible enough when glancing through LOGs.
Do you have another value that would be better?

tc/core/tc_executor_new_api-inl.h
nicolasvasilache: It is but it is also missing the extra CPU overhead from this function.

tc/core/tc_executor_new_api-inl.h
nicolasvasilache: Files are living side-by-side and I will run git mv -f as part of the last commit.

tc/core/cpu/cpu_tc_executor_new_api.cc
```diff
@@ -0,0 +1,69 @@
...
+#include "tc/core/cpu/cpu_mapping_options.h"
+#include "tc/core/cpu/cpu_mapping_options_cpp_printer.h"
+#include "tc/core/halide_utils.h"
+#include "tc/core/tc2halide.h"
+#include "tc/core/tensor.h"
+
+#include "tc/lang/parser.h"
+#include "tc/lang/sema.h"
+
+#include <version.h>
```
ftynse: why does this use angle brackets?

tc/core/cpu/cpu_tc_executor_new_api.cc
```diff
@@ -0,0 +1,69 @@
...
+CpuCompilationResult CpuBackend::compileWithTcMapper(
+    const std::string& tcName,
+    tc2halide::HalideComponents halideComponents,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const CpuMappingOptions& options) {
+  LOG(ERROR) << "NYI: CpuBackend::compileWithTcMapper";
+  return CpuCompilationResult{std::string("source"),
+                              std::string("specializedName"),
+                              std::vector<int>{}};
```
ftynse: mix of old-style and uniform initialization

tc/core/cpu/cpu_tc_executor_new_api.cc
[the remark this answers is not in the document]
ftynse: `0xFACEB00C` :P 
1000 looks like a legit execution time, something huge wouldn't

tc/core/cpu/cpu_tc_executor_new_api.h
```diff
@@ -0,0 +1,51 @@
...
+  /// No tensor-related information can be checked so it is the user's
+  /// responsibility to ensure that shapes and strides match. If the user
+  /// doesn't then segfault will likely occur.
+  void uncheckedRun(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const;
+
+  /// Calls uncheckedRun and profiles the cpu overhead and kernel runtime
+  /// (microseconds).
+  /// \returns profiling information (see: tc/core/utils/time.h)
```
ftynse: I wouldn't add a path to the comments.  It is likely to change without documentation being updated. Doxygen is smart enough to pick up the type (or use \ref typename to force it) and add a hyperlink to the documentation of the relevant class.

tc/core/cuda/cuda_tc_executor_new_api.cc
```diff
@@ -0,0 +1,163 @@
...
+
+CudaCompilationResult CudaBackend::compileWithTcMapper(
+    const std::string& tcName,
+    tc2halide::HalideComponents halideComponents,
+    const std::vector<const DLConstTensor*>& inputs,
+    /* TODO: in the future also pass outputs for stride and alignment info */
+    const CudaMappingOptions& options) {
+  // A bit chicken-and-eggy, need scop from TC to have the space to build the
+  // context to specialize the scop..
+  auto scopTmp = polyhedral::Scop::makeScop(
```
ftynse: why say this scop is temporary in its name?

tc/core/cuda/cuda_tc_executor_new_api.cc
```diff
@@ -0,0 +1,163 @@
...
+      isl::with_exceptions::globalIslCtx(), halideComponents);
+  auto globalParameterContext = scopTmp->makeContextFromInputs(inputs);
+  scopTmp = polyhedral::Scop::makeSpecializedScop(
+      *scopTmp,
+      globalParameterContext.intersect(scopTmp->globalParameterContext));
+  LOG_IF(INFO, FLAGS_debug_tc_mapper) << options;
+  LOG_IF(INFO, FLAGS_debug_tc_mapper) << *(scopTmp->scheduleRoot());
```
ftynse: some marker also print some marker above this, like "original schedule" ?

tc/core/cuda/cuda_tc_executor_new_api.cc
```diff
@@ -0,0 +1,163 @@
...
+
+  // Now we can build stuff
+  auto mappedScop =
+      polyhedral::MappedScop::makeWithOuterBlockInnerThreadStrategy(
+          std::move(scopTmp), options);
+  LOG_IF(INFO, FLAGS_debug_tc_mapper) << "Mapped schedule:" << std::endl
+                                      << *(mappedScop->schedule());
+
+  auto parameters = narrowParamsVector(
```
ftynse: the call to narrowParamsVector (and the function itself since it has only once call place) is only necessary because specializeKernelName takes a vector of ints with the parameter values.  I don't see why, it looks like it was needed to silence a warning.  Let's make specializeKernelName accept longs, or a template argument type, and remove the narrowing completely.

tc/core/cuda/cuda_tc_executor_new_api.cc
```diff
@@ -0,0 +1,163 @@
...
+      mappedScop->scop().getParameterValues(globalParameterContext));
+  auto specializedName = specializeKernelName(tcName, parameters);
+
+  // This updates the launch bounds with the actual result from compilation
+  // with tightening of launch_bounds.
+  // What you get is not what you asked for, the autotuner should adapt to
```
ftynse: is not _necessarily_, it may be what you asked for

tc/core/cuda/cuda_tc_executor_new_api.h
```diff
@@ -0,0 +1,66 @@
...
+      const std::vector<void*>& outputs) const;
+
+  /// Calls uncheckedRun and profiles the cpu overhead and kernel runtime
+  /// (microseconds).
+  /// \returns profiling information (see: tc/core/utils/time.h)
+  ProfilingInfo profileUnchecked(
+      const std::vector<const void*>& inputs,
+      const std::vector<void*>& outputs) const;
+
+  // GPU-specific results of compilation
```
ftynse: either add a more descriptive comment, or remove this one

tc/core/tc_executor_new_api-inl.h
```diff
@@ -0,0 +1,147 @@
...
+    if (static_cast<size_t>(actual->ndim) != expected.shape.size()) {
+      throw lang::ErrorReport(dbg) << "expected " << expected.shape.size()
+                                   << " dimensions but found tensor with "
+                                   << actual->ndim << " dimensions";
+    }
+    auto atype = toTypeToken(actual->dtype);
+    auto etype = toTypeToken(expected.dtype);
+    if (atype != etype) {
+      throw lang::ErrorReport(dbg)
+          << "expected " << lang::kindToString(etype) << " but found "
```
ftynse: maybe add "for dim i" here as well?

tc/core/tc_executor_new_api-inl.h
```diff
@@ -0,0 +1,147 @@
...
+      outputs, outputsInfo_, halideComponents_.getDef().returns());
+
+  std::vector<const void*> I;
+  std::vector<void*> O;
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    I.push_back(inputs[i]->data);
+  }
+  for (size_t i = 0; i < outputs.size(); ++i) {
+    O.push_back(outputs[i]->data);
+  }
```
ftynse: copy-paste from run...

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+ * TcExecutor is a backend-agnostic abstraction that provides the base
+ * functionality for an object returned by compilation to run.
+ * TcExecutor is templated by the Backend type.
+ * When a derived executor inherits from TcExecutor<Backend>, it
+ * provides support for running a compiled TC on the particular backend.
+ *
+ * A TcExecutor mixes templating and inheritance:
+ *   1. templating is necessary because we want type-safety with proper
+ *      dependent types (e.g. Backend::CompilationResultType)
+ *   2. inheritance is necessary because executors require specific
```
ftynse: I don't understand this motivation.  Inheritance is necessary when all derived
types need to have some _common_ information.  This comment makes it feel like
they need different information, so there would be no point for inheritance.

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+ *      additional information for running on different backends
+ *      (e.g. Grid/Block for CUDA)
+ *   3. virtual functions seem to be more confusing to people so we don't use
```
ftynse: virtual functions _combined with templates_

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+ *      them here. Still derived classes of TcExecutor<Backend> **must**
+ *      implement the following functions (or errors will occur at **all call
+ *      sites**):
+ *
+ * /// This is the "low-latency" mode in which we just propagate raw pointers to
```
ftynse: Why do we copy the documentation comment into another comment?  It may confuse
doxyxgen and is likely to be forgotten in future updates to the function.

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+      const tc2halide::HalideComponents& halideComponents,
+      const typename Backend::CompilationResultType& compilationResult);
+
+ public:
+  TcExecutor(TcExecutor&&) = delete;
+  TcExecutor& operator=(TcExecutor&&) = delete;
+  TcExecutor(const TcExecutor&) = delete;
+  TcExecutor& operator=(const TcExecutor&) = delete;
+
+  /// Run can be called multiple times given a compilation, inputs are allowed
```
ftynse: "given a compilation" sounds awkward

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+  /// The only thing that is allowed to change across runs is the input
+  /// and output pointers base address.
+  /// It is the caller's responsibility to ensure proper non-aliasing (or
+  /// advanced aliasing) properties of the input and output tensors.
+  void run(
+      const std::vector<const DLConstTensor*>& inputs,
+      const std::vector<const DLTensor*>& outputs) const;
+
+  /// Calls run and profiles the cpu overhead and kernel runtime (microseconds).
+  /// \returns profiling information (see: tc/core/utils/time.h)
```
ftynse: see my other comment about hardcoding paths

tc/core/tc_executor_new_api.h
```diff
@@ -0,0 +1,125 @@
...
+  /// It may be necessary to clear the RTC manually because it can throw and
+  /// we can't have that in the RTC destructor.
+  void clearRuntimeCompiledFunction();
+
+ public:
+  // For inspection purposes
+  const std::string compiledSource;
+
+ protected:
+  /// Used to check proper metadata when calling run
```
ftynse: Doxygen has groups (`@{` / `@}` or `\{` / `\}`) that make the comment apply to multiple fields, like in this case, rather than to the first one.

ftynse:
> This PR adds the files implementing the TcExecutor API (#323) on
top of the new compiler API (#322). The ExecutionEngine has been retired
and the TcExecutor reflects this.

> TcExecutor is a backend-agnostic abstraction that is the result of compilation
(see #322). When a concrete executor specializes TcExecutor it provides support

AFAIK, dependent clauses related to time and condition (introduced by "when" and "if" need a comma).  

> There are several other places in different commit messages.
for running a compiled TC on the particular backend.

> Specializing a TcExecutor consists in implementing (not overriding) 2 methods:
> 1. uncheckedRun provides the low-latency path on void* data
> 2. setupRuntimeCompiledFunction takes the result of compileWithTcMapper to:
>  a. save backend-specific information required at runtime (e.g. grid and block
>     sizes for the CUDA backend)
>  b. create the in-memory object resulting from JIT-compilation that implements
>     the TC

> Note that the CPU implementation is currently just a stub that only forcibly
prints to screen. This allows a simple sanity check in the branch
```pr/big-refactor``` that compiles without CUDA and prints the proper steps.

let's not use _three_ backticks at all; and use quotes when names must be quoted rather than backticks which imply code

> Implementation for the CPU will come once the refactoring has landed.

> The reason the consensus went for implementing (and not overriding) in #323 is
that mising CRTP at a distance via the Backend type with inheritance was

"missing"

I did not really understand this passage about "missing CRTP".  How about something like "the initial proposal resembled a CRTP except it used virtual member function overriding for polymorphic behavior.  A combination of static (CRTP) and dynamic (overriding virtual member functions) polymorphism was confusing." ?

> confusing. More specifically:

> "```"

let's not use triple backticks, especially for prose (not code)

> Since people have been confused by mixing CRTP and virtual functions I am fine
using the type of static dispatch @ftynse propose in order to remove virtual

proposed/proposes

> functions:

> `static_cast<Backend::ExecutorType*>(this)->uncheckedRun();`

> The tradeoff is that one will get error at the call sites with much more
complex error messages than with virtual functions.
I am personally more favorable to virtual functions (original proposal)
because error messages will pop up at the definition site in the abstract
base class but I can also see that mixing the two makes it more difficult
to understand what happens.
> "```"
the global refactoring.

double "the"

**Comment by nicolasvasilache** [comment]
> "missing"
> I did not really understand this passage about "missing CRTP"

typo should be "mixing"

**Comment by skimo-openhub** [comment]
On Mon, Apr 30, 2018 at 01:03:30PM -0700, Nicolas Vasilache wrote:
> > +  }
> +  for (size_t i = 0; i < outputs.size(); ++i) {
> +    O.push_back(outputs[i]->data);
> +  }
> +
> +  // Launch kernel and get **the kernel** time (without CPU overhead)
> +  ProfilingInfo pi(
> +      // Static dispatch instead of virtual functions requires this cast.
> +      static_cast<const typename Backend::ExecutorType&>(*this)
> +          .profileUnchecked(I, O));
> +
> +  // The total CPU overhead is the total time minus the (synchronized) kernel
> +  // runtime
> +  auto end = std::chrono::system_clock::now();
> +  Duration cpuOverhead(end - start);
> +  cpuOverhead = cpuOverhead - pi.kernelRuntime;
> 
> It is but it is also missing the extra CPU overhead from this function.

OK, but perhaps the duplicate code could be put in a helper function.
It may also be easier to let the caller compute the overhead.

skimo

**Comment by skimo-openhub** [comment]
On Mon, Apr 30, 2018 at 12:58:41PM -0700, Nicolas Vasilache wrote:
> > +                              std::vector<int>{}};
> +}
> +
> +void CpuTcExecutor::uncheckedRun(
> +    const std::vector<const void*>& inputs,
> +    const std::vector<void*>& outputs) const {
> +  LOG(ERROR) << "NYI: CpuTcExecutor::uncheckedRun";
> +}
> +
> +ProfilingInfo CpuTcExecutor::profileUnchecked(
> +    const std::vector<const void*>& inputs,
> +    const std::vector<void*>& outputs) const {
> +  LOG(ERROR) << "NYI: CpuTcExecutor::profileUnchecked";
> +  return ProfilingInfo{
> +      Duration(std::chrono::microseconds(static_cast<int64_t>(1000))),
> +      Duration(std::chrono::microseconds(static_cast<int64_t>(1000)))};
> 
> 0 wouldn't pass some checks, I find 1000 visible enough when glancing through LOGs.
> Do you have another value that would be better?

Which checks fail?  It seems to run fine over here with 0.

skimo

**Comment by nicolasvasilache** [comment]
> which checks fails

We have a sanity check in autotuner, if you run:
```
build$ make -j 16 && tc/examples/tensordot --tuner_threads=1
--tuner_gen_pop_size=1 --tuner_gen_generations=1 --tuner_gen_number_elites=0
--debug_tuner=1 --gtest_filter="*CPU*"
```

on #379 then you'll see this:

```
F0502 06:54:54.148926 30274 autotuner-inl.h:237] The measured runtime is 0, marking as invalid: Runtimes:
tc::MappingOptions::makeNaiveMappingOptions()
    .outerScheduleFusionStrategy(tc::FusionStrategy::Preserve3Coincident)
    .outerScheduleAllowSkewing(false)
    .outerSchedulePositiveOrthant(true)
    .intraTileScheduleFusionStrategy(tc::FusionStrategy::Preserve3Coincident)
    .intraTileScheduleAllowSkewing(false)
    .intraTileSchedulePositiveOrthant(true)
    .tile(32, 32, 32)
    .unroll(1)
    .tileImperfectlyNested(false)
    .matchLibraryCalls(false);
*** Check failure stack trace: ***
    @     0x7f3458f6180d  google::LogMessage::Fail()
    @     0x7f3458f63623  google::LogMessage::SendToLog()
    @     0x7f3458f612fb  google::LogMessage::Flush()
    @     0x7f3458f6400e  google::LogMessageFatal::~LogMessageFatal()
    @           0x606dc9  tc::autotune::detail::TuningHarness<>::doEvaluate()
    @           0x5fb0a7  _ZZN2tc8autotune6detail13TuningHarnessINS_10CpuBackendEE15runOneIterationINS0_13GeneticSearchEEEvRT_mENKUlvE3_clEv
    @           0x635090  _ZNSt12_Bind_simpleIFZN2tc8autotune6detail13TuningHarnessINS0_10CpuBackendEE15runOneIterationINS1_13GeneticSearchEEEvRT_mEUlvE3_vEE9_M_invokeIJEEEvSt12_Index_tupleIJXspT_EEE
    @           0x634cd4  _ZNSt12_Bind_simpleIFZN2tc8autotune6detail13TuningHarnessINS0_10CpuBackendEE15runOneIterationINS1_13GeneticSearchEEEvRT_mEUlvE3_vEEclEv
    @           0x6345d8  _ZNSt6thread5_ImplISt12_Bind_simpleIFZN2tc8autotune6detail13TuningHarnessINS2_10CpuBackendEE15runOneIterationINS3_13GeneticSearchEEEvRT_mEUlvE3_vEEE6_M_runEv
    @     0x7f34440ac510  (unknown)
    @     0x7f34438ba6ba  start_thread
    @     0x7f34435f041d  clone
    @              (nil)  (unknown)
0 0 0 0 0 0 0 0 0 0 Aborted
```

nicolasvasilache:
> perhaps the duplicate code could be put in a helper function.

done last night, see detail::prepareRun. 
It is inlined to avoid introducing another .cc file with a single small function.

nicolasvasilache:

PR #365, opened 2018-04-26
[C++ API] Step 6: add the new OptionsCache

nicolasvasilache:
This PR adds the files implementing the OptionsCache API (#339).
Options cache is used only in the autotuner as a consequence it is moved in the
proper directory. Upon further inspection and rewriting of the autotuner logic
I found the cache implementation was convoluted by virtue of trying to do
too much factoring out with the CRTP-based cache used for compilation.

This is a simpler implementation that uses the same key/value representation
in both proto and memory (for collision avoidance, the proto is not submitted
in this PR and will land at the end).
A lightweight wrapper with conversion to/from proto is provided as usual because
the proto API doesn't hold a candle to STL.
The implementation is currently using ```std::unordered_multimap``` there are
some inefficiencies related to linear search and duplication of keys in the
proto. These can go away after refactoring #307 is completed along the lines of:
1. switch to proto3 which has support for map
2. do better than linear search

The above should be backed by data: case 1 will likely happen when we start
having larger options files but the benefits should be a factor 2 at most
(see comment in proto in #339); case 2 should not be a problem unless we start
storing much more than 10 options per TC. In any case I do not see anything
suspicious when running the tuner locally.

the global refactoring.

### DIFF
--- a/tc/autotuner/options_cache-inl.h
+++ b/tc/autotuner/options_cache-inl.h
@@ -1,0 +1,342 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cstdint>
+#include <fstream>
+#include <memory>
+#include <mutex>
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include <llvm/ADT/Optional.h>
+
+#include "tc/core/tensor.h"
+#include "tc/core/utils/math.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/canonicalize.h"
+
+#include "version.h"
+
+namespace tc {
+namespace autotune {
+
+bool OptionsCacheKey::operator==(const OptionsCacheKey& other) const {
+  if (id != other.id) {
+    return false;
+  }
+  if (deviceStr != other.deviceStr) {
+    return false;
+  }
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    if (inputs[i] != other.inputs[i]) {
+      return false;
+    }
+  }
+  for (size_t i = 0; i < outputs.size(); ++i) {
+    if (not(outputs[i] == other.outputs[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool OptionsCacheKey::operator!=(const OptionsCacheKey& other) const {
+  return !(*this == other);
+}
+
+OptionsCacheKeyProto OptionsCacheKey::toProtobuf() const {
+  OptionsCacheKeyProto bufKey;
+  bufKey.set_id(id);
+  for (const auto& in : inputs) {
+    auto pin = bufKey.add_inputs();
+    *pin = in.toProtobuf();
+  }
+  for (const auto& out : outputs) {
+    auto pout = bufKey.add_outputs();
+    *pout = out.toProtobuf();
+  }
+  bufKey.set_device_str(deviceStr);
+  bufKey.set_git_version(tc::git_version);
+  return bufKey;
+}
+
+OptionsCacheKey OptionsCacheKey::fromProtobuf(
+    const OptionsCacheKeyProto& proto) {
+  OptionsCacheKey res;
+  res.id = lang::CanonicalTcString(proto.id());
+  for (int i = 0; i < proto.inputs().size(); ++i) {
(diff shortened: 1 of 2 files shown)


tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+
+bool OptionsCacheKey::operator==(const OptionsCacheKey& other) const {
+  if (id != other.id) {
+    return false;
+  }
+  if (deviceStr != other.deviceStr) {
+    return false;
+  }
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    if (not(inputs[i] == other.inputs[i])) {
```
ftynse: is there a reason why we can't use `!=` here?

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+  for (size_t i = 0; i < outputs.size(); ++i) {
+    if (not(outputs[i] == other.outputs[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+OptionsCacheKeyProto OptionsCacheKey::toProtobuf() const {
+  OptionsCacheKeyProto buf_key;
```
ftynse: camelCase

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+  for (const auto& in : inputs) {
+    auto pin = buf_key.add_inputs();
+    *pin = in.toProtobuf();
+  }
+  for (const auto& out : outputs) {
+    auto pout = buf_key.add_outputs();
+    *pout = out.toProtobuf();
+  }
+  buf_key.set_device_str(deviceStr);
+  buf_key.set_git_version("");
```
ftynse: git_version seems to be never modified by anything else in this PR, it will likely remain empty, and it isn't a likely intended behavior

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+  std::vector<Duration> runtimes;
+  for (auto d : proto.recorded_runtimes()) {
+    runtimes.push_back(Duration(d));
+  }
+  return OptionsCacheValue<Backend>{
+      runtimes, typename Backend::MappingOptionsType(proto.kernel_options())};
+}
+
+template <typename Backend>
+OptionsCache<Backend>::OptionsCache() {}
```
ftynse: explicit deafult instead?

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+std::vector<typename Backend::MappingOptionsType>
+OptionsCache<Backend>::getTopKOptions(
+    const lang::CanonicalTcString& tc,
+    const std::vector<TensorInfo>& inputs,
+    const std::vector<TensorInfo>& outputs,
+    const std::string& deviceStr,
+    size_t K) const {
+  std::lock_guard<std::mutex> lock(mutex);
+  ++numberAttemptedRetrievals;
+  struct WithTime {
```
ftynse: `OptionsWithTime / OptionsWithDuration`?  otherwise has spurious semantic correlation with `WithDevice`

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+}
+
+template <typename Backend>
+std::unordered_set<OptionsCacheKey, OptionsCacheKeyHash>
+OptionsCache<Backend>::getKeys() const {
+  std::lock_guard<std::mutex> lock(mutex);
+  std::unordered_set<OptionsCacheKey, OptionsCacheKeyHash> keys;
+  for (auto kvp : store_) {
+    if (keys.count(kvp.first) > 0) {
+      continue;
```
ftynse: is this check necessary?

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+    }
+    keys.emplace(kvp.first);
+  }
+  return keys;
+}
+
+template <typename Backend>
+void OptionsCache<Backend>::pruneKeepTopK(size_t K) {
+  auto keys = getKeys();
+  std::lock_guard<std::mutex> lock(mutex);
```
ftynse: I find it useful to create a block around a lock and thus highlight the critical section (when it does not span the existing block)

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,316 @@
...
+    for (auto it = range.first; it != range.second; ++it) {
+      toSort.push_back(it);
+    }
+    // this is of course wasteful but real topk is premature optimization atm
+    std::sort(
+        toSort.begin(),
+        toSort.end(),
+        [](const typename MultiMapType::iterator& a,
+           const typename MultiMapType::iterator& b) {
+          return median(a->second.runtimes) < median(b->second.runtimes);
```
ftynse: you could reuse `WithTime` struct from some function above instead of recomputing the median inside comparison, in a loop inside critical section

tc/autotuner/options_cache.h
```diff
@@ -0,0 +1,169 @@
...
+
+  std::vector<Duration> runtimes;
+  typename Backend::MappingOptionsType mappingOptions;
+};
+
+/**
+ * An Options cache is a simple abstraction around an unordered_multimap of
+ * protobuf-backed key value pairs. It has an underlying store object and
+ * provides simple functions to load/store from proto, record, prune and
+ * extract topK values ordered by runtime.
```
ftynse: mention thread-safety through the mutex?

PR #366, opened 2018-04-26
[C++ API] Step 7: Add files for the refactored autotuner

nicolasvasilache:
This PR adds the files implementing the refactored autotuner API (#342).

This PR makes the following changes to the autotuner:
1. since we don't have ExecutionEngine anymore, the tuning harness owns
the TcExecutor resulting from compilation. This results in significantly
simpler concurrency implementation and allows the pruning function to not
leak into the executor;
2. the autotuner and the harness are now templated by the backend type;
3. the autotuner is also templated by the search type;
4. the autotuner can only be created from a TC string and only exposes the
function tune;
5. anything related to multiple starting points was premature and disappears.
A good smell check is that all (almost all) calls to the tune function pass
both an options and an {options} parameter. We can make always turn baseMapping
into a vector in the future if we want to;
6. backing options caches are loaded and stored automatically if a filename
is passed;

the global refactoring.

### DIFF
--- a/tc/autotuner/autotuner-inl.h
+++ b/tc/autotuner/autotuner-inl.h
@@ -1,0 +1,538 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "tc/autotuner/autotuner.h"
+
+#include <atomic>
+#include <chrono>
+#include <numeric>
+#include <thread>
+
+#include <glog/stl_logging.h>
+
+#include "tc/autotuner/utils.h"
+#include "tc/core/compiler.h"
+#include "tc/core/flags.h"
+#include "tc/core/scope_guard.h"
+#include "tc/core/tensor.h"
+#include "tc/core/utils/math.h"
+#include "tc/lang/canonicalize.h"
+
+namespace tc {
+namespace autotune {
+namespace detail {
+template <typename Backend>
+TuningHarness<Backend>::TuningHarness(
+    size_t maxPopulationSize,
+    lang::TreeRef tcTree,
+    const std::unordered_map<size_t, std::vector<const DLConstTensor*>>& inputs,
+    std::unordered_map<size_t, std::vector<const DLTensor*>>& outputs,
+    const typename Backend::MappingOptionsType& baseMapping,
+    const TuningParameterFixer& fixedParams,
+    std::shared_ptr<OptionsCache<Backend>> optionsCache)
+    : stopRequested_(false),
+      currentCompilationJob_(0),
+      numEvaluations_(0),
+      tcTree_(tcTree),
+      kBaseMapping_(baseMapping),
+      kInputs_(inputs),
+      outputs_(outputs),
+      bestTime_(std::numeric_limits<size_t>::max()),
+      bestMappingOptions_(baseMapping),
+      optionsCache_(optionsCache) {}
+
+template <typename Backend>
+template <typename SearchStrategy>
+void TuningHarness<Backend>::run(SearchStrategy& searchStrategy) {
+  // TODO: kNumGenerations -> iterations
+  for (size_t i = 0; i < searchStrategy.kNumGenerations; ++i) {
+    if (not stopRequested_) {
+      runOneIteration(searchStrategy, i);
+    }
+  }
+}
+
+template <typename Backend>
+void TuningHarness<Backend>::stopAfterCurrentIteration() {
+  stopRequested_ = true;
+}
+
+template <typename Backend>
+const typename Backend::MappingOptionsType&
+TuningHarness<Backend>::bestMappingOptions() const {
+  std::lock_guard<std::mutex> lock(bestTimeMutex_);
+  return bestMappingOptions_;
+}
+
+#define LOG_LINE_BY_LINE(GSTREAM, ISTREAM)               \
+  for (std::string line; std::getline(ISTREAM, line);) { \
+    LOG(GSTREAM) << line;                                \
+  }
+
+template <typename Backend>
+template <typename SearchStrategy>
+void TuningHarness<Backend>::doCompile(SearchStrategy& searchStrategy) {
+  // Atomically fetch and add the next job until there are no jobs left
+  while (true) {
+    auto current = currentCompilationJob_.fetch_add(1);
+    if (current >= searchStrategy.population.size()) {
+      break;
+    }
+    std::unique_ptr<typename Backend::ExecutorType> pExecutor(nullptr);
+    auto pConf = searchStrategy.population.at(current).get();
+    auto options = makeOptions<Backend>(kBaseMapping_, *pConf);
+    try {
+      if (FLAGS_debug_tuner) {
+        std::stringstream ssInfo;
+        typename Backend::MappingOptionsCppPrinter infoPrinter(ssInfo);
+        infoPrinter << options;
+        LOG(INFO) << "[COMPILE] Start compilation @:" << current;
+        LOG_LINE_BY_LINE(INFO, ssInfo);
+      }
+      pExecutor =
+          tc::compile<Backend>(tcTree_, kInputs_.begin()->second, options);
+      LOG_IF(INFO, FLAGS_debug_tuner) << "[COMPILE] Done compilation";
+    } catch (const std::exception& e) {
+      LOG(WARNING) << "[TUNER][COMPILE] failed compilation: " << e.what();
+      std::stringstream ssWarning;
+      typename Backend::MappingOptionsCppPrinter warningPrinter(ssWarning);
+      warningPrinter << options;
+      LOG_LINE_BY_LINE(WARNING, ssWarning);
+      pConf->invalid = true;
+    }
+
+    // Emplace pExecutor (nullptr if compilation failed)
+    std::lock_guard<std::mutex> lock(executorsMutex_);
+    executors_.push(std::move(pExecutor));
+    configurations_.push(pConf);
+  }
+}
+
+template <typename Backend>
+void TuningHarness<Backend>::doEvaluate(
+    size_t device,
+    size_t populationSize,
+    Printer& printer) {
+  typename Backend::WithDevice wd(device);
+  CHECK_EQ(1, kInputs_.count(device));
+  auto& inputs = kInputs_.at(device);
+  CHECK_EQ(1, outputs_.count(device));
+  auto& outputs = outputs_.at(device);
+
+  while (true) {
+    auto current = numEvaluations_.load();
+    if (current >= populationSize) {
+      // We have seen enough, exit
+      break;
+    }
+
+    CandidateConfiguration* pConf;
+    std::unique_ptr<typename Backend::ExecutorType> pExecutor;
+    {
+      std::lock_guard<std::mutex> lock(executorsMutex_);
+      if (executors_.size() == 0) {
+        // No executors atm, wait for the queue to populate
+        continue;
+      }
+      pExecutor = std::move(executors_.front());
+      executors_.pop();
+      pConf = configurations_.front();
+      configurations_.pop();
+    }
+
+    // Properly keep track of count, RAII way
+    ScopeGuard sg([this]() { this->numEvaluations_.fetch_add(1); });
+    if (!pExecutor.get()) {
+      // If I popped an empty executor then compilation didn't go as
+      // planned, skip it.
+      CHECK(pConf->invalid);
+      continue;
+    }
+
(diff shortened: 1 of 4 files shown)


tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+  } else {
+    std::cout << "Dumping cache to " << tc::makeOptionsFilename(cacheFilename)
+              << "/" << Backend::makeDeviceFilename(cacheFilename) << std::endl;
+    optionsCache->pruneKeepTopK(10);
+    optionsCache->storeCacheToFile(tc::makeOptionsFilename(cacheFilename));
+    optionsCache->pruneKeepTopK(1);
+  }
+}
+
+std::vector<size_t> filterHigherThan(
```
ftynse: this function is no longer used AFAIK

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+    std::unordered_map<size_t, std::vector<const DLTensor*>>& outputs,
+    const typename Backend::MappingOptionsType& baseMapping,
+    const std::string& cacheFileName,
+    const TuningParameterFixer& fixedParams) {
+  CHECK_EQ(1, tcEntryPointMap_.count(tcEntryPoint))
+      << "Error looking up " << tcEntryPoint;
+
+  // Initialize a model configuration
+  TuningConfiguration modelConfiguration;
+  setupTuningParameters(inputs.begin()->second, modelConfiguration);
```
ftynse: check that `inputs` is not empty before this

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+
+  CHECK_GT(inputs.size(), 0);
+  auto inputInfos = makeTensorInfoVector(inputs.begin()->second);
+  auto outputInfos = makeTensorInfoVector(outputs.begin()->second);
+  return optionsCache_->getTopKOptions(
+      canonicalTc(tcEntryPointMap_.at(tcEntryPoint)),
+      inputInfos,
+      outputInfos,
+      Backend::deviceString(),
+      1);
```
ftynse: why doesn't this use `harness.bestMappingOptions()` ?  It looks like we have two ways of doing the same thing.

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+#include "tc/autotuner/utils.h"
+#include "tc/core/tensor.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/parser.h"
+
+namespace tc {
+namespace autotune {
+
+namespace detail {
+/**
```
ftynse: let's use the same documentation comment style in one file (the comments below use triple slash)

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+  void runOneIteration(SearchStrategy& searchStrategy, size_t iteration);
+
+  /// Helper function to delegate compiling on the cpu to different threads
+  template <typename SearchStrategy>
+  void doCompile(SearchStrategy& searchStrategy);
+
+  /// Helper function to delegate running the compiled code to different
+  /// threads.
+  /// This function must be specialized per Backend
+  /// TODO: virtual inheritance?
```
ftynse: virtual inheritance makes me think of virtual base classes, which is not what was intended here

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+  std::mutex bestTimeMutex_;
+  std::mutex executorsMutex_;
+  std::atomic_bool stopRequested_;
+  std::atomic_size_t currentCompilationJob_;
+  std::atomic_size_t numEvaluations_;
+  std::queue<std::unique_ptr<ExecutorType>> executors_;
+  std::queue<CandidateConfiguration*> configurations_;
+
+  /// inputs
+  lang::TreeRef kTcTree_;
```
ftynse: isn't `k` prefix for constants (not even constant members?)

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+  const MappingOptionsType kBaseMapping_;
+  /// maps of inputs and outputs per each device (represented by a size_t)
```
ftynse: "per device" or "for each device" I believe

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+
+  /// Runs autotuning on the TC function tcEntryPoint.
+  /// This assumes input and output tensors of proper sizes have been
+  /// initialized on each device. For now this initialization is left to the
+  /// client of the Autotuner.
+  /// Tuning performs a search of template type SearchStrategy starting from
+  /// baseMapping.
+  ///
+  /// Alternatively an OptionsCache cacheFileName serialized path
+  /// can be specified from which the tuner recovers multiple starting points
```
ftynse: This seems to only be true if a specific CLI flag is set...

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+  /// situation where short tunings are run and their results cached
+  /// iteratively.
+  ///
+  /// Lastly a TuningParameterFixer function can be specified to limit the
+  /// search space (i.e. when certain parameters are known to be good/bad
+  /// independently on a particular TC).
+  ///
+  /// \return a vector MappingOptions, if it is empty then tuning did not find
+  /// a single good configuration. This should be a very rare occurrence but
+  /// it is possible in particular if the skipExecutionOrWarmup functin is too
```
ftynse: "function"

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+  /// aggressive and the problem size is too small.
```
ftynse: looking at the code, this vector may contain only 0 or 1 elements;  worth mentioning here

tc/autotuner/autotuner.h
```diff
@@ -0,0 +1,229 @@
...
+      const TuningParameterFixer& fixedParams = TuningParameterFixer());
+
+ private:
+  std::map<std::string, lang::TreeRef> tcEntryPointMap_;
+  std::shared_ptr<OptionsCache<Backend>> optionsCache_;
+};
+
+/// Helper functions that need specializing for various backends.
+/// The implementation is not set in stone but just a first approximation to
+/// break the CUDA dependence.
```
ftynse: is this still relevant?

tc/autotuner/cuda/autotuner.cc
```diff
@@ -0,0 +1,174 @@
...
+namespace detail {
+
+using CudaTuningHarness = class TuningHarness<tc::CudaBackend>;
+
+template <>
+typename CudaBackend::MappingOptionsType makeOptions<CudaBackend>(
+    const typename CudaBackend::MappingOptionsType& baseMapping,
+    const CandidateConfiguration& c) {
+  auto options = baseMapping;
+  c.configuration.applyToCudaMappingOptions(options);
```
ftynse: This looks like TuningConfiguration must be made aware of different backend types... Not sure this is the as clean as other separations you've introduced so far.

tc/autotuner/cuda/autotuner.cc
```diff
@@ -0,0 +1,174 @@
...
+    return true;
+  }
+  // 2.b. during autotuning we don't want to spend too much time executing,
+  // use a reduced number of iterations for warmup.
+  constexpr int kReducedWarmupIterations = 2;
+  for (size_t i = 1; i < kReducedWarmupIterations - 1; ++i) {
+    executor.profile(inputs, outputs);
+  }
+
+  // 3. After reasonable warmup, look at the performance and prune if
```
ftynse: is 2 iterations a reasonable warmup?

tc/autotuner/cuda/autotuner.cc
```diff
@@ -0,0 +1,174 @@
...
+
+template <>
+void handleDeviceRuntimeError<CudaBackend>(
+    size_t device,
+    typename CudaBackend::MappingOptionsType& options) {
+  while (cudaGetLastError() != cudaSuccess) {
+    // In case of errors in the generated, we cannot rely on deviceReset to
+    // set the device in a clean state. So instead we just pop and discard
+    // all the errors accumulated on the device until we get to a clean
+    // slate (i.e. cudaSuccess).
```
ftynse: "state"

ftynse:

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+      LOG(WARNING) << "Aborted execution on device " << device;
+      LOG_LINE_BY_LINE(WARNING, ssWarning);
+      handleDeviceRuntimeError<Backend>(device, options);
+      pConf->invalid = true;
+      continue;
+    }
+
+    auto prof = median(runtimes);
+    size_t prof_us =
+        std::chrono::duration_cast<std::chrono::microseconds>(prof).count();
```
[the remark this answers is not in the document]
nicolasvasilache: right, legacy, cleaning up

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+    // Save best time under lock
+    {
+      std::lock_guard<std::mutex> lock(bestTimeMutex_);
+      if (prof_us < bestTime_) {
+        bestTime_ = prof_us;
+        bestMappingOptions_ = options;
+      }
+    }
+
+    // Trailing sanity check: this looks like a spurious case, fail very hard
```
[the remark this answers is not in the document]
nicolasvasilache: legacy, killing this

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+    const std::shared_ptr<OptionsCache<Backend>> optionsCache,
+    const std::string& cacheFilename) {
+  if (cacheFilename.empty()) {
+    std::cout << "No filepath provided, not saving cache" << std::endl;
+  } else {
+    std::cout << "Dumping cache to " << tc::makeOptionsFilename(cacheFilename)
+              << "/" << Backend::makeDeviceFilename(cacheFilename) << std::endl;
+    optionsCache->pruneKeepTopK(10);
+    optionsCache->storeCacheToFile(tc::makeOptionsFilename(cacheFilename));
+    optionsCache->pruneKeepTopK(1);
```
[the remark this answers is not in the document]
nicolasvasilache: yeah, bottom-up refactoring, missed that thanks!

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,560 @@
...
+template <typename Backend, typename SearchStrategy>
+std::vector<typename Backend::MappingOptionsType>
+Autotuner<Backend, SearchStrategy>::tune(
+    const std::string& tcEntryPoint,
+    const std::unordered_map<size_t, std::vector<const DLConstTensor*>>& inputs,
+    std::unordered_map<size_t, std::vector<const DLTensor*>>& outputs,
+    const typename Backend::MappingOptionsType& baseMapping,
+    const std::string& cacheFileName,
+    const TuningParameterFixer& fixedParams) {
+  CHECK_EQ(1, tcEntryPointMap_.count(tcEntryPoint))
```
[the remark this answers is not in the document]
nicolasvasilache: Punting on this but this raises the fact that tune should take both tc and entryPoint such that in the future we can do a self-contained call to a lightweight compile where we can put those checks. Addressing this in the API.

tc/autotuner/autotuner.h
nicolasvasilache: I've been using /* */ for structs and /// for methods in various place

tc/autotuner/autotuner.h
nicolasvasilache: killing that flag with fire

tc/autotuner/cuda/autotuner.cc
nicolasvasilache: Yes and I have been punting on rewriting this part too. I'd much rather land what we have now and collect the last crumbs as I finally do the functional CPU path.

tc/autotuner/cuda/autotuner.cc
nicolasvasilache: this is for ranking not absolute perf, I haven't seen cases where 2 is too small but I haven't conducted a statistical study either so YMMV

tc/autotuner/autotuner.h
nicolasvasilache: unfortunately yes, I still haven't decided what to do with these, they are related to TuningConfiguration which I am leaving for a later time

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,540 @@
...
+    const std::string& cacheFileName,
+    const std::vector<const DLConstTensor*>& inputs,
+    const size_t numCandidates) {
+  LOG_IF(INFO, FLAGS_debug_tuner)
+      << "Loading proto from: " << tc::makeOptionsFilename(cacheFileName)
+      << std::endl;
+  if (!cacheFileName.empty()) {
+    optionsCache->loadCacheFromFile(tc::makeOptionsFilename(cacheFileName));
+  }
+  tc::FLAGS_tuner_gen_restore_number =
```
ftynse: Sorry, I may have missed this part in the first pass (or it was not there).

Flags are essentially global variables.  Here, you change the global variable only to pass the result to a single function call, does not sound like a good idea.  What would happen if somebody wants to tune two different kernels in the same program (which does happen, e.g., in python) ?

nicolasvasilache: legacy.. not only what you mention but it is utterly useless, killing.

nicolasvasilache:

PR #367, opened 2018-04-26
[C++ API] Step 8: Add files for the refactored aten tuner

nicolasvasilache:
This PR adds the files implementing the refactored AtenAutotuner API (#345).

This refactoring essentially:
1. makes the tuner future-proof by templating by backend type and search strategy
2. drops the user-exposed load function which retrieves options from a
3. serialized proto cache: the tuner has no business providing this functionality.
If one wants to query the options cache one can do so explicitly
4. drops the owning pointer to a genetic autotuner: an ATenAutotuner is a
Autotuner<Backend, SearchStrategy> that operates on ATen tensors.
5. adds documentation and example of usage

the global refactoring.

### DIFF
--- a/tc/aten/aten_autotuner.h
+++ b/tc/aten/aten_autotuner.h
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "tc/aten/aten.h"
+#include "tc/autotuner/autotuner.h"
+
+namespace tc {
+namespace aten {
+/**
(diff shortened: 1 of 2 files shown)


skimo-openhub:
Is this supposed to replace #345?
I'm confused.

skimo-openhub:
On Fri, Apr 27, 2018 at 08:16:41AM -0700, skimo-openhub wrote:
> Is this supposed to replace #345?
> I'm confused.

Apparently, it is.  Shouldn't you close #345 then?

skimo

tc/aten/aten_autotuner.h
```diff
@@ -0,0 +1,92 @@
...
+  /// protobuf file.
+  ///
+  /// Lastly a TuningParameterFixer function can be specified to limit the
+  /// search space (i.e. when certain parameters are known to be good/bad
+  /// independently on a particular TC).
+  ///
+  /// \return a vector MappingOptions, if it is empty then tuning did not find
+  /// a single good configuration. This should be a very rare occurrence but
+  /// it is possible in particular if the skipExecutionOrWarmup function is too
+  /// aggressive and the problem size is too small.
```
skimo-openhub: What kind of MappingOptions does it return if the vector is not empty
and in what order?

tc/aten/aten_autotuner.h
```diff
@@ -0,0 +1,92 @@
...
+      const std::string& tcEntryPoint,
+      const std::vector<at::Tensor>& inputs,
+      const MappingOptionsType& baseMapping,
+      const std::string& cacheFileName = "",
+      const tc::autotune::TuningParameterFixer& fixedParams = {});
+
+ private:
+  /// The TC string is stored internally so we can tune independent TC
+  /// functions on demand.
+  const std::string kTc_;
```
skimo-openhub: I think @ftynse asked about this before, but isn't the "k" prefix used for (compile time) constants?

nicolasvasilache:

PR #368, opened 2018-04-27
isPromotableToRegisterBelowThreads: directly use relevant parts of schedule

skimo-openhub:
The original code would start off from a "full schedule" and
then project out the bits that aren't actually needed.
Directly use the relevant parts of the schedule instead.
This greatly simplifies the code.


ftynse:
please rebase on master

ftynse:
Running tests...

PR #378, opened 2018-04-30
halide2isl: handle non-affine subexpressions

### LINKED ISSUE #306: Indexing error
Hi guys -- thank you for the amazing project!

I have faced a small problem while trying to implement a perforated convolution layer -- basically just running convolution in the specified positions:
```
import os
import tensor_comprehensions as tc
import torch

# Use second GPU -pytorch-segmentation-detection- change if you want to use a first one
os.environ["CUDA_VISIBLE_DEVICES"] = '0'


LANG = """
def perforated_convolution(float(N, C, H, W) input,
                   float(M, C, KH, KW) weights,
                   int32(N, L) index) -> (output)
{{

output(n, m, l) +=! input(n, c, index(n, l) + kh, index(n, l) + kw) * weights(m, c, kh, kw) where l in 0:{L}

}}
"""

perforated_convolution = tc.define(LANG, name="perforated_convolution", constants={"L":3})

N, C, H, W, O, kH, kW = 2, 4, 56, 56, 16, 3, 3

tensor, weights = torch.randn(N, C, H, W).cuda(), torch.randn(O, C, kH, kW).cuda()

index = torch.IntTensor([ [10, 10, 12], [10, 10, 10] ]).contiguous().cuda()

out = perforated_convolution(tensor, weights, index)

# idx = torch.IntTensor([1, 1]).cuda()
# out = indexing(inp, idx)
```

This gave me this error:
```
[WARNING]: No mapping options passed, 'naive' type mapping options will be used and will likely have bad performance. See help(your_layer.__call__) for setting mapping options.
[ERROR]: Caught Exception: /home/daniil/anaconda3/include/isl/interface/isl.h:3079: NULL input
```

I was able to debug it and find a minimal example with indexing that gives me the same errror:
```
LANG = """
        def indexing(float(H, W) input, int32(L) index) -> (output) {{
            output(l, w) = input(index(l) + 1), w) where l in 0:{L}
        }}
        """
indexing = tc.define(LANG, name="indexing", constants={"L":2})
inp = torch.arange(0, 16).view(4, 4).cuda()
idx = torch.IntTensor([1, 1]).cuda()
out = indexing(inp, idx)
```

Do you know if there is a workaround for this?

ftynse:
1ee5f542af3f06ca25fe590282c6031152cb84c0 modified the conversion from
Halide Exprs to isl affs to add some basic support for min/max
operations.  When processing binary operators in Halide Exprs, it did
not take into account that one of the operands may be non-affine,
resulting in no affine boundary being computed.  The binary combinator
assumed both LHS and RHS bounds exist.  Remove this assumption and
remove an empty affine bound if either of LHS or RHS bound expressions
is not affine.

Closes #306


tc/core/halide2isl.cc
```diff
@@ -117,23 +117,29 @@ concatAffs(isl::space space, T op, bool allowMin, bool allowMax) {
 /*
  * Convert Halide binary expression "op" into an isl affine function by
  * converting its LHS and RHS into affs and combining them with "combine"
- * into a single expression.  LHS and RHS are expected to only produce a single
- * expression.
+ * into a single expression.  LHS and RHS are expected to only produce at most
+ * one expression.  If either of them produces zero expressions, meaning there
```
nicolasvasilache: their or the

nicolasvasilache:
LG, minor typo

ftynse:
updated

PR #379, opened 2018-05-01
[wip][c++api] refactor

nicolasvasilache:
This is the stacked PR that @skimo-openhub asked for to avoid seeing PRs in isolation.

Note the last commit is still in progress as I should be able to split off a few more things and I still need to fix some of the python bindings I disabled explicitly for now.


tc/autotuner/utils.cc
```diff
@@ -0,0 +1,206 @@
...
+#include "tc/core/flags.h"
+#include "tc/core/utils/dlpack.h"
+#include "tc/core/utils/time.h"
+#include "tc/lang/canonicalize.h"
+
+namespace tc {
+namespace autotune {
+
+std::vector<std::size_t> powers2andCeilDivisors(std::size_t val) {
+  auto numPowers = static_cast<std::size_t>(std::ceil(std::log2(val)));
```
skimo-openhub: This requires a "#include <cmath>"

.gitignore
```diff
@@ -20,3 +20,5 @@ tensor_comprehensions/version.py
 tensor_comprehensions/*.proto
 slurm-*
 examples/results*
+tensor_comprehensions/__init__.pyc
```
ftynse: `*.pyc` can be ignored, I think.  It's the same as `*.o`.

tc/autotuner/utils.cc
nicolasvasilache: I added #include <cmath> if that is what you meant?

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,342 @@
...
+  std::lock_guard<std::mutex> lock(mutex);
+  ++numberCacheAttempts;
+  OptionsCacheKey key{tc, inputs, outputs, deviceStr};
+  bool inserted = false;
+  auto range = store_.equal_range(key);
+  for (auto it = range.first; it != range.second; ++it) {
+    if (it->second.mappingOptions == options) {
+      it->second.runtimes.push_back(duration);
+      inserted = true;
+      break;
```
skimo-openhub: Why not simply return here instead of introducing an extra `inserted` boolean?

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,342 @@
...
+      throw std::runtime_error("No runtime for cache option");
+    }
+    toSort.push_back(Options{median(it->second.runtimes),
+                             it->second.runtimes,
+                             it->second.mappingOptions});
+  }
+  std::sort(
+      toSort.begin(), toSort.end(), [](const Options& a, const Options& b) {
+        // fun with C++, a.median < b.median does not mix with templates
+        return operator<(a.median, b.median);
```
skimo-openhub: Is this something compiler specific?
`a.median < b.median` seems to work fine on my system.

tc/autotuner/options_cache.h
```diff
@@ -0,0 +1,173 @@
...
+  void recordRuntime(
+      const lang::CanonicalTcString& tc,
+      const std::vector<TensorInfo>& inputs,
+      const std::vector<TensorInfo>& outputs,
+      const std::string& deviceStr,
+      const typename Backend::MappingOptionsType& options,
+      Duration duration);
+
+  /// Returns the top-K best mapping options for a particular
+  /// TC/inputs/outputs/device. Note that the result may be empty (in
```
skimo-openhub: Looking at the implementation, it seems "best" is defined in terms of
the median run-time.  Perhaps this is obvious, by I wouldn't have guessed that.

tc/autotuner/options_cache-inl.h
```diff
@@ -0,0 +1,342 @@
...
+    for (auto& k : keys) {
+      // this is of course wasteful but real topk is premature optimization atm
+      auto sorted = detail::sortedOptions<Backend>(k, store_);
+      // erase all then reinsert makes it easier to reuse code
+      store_.erase(k);
+      size_t kept = 0;
+      for (auto option : sorted) {
+        if (kept >= K) {
+          break;
+        }
```
skimo-openhub: This looks a bit convoluted.  How about simply
```
for (size_t i = 0; i < std::min(sorted.size(), K); ++i) {
  auto option = sorted[i];
```
Range-based for loops are nice if they improve readability,
but in this case, I would say it's not an improvement.

tc/autotuner/options_cache.h
```diff
@@ -0,0 +1,173 @@
...
+      OptionsCacheKey,
+      OptionsCacheValue<Backend>,
+      OptionsCacheKeyHash>
+      store_;
+};
+} // namespace autotune
+
+inline std::string makeOptionsFilename(const std::string& fn) {
+  return fn + ".options";
+}
```
skimo-openhub: I'm not sure about the exact rule for putting things in `-inl.h`, but perhaps
this should be put there too.

tc/autotuner/options_cache-inl.h
ttheodor: My guess is that the way `std::sort` is implement in some version of the standard library prevents ADL from working properly.

ttheodor:
@skimo-openhub the function templates don't need to be inline. The other ones could (should) probably be moved to a .cc file (?).

skimo-openhub:
Commit message of 1c20b18bddc40236e58568d8ce7fc209e9bd93a0:
> This PR makes the following changes to the autotuner:
> 1. since we don't have ExecutionEngine anymore, the tuning harness owns
> the TcExecutor resulting from compilation. This results in significantly
> simpler concurrency implementation and allows the pruning function to not
> leak into the executor;
> 2. the autotuner and the harness are now templated by the backend type;
> 3. the autotuner is also templated by the search type;
> 4. the autotuner can only be created from a TC string and only exposes the
> function tune;
> 5. anything related to multiple starting points was premature and disappears.
> A good smell check is that all (almost all) calls to the tune function pass
> both an options and an {options} parameter. We can make always turn baseMapping

drop "make"

> into a vector in the future if we want to;
> 6. backing options caches are loaded and stored automatically if a filename
> is passed;
> 
> (or even compiled) therefore the commit is guaranteed correct.

In some other commits, you replaced this by "cannot be tested independently",
which I think is more appropriate.

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,538 @@
...
+      for (auto& cpuCompilationThread : cpuCompilationThreads) {
+        cpuCompilationThread.join();
+      }
+    });
+    for (size_t i = 0; i < FLAGS_tuner_threads; ++i) {
+      cpuCompilationThreads.emplace_back(
+          [this, &searchStrategy]() { this->doCompile(searchStrategy); });
+    }
+
+    // Just spawn and join new threads for each iteration
```
skimo-openhub: each "device" instead of "iteration"?

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,538 @@
...
+
+// Creates well-chosen parameter sizes to match the input shapes.
+void setupTuningParameters(
+    const std::vector<const DLConstTensor*>& inputs,
+    TuningConfiguration& configuration) {
+  CHECK_GE(inputs.size(), 1);
+  auto range = inputDivisorsAndPowers2(inputs);
+  // 0 is a valid tiling annotation and signals no tiling of that dimension
+  // 0 is not a valid block / grid annotation
+  auto nTilesDim = largestDim(inputs) + 1;
```
skimo-openhub: Why `+ 1`?

tc/autotuner/cuda/autotuner.cc
```diff
@@ -0,0 +1,174 @@
...
+template <>
+TuningConfiguration makeTuningConfiguration<CudaBackend>(
+    const typename CudaBackend::MappingOptionsType& options,
+    const TuningConfiguration& configuration) {
+  TuningConfiguration conf = configuration;
+  conf.fromCudaMappingOptions(options);
+  return conf;
+}
+
+// This function is ran on a single pre-determined GPU, in a single thread
```
skimo-openhub: "is run"

tc/autotuner/cuda/autotuner.cc
```diff
@@ -0,0 +1,174 @@
...
+// It takes the input/output DLTensor objects that reside on that GPU
+//
+// We pass the bestTimeSoFar as an option to avoid taking locks in this
+// function. This trades off a bit of conservativeness for code sanity.
+//
+// The function returns true if purning is possible and we can skip poorly
```
skimo-openhub: pruning?

tc/autotuner/cuda/autotuner.cc
```diff
@@ -0,0 +1,174 @@
...
+template <>
+void handleDeviceRuntimeError<CudaBackend>(
+    size_t device,
+    typename CudaBackend::MappingOptionsType& options) {
+  while (cudaGetLastError() != cudaSuccess) {
+    // In case of errors in the generated, we cannot rely on deviceReset to
+    // set the device in a clean state. So instead we just pop and discard
+    // all the errors accumulated on the device until we get to a clean
+    // state (i.e. cudaSuccess).
+    ;
```
skimo-openhub: spurious empty statement?

tc/autotuner/utils.cc
```diff
@@ -0,0 +1,207 @@
...
+  res.reserve(restoreNumber);
+  std::transform(
+      candidates.begin(),
+      candidates.begin() + restoreNumber,
+      std::back_inserter(res),
+      [](const OptionsWithMedianTime& rr) { return rr.options; });
+  return res;
+}
+
+llvm::Optional<CudaMappingOptions> getBestOptions(
```
ftynse: do we want an llvm dependence here?  (not critical)

tc/core/cpu/cpu_backend.h
```diff
@@ -0,0 +1,101 @@
...
+ */
+struct CpuBackend {
+  using ExecutorType = CpuTcExecutor;
+  using MappingOptionsType = CpuMappingOptions;
+  using CompilationResultType = CpuCompilationResult;
+  using OptionsCacheProtoType = CpuOptionsCacheProto;
+  using OptionsCacheValueProtoType = CpuOptionsCacheValueProto;
+  using RTCFunctionType = CpuRTCFunction;
+
+  using WithDevice = tc::WithCpuDevice;
```
ftynse: why do these use fully-qualified names but the the statements above don't?

tc/core/cuda/cuda_backend.h
```diff
@@ -0,0 +1,84 @@
...
+ */
+struct CudaBackend {
+  using ExecutorType = CudaTcExecutor;
+  using MappingOptionsType = CudaMappingOptions;
+  using CompilationResultType = CudaCompilationResult;
+  using OptionsCacheProtoType = CudaOptionsCacheProto;
+  using OptionsCacheValueProtoType = CudaOptionsCacheValueProto;
+  using RTCFunctionType = CudaRTCFunction;
+
+  using WithDevice = tc::WithCudaDevice;
```
ftynse: why do these use fully-qualified names but the the statements above don't?

tc/autotuner/options_cache-inl.h
nicolasvasilache: it fails with 5.4.1

tc/autotuner/autotuner-inl.h
```diff
@@ -0,0 +1,538 @@
...
+      outputs_(outputs),
+      bestTime_(std::numeric_limits<size_t>::max()),
+      bestMappingOptions_(baseMapping),
+      optionsCache_(optionsCache) {}
+
+template <typename Backend>
+template <typename SearchStrategy>
+void TuningHarness<Backend>::run(SearchStrategy& searchStrategy) {
+  // TODO: kNumGenerations -> iterations
+  for (size_t i = 0; i < searchStrategy.kNumGenerations; ++i) {
```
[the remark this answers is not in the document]
nicolasvasilache: yes, it will become useful once we have a different search strategy that is not genetic-based, punting for now

tc/autotuner/utils.cc
nicolasvasilache: this will die when swap in the files, it's necessary for keeping git bisect working

tc/examples/README.md
```diff
@@ -31,11 +31,11 @@ Loading proto from: /tmp/batchmatmul_cache_B_500_K_26_M_72_N_26.options and /tmp
 
 
 Also when running manually, the number of CPU compilation threads and GPUs used for evaluation can be controlled via gflags
-```--tuner_threads``` and ```--tuner_gpus```
+```--tuner_threads``` and ```--tuner_devices```
 
 For instance, on a 4 GPU system with 20 threads:
 ```
-./build/examples/example_batchmatmul --disable_version_checks=true --autotune=true --tuner_gen_generations=10 --tuner_gen_pop_size=10 --tuner_threads=20 --tuner_gpus="0,1,2,3"
+./build/examples/example_batchmatmul --disable_version_checks=true --autotune=true --tuner_gen_generations=10 --tuner_gen_pop_size=10 --tuner_threads=20 --tuner_devices="0,1,2,3"
```
ftynse: you have probably forgotten to update the autotuning scripts that use this flag ;)

tc/examples/tensordot.cc
```diff
@@ -21,18 +21,33 @@
...
+#include "tc/core/cuda/cuda_tc_executor_new_api.h"
 #include "tc/core/flags.h"
 
-#include "../test/test_harness_aten_cuda.h"
-
 DEFINE_string(proto_path, "", "Filename to load and store proto cache ");
 
-TEST(TensorDot, SimpleAutotune) {
+template <typename Backend>
+at::Tensor makeATenTensor(at::ArrayRef<long int> sizes);
```
ftynse: it's a shame this thing is hidden in a particular example, could be useful elsewhere

tc/examples/wavenet.cc
```diff
@@ -170,6 +170,5 @@ int main(int argc, char** argv) {
   ::testing::InitGoogleTest(&argc, argv);
   ::gflags::ParseCommandLineFlags(&argc, &argv, true);
   ::google::InitGoogleLogging(argv[0]);
-  setAtenSeed(tc::initRandomSeed(), at::Backend::CUDA);
```
ftynse: Can't this be abstracted away the same way you did with tensor creation?

tensor_comprehensions/pybinds/pybind_autotuner.cc
```diff
@@ -22,18 +22,51 @@
...
+class ATenCudaTuner : public ATenCudaGeneticTuner {
+ public:
+  ATenCudaTuner(const std::string& tc)
+      : ATenCudaGeneticTuner(tc), tcEntryPointMap_(tc::detail::parse(tc)) {}
+
+  std::vector<tc::CudaMappingOptions> load(
+      const std::string& cacheFileName,
+      const std::string& entryPoint,
+      const std::vector<at::Tensor>& inputs,
+      const size_t& numCandidates) {
```
ftynse: size_t by-reference?

tensor_comprehensions/tc_unit.py
```diff
@@ -224,7 +224,7 @@ def tune_and_store(self, tc_name, inputs, mapping_options, cache_file=""):
         if not isinstance(options, Options):
             options = Options(options)
         try:
-            best_options = self.autotuner.tune(cache_file, tc_name, inputs, options, [options])
+            best_options = self.autotuner.tune(tc_name, inputs, options, cache_file, )
```
ftynse: spurious trailing comma in the argument list

test/cuda/test_autotuner_utility.cc
```diff
@@ -46,86 +50,85 @@ TEST(DivisorsAndPowers, Default) {
...
+    optionsCache.recordRuntime(
+        lang::canonicalTc(tc_),
+        makeTensorInfoVector(extractRawPtrs(inputDLTensors)),
+        makeTensorInfoVector(extractRawPtrs(outputDLTensors)),
+        CudaGPUInfo::GPUInfo().getCudaDeviceStr(),
+        options,
+        timings.kernelRuntime);
+  }
 
   FLAGS_tuner_gen_restore_number = 2;
```
ftynse: flags are global variables, we should really stop using them this way; but in a separate PR to make this clear to everyone

test/cuda/test_compilation_cache.cc
```diff
@@ -49,139 +78,238 @@ class OptionsCacheTest : public ::testing::Test {
...
+  {
+    OptionsCacheKey key{lang::CanonicalTcString("kernel2"),
+                        tc::makeTensorInfoVector(inputPtrs),
+                        tc::makeTensorInfoVector(outputPtrs),
+                        backendStr()};
+    ASSERT_EQ(optionsCache->count(key), 0u);
+  }
+
+  ASSERT_EQ(optionsCache->size(), 2u);
+  ASSERT_EQ(optionsCache->numberAttemptedRetrievals, 0u);
```
ftynse: Are these counters still useful?  You have clearly attempted to retrieve something from the cache without them being incremented...  If there is another way to retrieve stuff from caches, we should not use the "unsafe" way here (and better remove it completely).  Otherwise, just drop those counters.

test/cuda/test_compilation_cache.cc
```diff
@@ -49,139 +78,238 @@ class OptionsCacheTest : public ::testing::Test {
...
+  }
+  ASSERT_EQ(it1->second.mappingOptions, options0);
+  ASSERT_EQ(it1->second.runtimes.size(), 1u);
+  ASSERT_EQ(it1->second.runtimes[0], std::chrono::microseconds(1));
+
+  ASSERT_EQ(it2->second.mappingOptions, options1);
+  ASSERT_EQ(it2->second.runtimes.size(), 1u);
+  ASSERT_EQ(it2->second.runtimes[0], std::chrono::microseconds(2));
+
+  ASSERT_EQ(optionsCache->numberAttemptedRetrievals, 0u);
```
ftynse: idem for attempts

test/cuda/test_execution_engine.cc
```diff
@@ -21,38 +21,44 @@
 #include <glog/logging.h>
 #include <gtest/gtest.h>
```
ftynse: we don't have `execution_engine` anymore, rename this test and/or merge somewhere?

test/cuda/test_execution_engine.cc
```diff
@@ -255,7 +251,7 @@ def convolutionStrided(float(N,C,H,W) I, float(O,C,KH,KW) W1, float(O) B)
   tcStr = tc::replaceString(tcStr, "<sh>", std::to_string(SH));
   tcStr = tc::replaceString(tcStr, "<sw>", std::to_string(SW));
 
-  Check(
```
ftynse: outputs is used as argument of the call to `Check` here, but is not fully defined until `Check` returns.  Maybe you want to _drop_ the `outputs` argument and only use the returned vector?

test/cuda/test_execution_engine.cc
```diff
@@ -268,15 +264,13 @@ def convolutionStrided(float(N,C,H,W) I, float(O,C,KH,KW) W1, float(O) B)
...
+TEST_F(CompilationTest, Casts) {
   at::Tensor a = at::CUDA(at::kFloat).ones({2, 4});
   at::Tensor b = at::CUDA(at::kInt).tensor({}).fill_(4);
   a = a / 2.0 + 1;
   at::Tensor c = at::CUDA(at::kFloat).rand({3, 5});
 
-  std::vector<at::Tensor> outputs;
-
-  Check(
+  std::vector<at::Tensor> outputs = Check(
```
ftynse: same problem with outputs used in arguments

test/cuda/test_execution_engine_cache.cc
```diff
@@ -21,43 +21,42 @@
 #include <glog/logging.h>
 #include <gtest/gtest.h>
```
ftynse: execution_engine does not exist anymore, rename this file?

test/cuda/test_execution_engine_cache.cc
```diff
@@ -66,18 +65,20 @@ def matmul(float(M,K) A, float(K,N) B) -> (output) {
...
-  handle = atCompl.compile("matmul", inputs, mappingOptions);
-  atCompl.run("matmul", inputs, outputs2, handle);
+  auto outputs2 = tc::aten::prepareOutputs(tc, "matmul", inputs);
+  pExecutor =
+      tc::aten::compile<tc::CudaBackend>(tc, "matmul", inputs, mappingOptions);
+  tc::aten::run(*pExecutor, inputs, outputs2);
   diff = outputs2[0].sub(a2.mm(b2));
   checkRtol(diff, inputs, 8); // reduction size is dimension of n
 
   // run the first cached matmul again
```
ftynse: Does this file test anything useful?  It appears to only compile the same thing multiple times, but it never checks that some cache was hit or not.  I'm not even sure that compilation cache exists after refactorings.

test/cuda/test_execution_engine_db.cc
```diff
@@ -21,18 +21,17 @@
 #include <glog/logging.h>
 #include <gtest/gtest.h>
```
ftynse: execution_engine does not exist anymore, rename/merge this test?

test/cuda/test_tc_mapper_bugs.cc
```diff
@@ -690,10 +685,8 @@ def layernorm(float(T, B, C) I) -> (O, mean, centered, var) {
                      .unrollCopyShared(false)
                      .matchLibraryCalls(false);
 
-  tc::ATenCompilationUnit<tc::CudaTcExecutor> atCompl;
-  atCompl.define(TC);
-  // Expecting this to compile without dying.
```
ftynse: I would keep the comment

test/test_harness_aten_cuda.h
```diff
@@ -23,16 +23,16 @@
...
 
-#include "tc/aten/aten_compiler.h"
+#include "tc/aten/aten_compiler_new_api.h"
 #include "tc/core/cuda/cuda.h"
-#include "tc/core/cuda/cuda_tc_executor.h"
+#include "tc/core/cuda/cuda_tc_executor_new_api.h"
 #include "tc/core/flags.h"
 
-#include "test_harness_aten.h"
+#include "../../test/test_harness_aten.h"
```
ftynse: eh, why?  doing `../../` in the `./test` will get you outside of tc root dir.  These two files are next to each other in the hierarchy...

tc/c2/tc_op.h
```diff
@@ -85,48 +82,52 @@ class TcOp : public Operator<Context> {
...
+        rawOutputDLTensors_.push_back(outputDLTensors_.back().get());
+        outputVoidPtrs_.push_back(outputDLTensors_[i]->data);
       }
-      handle_ = executionEngine_->compile(
-          tcName_,
-          inputDLTensors_,
-          cudaMappingOptions_.toProtobufSerializedString());
+
+      // compile and run
+      executor_ = tc::compile<tc::CudaBackend>(
```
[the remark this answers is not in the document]
nicolasvasilache: For now yes, punting on this until after we sync up with internal stuff. I need to see some of @salexspb commits internally then we'll make a decision. IT's not degrading current behavior.

tc/c2/tc_op.h
```diff
@@ -85,48 +82,52 @@ class TcOp : public Operator<Context> {
...
+        inputDLTensors_.emplace_back(
+            caffe2::dlpack::makeDLConstTensor(this->Input(idx)));
+        rawInputDLTensors_.push_back(inputDLTensors_.back().get());
+        inputVoidPtrs_.push_back(inputDLTensors_.back()->data);
       }
-      auto outTensorInfo =
-          executionEngine_->inferOutputTensorInfo(tcName_, inputDLTensors_);
+
+      auto parsedTcs = tc::detail::parse(tc_);
+      CHECK_EQ(1, parsedTcs.count(tcName_))
```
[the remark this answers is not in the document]
nicolasvasilache: yeah same as usual we need a simple TC compile that will just give us parsing and sema errors.
Future work

tc/examples/README.md
nicolasvasilache: of course :) thanks!

tc/examples/tensordot.cc
nicolasvasilache: it's pretty vanilla .. we could add a few more convenience functions to aten.h, not here

tensor_comprehensions/pybinds/pybind_autotuner.cc
nicolasvasilache: lol

test/cuda/test_autotuner_utility.cc
nicolasvasilache: yup .. legacy

test/cuda/test_execution_engine_cache.cc
nicolasvasilache: just killing it, doesn't test anything useful anymore

test/cuda/test_execution_engine_cache.cc
nicolasvasilache: exactly since there are no more caches ever since before refactoring

test/cuda/test_execution_engine_db.cc
nicolasvasilache: killing it too

test/test_harness_aten_cuda.h
nicolasvasilache: wrong rebase, thanks!

nicolasvasilache:
> does C2 TcOp has Cuda backend hardcoded on purpose?

yes still for now

> there is one test where you use outputs in the same line that defines it, looks dangerous;

utter crap thanks

> some unnecessary weird relative inclusion path

wrong rebase, fixed thanks!

nicolasvasilache:
> tests still called "execution engine", there is no such thing anymore;
> furthermore, some tests do not test anything useful because the execution engine and the compilation caches disappeared;

killed, they should have died when I removed the compilation cache itself earlier

> python interface still uses define/compile scheme and named runs; eventually, we want to have similar interfaces

I went with not breaking the user experience and impl the thin abstraction directly inside the pybinds, we can iterate either but agreed, would be better to have a consistent CPI with C++

tc/aten/aten_autotuner.h
```diff
@@ -0,0 +1,94 @@
...
+  /// Optionally an OptionsCache cacheFileName serialized path
+  /// can be specified to which the tuner will save the best options found for
+  /// later offline reuse, in the proper protobuf format.
+  ///
+  /// Additionally, if such a cacheFileName is specified and if it contains a
+  /// previously saved protobuf then the autotuner will load it. In that case
+  /// the tuner recovers multiple starting points and appends them to the
+  /// baseMapping. This can be useful in a reinforcement situation where short
+  /// tunings are run and their results cached iteratively. The best options
+  /// are still saved at the end of tuning, altering that previously saved
```
skimo-openhub: Is any part of the original content preserved?
If not, it's better to say "overwriting" instead of "altering".

skimo-openhub:
Commit message of 2c279e27d9a3040624e713afe2992f5439c765c1:
> This allows better decoupling of the genetic search from the backend and
> removes the hardwired Cuda dependency.
> A generic way of printing the best options was previously added in
> autotuner-inl.h.

Do I understand correctly that this means the printing is temporarily disabled
(because `autotuner-inl.h` isn't actively used at this point)?
If so, it may be worth mentioning this in case we ever land here during a bisect.
Otherwise, looks good.

tc/core/cpu/cpu_backend.h
```diff
@@ -0,0 +1,101 @@
...
+#include "tc/core/cpu/cpu_mapping_options_cpp_printer.h"
+// TODO: activate when ready
+// #include "tc/core/cpu/cpu_rtc.h"
+#include "tc/core/halide_utils.h"
+#include "tc/core/tensor.h"
+
+namespace tc {
+/**
+ * This is the class that holds a binary object resulting from compilation.
+ * It must be provide a clear function and be callable.
```
skimo-openhub: Drop "be".  Maybe also replace "clear" by "clear()" to clarify that "clear" is the name
of the function rather than a qualification of the function.

tc/core/cpu/cpu_backend.h
```diff
@@ -0,0 +1,101 @@
...
+ * TODO: Impl me
+ */
+struct CpuRTCFunction {
```
skimo-openhub: Why not just put that in `tc/core/cpu/cpu_rtc.h` from the start?

tc/core/utils/time.h
```diff
@@ -18,7 +18,61 @@
...
 
 namespace tc {
-using Duration = std::chrono::high_resolution_clock::duration;
+struct Duration {
+ private:
+  explicit Duration(size_t us) : val_(us) {}
+  explicit Duration(std::chrono::microseconds us) : val_(us) {}
+
+ public:
+  Duration() : val_(std::chrono::microseconds::max()) {}
```
ftynse: I'd expect default-constructed duration to be zero

tc/autotuner/parameters.h
```diff
@@ -228,7 +229,6 @@ class TuningParameterFixer {
 };
 
 using TimePoint = std::chrono::high_resolution_clock::time_point;
```
ftynse: this typedef is useless

tc/autotuner/utils.cc
```diff
@@ -121,21 +115,19 @@ void Printer::stop() {
 }
 
 void Printer::printAll() {
-  auto runtimes = [this]() {
+  auto getSortedRuntimes = [this]() {
     std::lock_guard<std::mutex> lock(runtimesMtx_);
     std::sort(runtimes_.begin(), runtimes_.end());
```
ftynse: not sure why this sorts the internal `runtimes_` list each time this function called, under lock, just to _copy_ them to another vector for the sake of output

tc/core/cpu/cpu_tc_executor.cc
```diff
@@ -60,8 +60,7 @@ ProfilingInfo CpuTcExecutor::profileUnchecked(
     const std::vector<const void*>& inputs,
     const std::vector<void*>& outputs) const {
   LOG(ERROR) << "NYI: CpuTcExecutor::profileUnchecked";
-  return ProfilingInfo{
-      Duration(std::chrono::microseconds(static_cast<int64_t>(999999999999))),
-      Duration(std::chrono::microseconds(static_cast<int64_t>(999999999999)))};
+  return ProfilingInfo{Duration(Duration::fromMicroSeconds(999999999999)),
```
ftynse: are we sure this value is representable?  why not `Duration::max()` ?

ttheodor: I really dislike those sentinel values. We should be using optional (and we need to agree if we are going to roll our own or use a 3rd party one).

tc/autotuner/utils.cc
ttheodor: There's no need to. Should be: 

1. copy under lock
2. release lock
3. sort local and print

tc/core/cpu/cpu_tc_executor.cc
ftynse: This is _not_ an optional return value.  CPU run is not implemented, so this just returns something to silence the warning

tc/autotuner/utils.cc
```diff
@@ -0,0 +1,207 @@
...
+
+struct OptionsWithMedianTime {
+  CudaMappingOptions options;
+  Duration medianRuntime;
+};
+
+std::vector<OptionsWithMedianTime> getOptionsAndMedianRuntimes(
+    const lang::CanonicalTcString& id,
+    const std::vector<const DLTensor*>& inputs,
+    const std::vector<const DLTensor*>& outputs) {
```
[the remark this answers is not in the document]
nicolasvasilache: yes just legacy needed for bisect purposes

test/test_cuda_mapper.cc
```diff
@@ -758,7 +757,6 @@ def fun(float(N, M) A, float(N, M) B) -> (C,D) {
   auto tiledBand =
       ScheduleTree::makeScheduleTree(*scop->tileOuterBand(tiling.view));
 
-  using detail::ScheduleTreeElemBand;
```
skimo-openhub: The removal of the `detail::` qualification in this file could have been done up front.

test/test_tc2halide.cc
```diff
@@ -24,23 +24,14 @@
...
   void SetUp() {}
-  void Check(const string& tc, const std::vector<long>& inputSizes) {
-    auto ctx = getCPUDLContext();
-    DLDataType dtype;
-    dtype.code = kDLFloat;
-    dtype.bits = 32;
-    dtype.lanes = 1;
-    auto UPtr = makeDLTensorWithSizes(ctx, dtype, inputSizes);
-    std::vector<const DLTensor*> inputs{UPtr.get()};
-
```
skimo-openhub: Please mention why you removed this.
Bonus points if you can do it in a separate commit.

tc/core/utils/time.h
```diff
@@ -18,7 +18,61 @@
...
+  Duration(const Duration& d) : val_(d.val_) {}
+
+  static inline Duration fromStartStop(
+      std::chrono::time_point<std::chrono::system_clock> begin,
+      std::chrono::time_point<std::chrono::system_clock> end) {
```
skimo-openhub: The second argument to `fromStartStop` is always `std::chrono::system_clock::now()`.
Drop the second argument and rename (say, to `since`).

tc/core/utils/time.h
```diff
@@ -18,7 +18,61 @@
...
+  static inline Duration fromMicroSeconds(size_t us) {
+    return Duration(us);
+  }
+  static inline Duration zero() {
+    return Duration(0);
+  }
+  static inline Duration max() {
+    return Duration(std::chrono::microseconds::max());
+  }
+  inline size_t count() {
```
skimo-openhub: `count` doesn't mean much on something called `Duration`.
Given that you also have a `fromMicroSeconds`, maybe `inMicroSeconds`?

tc/core/cpu/cpu_tc_executor.cc
skimo-openhub: I also think `Duration::max()` would be more appropriate here,
or just use the default constructor.

tc/autotuner/cuda/autotuner.cc
```diff
@@ -107,9 +107,8 @@ bool skipExecutionOrWarmup<CudaBackend>(
   auto timings = executor.profile(inputs, outputs);
   // 2.a.
   constexpr size_t kCatastrophicPerfFactor = 100;
-  if (bestTimeSoFar < std::numeric_limits<size_t>::max() and
-      timings.kernelRuntime >= std::chrono::microseconds(
-                                   (kCatastrophicPerfFactor * bestTimeSoFar))) {
+  if (bestTimeSoFar < Duration::max() and
```
skimo-openhub: Do we need to expose the fact that a Duration defaults to max()?
We could have an `isSet()` function to perform this check instead.

tc/autotuner/utils.cc
nicolasvasilache: legacy I'll give it a quick shot but if I run into issues I'll open a task for @ttheodor  to clean up

tc/core/cuda/cuda_rtc.cc
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <iostream>
```
skimo-openhub: This looks like either an independent change or a spurious include.

tc/autotuner/autotuner-inl.h
```diff
@@ -54,8 +54,8 @@ TuningHarness<Backend>::TuningHarness(
 template <typename Backend>
 template <typename SearchStrategy>
 void TuningHarness<Backend>::run(SearchStrategy& searchStrategy) {
-  // TODO: kNumGenerations -> iterations
-  for (size_t i = 0; i < searchStrategy.kNumGenerations; ++i) {
+  // TODO: numGenerations -> iterations
+  for (size_t i = 0; i < searchStrategy.numGenerations; ++i) {
```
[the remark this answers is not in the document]
nicolasvasilache: we're still only using generations for genetic search, I'm not 100% clear iteration is the right name either. I'll remove the TODO and let whoever implements the next search strategy make the determination.

tc/autotuner/utils.cc
```diff
@@ -54,79 +50,17 @@ std::vector<std::size_t> powers2andCeilDivisors(std::size_t val) {
...
 void Printer::record(Duration runtime) {
   std::lock_guard<std::mutex> lock(runtimesMtx_);
   runtimes_.push_back(runtime);
 }
 
-namespace {
-uint64_t toMicroseconds(const Duration& d) {
-  return std::chrono::duration_cast<std::chrono::microseconds>(d).count();
-}
-} // namespace
```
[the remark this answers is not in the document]
nicolasvasilache: to make all Printer functions come together; the file is small enough that my discriminator chose locality of Printer functions over dependence distance

test/cuda/test_autotuner_utility.cc
```diff
@@ -46,86 +50,85 @@ TEST(DivisorsAndPowers, Default) {
...
+
+  auto inputDLTensors = makeDLConstTensors(inputs);
+  auto outputDLTensors = makeDLTensors(outputs);
+  optionsCache.recordRuntime(
+      lang::canonicalTc(tc_),
+      makeTensorInfoVector(extractRawPtrs(inputDLTensors)),
+      makeTensorInfoVector(extractRawPtrs(outputDLTensors)),
+      CudaGPUInfo::GPUInfo().getCudaDeviceStr(),
+      options,
+      timings.kernelRuntime);
```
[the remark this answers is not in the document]
nicolasvasilache: All these parameters are still needed, so best you can do is wrap them in a struct.
That's extra code, I don't really see the value.
Am I missing something?

test/cuda/test_compilation_cache.cc
```diff
@@ -14,30 +14,59 @@
...
-    tc::OptionsCache::enableCache();
-    ASSERT_TRUE(tc::OptionsCache::cacheEnabled());
-    tc::OptionsCache::getCache()->clear();
-    ASSERT_EQ(tc::OptionsCache::getCache()->size(), 0u);
-    ASSERT_EQ(tc::OptionsCache::getCache()->numberAttemptedRetrievals, 0);
-    ASSERT_EQ(tc::OptionsCache::getCache()->numberSuccessfulRetrievals, 0);
-    ASSERT_EQ(tc::OptionsCache::getCache()->numberCacheAttemps, 0);
+    optionsCache = std::shared_ptr<CudaOptionsCacheForTesting>(
+        new CudaOptionsCacheForTesting);
+    optionsCache->clear();
```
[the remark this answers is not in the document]
nicolasvasilache: legacy crumbs from the days caches used to be a global thing, killing

tc/examples/tensordot.cc
```diff
@@ -95,6 +122,5 @@ int main(int argc, char** argv) {
   ::testing::InitGoogleTest(&argc, argv);
   ::gflags::ParseCommandLineFlags(&argc, &argv, true);
   ::google::InitGoogleLogging(argv[0]);
-  setAtenSeed(tc::initRandomSeed(), at::Backend::CUDA);
```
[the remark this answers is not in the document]
nicolasvasilache: Used to be in harness and I don't want a harness in examples, graduating the function to aten.h

test/cuda/test_autotuner_utility.cc
```diff
@@ -46,86 +50,85 @@ TEST(DivisorsAndPowers, Default) {
...
-  ScopeGuard g([&]() {
-    deleteDlmTensors(inputsPair.second);
-    deleteDlmTensors(outputsPair.second);
-  });
-
-  return tc::autotune::restoreCandidates(
-      lang::canonicalTc(tc), inputsPair.first, outputsPair.first);
-}
-
-TEST(RestoreCandidates, NoCache) {
```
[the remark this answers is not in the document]
nicolasvasilache: This test relies on the implicit assumption that there is a global cache which is not true once we use the new API.
From now on, if there is no cache well there is no cache..

test/cuda/test_compilation_cache.cc
```diff
@@ -14,30 +14,59 @@
...
+#include "tc/lang/canonicalize.h"
 
 #include "test_harness_aten_cuda.h"
 
+using tc::autotune::OptionsCacheKey;
+using CudaOptionsCache = tc::autotune::OptionsCache<tc::CudaBackend>;
+struct CudaOptionsCacheForTesting : public CudaOptionsCache {
+ public:
+  // Unsafe equal_range not performed under lock, for testing only
+  std::pair<MultiMapType::iterator, MultiMapType::iterator> equal_range(
```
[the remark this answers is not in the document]
nicolasvasilache: I'll punt to @ttheodor since he owns this test #395

skimo-openhub:
On Fri, May 04, 2018 at 08:26:57AM -0700, Nicolas Vasilache wrote:
> > -  handle = atCompl.compile("matmul", inputs, options);
> -  atCompl.run("matmul", inputs, outputs_, handle, true);
> +  auto options = CudaMappingOptions::makeMlpMappingOptions();
> +  auto pExecutor = compile<CudaBackend>(tc_, "matmul", inputs, options);
> +  std::vector<at::Tensor> outputs = prepareOutputs(tc_, "matmul", inputs);
> +  auto timings = profile(*pExecutor, inputs, outputs);
> +
> +  auto inputDLTensors = makeDLConstTensors(inputs);
> +  auto outputDLTensors = makeDLTensors(outputs);
> +  optionsCache.recordRuntime(
> +      lang::canonicalTc(tc_),
> +      makeTensorInfoVector(extractRawPtrs(inputDLTensors)),
> +      makeTensorInfoVector(extractRawPtrs(outputDLTensors)),
> +      CudaGPUInfo::GPUInfo().getCudaDeviceStr(),
> +      options,
> +      timings.kernelRuntime);
> 
> All these parameters are still needed, so best you can do is wrap them in a struct.
> That's extra code, I don't really see the value.
> Am I missing something?

There's several things you could do to reduce boilerplate.
You could have a wrapper that performs
the `makeTensorInfoVector(extractRawPtrs())`.
You could let recordRuntime perform the canonicalization,
introducing a special recordCanonicalRuntime for the single
case in the code base where the caller does not want/need
to perform canonicalization.
You could have a recordCUDARuntime wrapper that hides
the `CudaGPUInfo::GPUInfo().getCudaDeviceStr()`.

It may also make sense to let the executor keep track of some
of the information.  It's definitely available when the executor
gets constructed and it already keeps track of the backend.
So you could have a call of the form

    optionsCache.recordRuntime(*pExecutor, timings.kernelRuntime)

meaning "record this runtime obtained with this executor".
Not for this PR, obviously.

skimo

test/cuda/test_corner_cases.cc
```diff
@@ -134,8 +132,7 @@ TEST(TestCornerCases, E9) {
 }
 
 TEST(TestCornerCases, E10) {
-  Succeed(
-      "def f(int32 a) -> (b) { b(i) = a where i in 0:10 }", {I()}, {I(10, 10)});
+  Succeed("def f(int32 a) -> (b) { b(i) = a where i in 0:10 }", {I()}, {I(10)});
```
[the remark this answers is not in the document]
nicolasvasilache: actually not independent at all, something in the old API was missing this, the new API catches this as an error so I fixed

skimo-openhub:
On Fri, May 04, 2018 at 04:45:03PM +0000, Nicolas Vasilache wrote:
> > @@ -134,8 +132,7 @@ TEST(TestCornerCases, E9) {
>  }
>  
>  TEST(TestCornerCases, E10) {
> -  Succeed(
> -      "def f(int32 a) -> (b) { b(i) = a where i in 0:10 }", {I()}, {I(10, 10)});
> +  Succeed("def f(int32 a) -> (b) { b(i) = a where i in 0:10 }", {I()}, {I(10)});
> 
> actually not independent at all, something in the old API was missing this, the new API catches this as an error so I fixed

But the fix could go in independently (before this commit),
couldn't it?

skimo

nicolasvasilache:
> But the fix could go in independently (before this commit), couldn't it?

That wouldn't be a fix but just hiding of the problem, I'll add a test for the negative case that is supposed to fail as a separate commit after the API has landed, see #397

tc/autotuner/cuda/autotuner.cc
nicolasvasilache: Nah, just made the constructor default to 0 and private

skimo-openhub:
9fc2dadb933fbd278e4369d54949a4c1c686f734 has some typos in the commit message
> Drop examples reliance on test harness

Drop reliance on the harness from examples
(or "Drop examples' reliance on test harness", but it sounds a bit weird)

> The idea of exmaples is they should be standalone and as simple as possible.

examples

> Relying on test harness goes against this principle.
> set/getAtenSeed graduated to aten.h in the previous commit, no need to
> rely on the test harnedd anymore.

harness

Otherwise looks good.

nicolasvasilache:
> 0c5e9e4 has a typo: "usefule" -> useful. Otherwise looks good.

No that's the french spelling for "useful" :)
Fixed, thanks!

skimo-openhub:
On Thu, May 03, 2018 at 07:04:03PM -0700, Nicolas Vasilache wrote:
>  For instance, on a 4 GPU system with 20 threads:
>  ```
> -./build/examples/example_batchmatmul --disable_version_checks=true --autotune=true --tuner_gen_generations=10 --tuner_gen_pop_size=10 --tuner_threads=20 --tuner_gpus="0,1,2,3"
> +./build/examples/example_batchmatmul --disable_version_checks=true --autotune=true --tuner_gen_generations=10 --tuner_gen_pop_size=10 --tuner_threads=20 --tuner_devices="0,1,2,3"
> 
> of course :) thanks!

Was this a reference to tc/benchmarks/scripts/AUTOTUNER_COMMANDS?
It still uses "--tuner_gpus".

skimo

PR #380, opened 2018-05-01
[WIP] Add bitwise operators and propagate them from parser -> halide -> isl

### LINKED ISSUE #131: Bitwise Operation Support
Tensor Comprehensions seem like a great way to implement binary approximation of layers as in [BinaryNet](https://arxiv.org/abs/1602.02830). However, tc does not currently support bitwise operations such as xnor (^), or (|), and and (&). Additionally, implementing these approximations would require the popc cuda intrinsic, which is no supported. Getting these functions integrated would enable really cool and fast applications that could highlight the strengths of tc.

prigoyal:
trying to add bitwise operators. so far they are parsed correctly in their precedence order. `tc2halide` conversion is also done.

I am not sure about how to propagate these through to isl yet.

**NOTE**: I stacked this PR on add-operators branch to avoid merge conflicts.

closes #131

### DIFF
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -201,7 +201,7 @@ std::vector<isl::aff> makeIslAffBoundsFromExpr(
     std::vector<isl::aff> result;
     // We cannot span multiple constraints if a modulo operation is involved.
     // x > max(a,b) % C is not equivalent to (x > a % C && x > b % C).
-    auto lhs = makeIslAffBoundsFromExpr(space, e, false, false);
+    auto lhs = makeIslAffBoundsFromExpr(space, op->a, false, false);
     CHECK_EQ(lhs.size(), 1u);
     if (const int64_t* b = as_const_int(op->b)) {
       return {lhs[0].mod(isl::val(space.get_ctx(), *b))};
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -35,7 +35,8 @@ using namespace tc::polyhedral::detail;
 
 SymbolTable makeSymbolTable(const tc2halide::HalideComponents& components) {
   // const Stmt& s) {
-  // Collect and categorize all the Variable symbols
+  // Collect and categorize all the Halide Variable symbols as reduction
+  // or index variables
   class BuildSymbolTable : public IRVisitor {
     using IRVisitor::visit;
     std::set<std::string> included;
@@ -60,7 +61,7 @@ SymbolTable makeSymbolTable(const tc2halide::HalideComponents& components) {
 
   components.stmt.accept(&builder);
   // Get params from components.params which contain everything declared in
-  // tcdef. However, the 0-D tensors are registered as both params and inputs,
+  // TC Def. However, the 0-D tensors are registered as both params and inputs,
   // filter those out.
   for (auto kvp : components.params) {
     bool skip = false;
--- a/tc/core/libraries.h
+++ b/tc/core/libraries.h
@@ -145,6 +145,12 @@ template<typename T> inline __device__ T floord(T n, T d) {
   return n < 0 ? - (-n + d - 1)/d : n / d;
 }
 #define if_then_else(cond,a,b) ((cond) ? (a) : (b))
+#define shift_left(a,b) ((a) << (b))
+#define shift_right(a,b) ((a) >> (b))
+#define bitwise_and(a,b) ((a) & (b))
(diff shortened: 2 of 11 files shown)


prigoyal:
@skimo-openhub , @ftynse , how do the operators propagate to isl? I see files like `tc/external/detail/islpp.h` but I don't fully understand how/when/if operators are added there. Can you provide some guidance? Thanks a lot.

abadams:
Note that the Halide simplifier will convert all the simpler bit operations to multiplication (x << 5), division (x >> 5), and modulo (x & 0x3f), so the only bit operations that reach isl will be the nastier ones. That said, bit operations do have well-defined bounds that we may be able to exploit (e.g. for unsigned ints (x & y) <= x)

prigoyal:
thanks @abadams . I have two questions for @skimo-openhub and @ftynse: 

1. should we support quasi-affine bitwise or can it be more general? from looking at halide bitwise ops, https://github.com/halide/Halide/blob/master/test/correctness/bitwise_ops.cpp, it looks like it doesn't need to be quasi-affine but not sure about isl
2. any ideas around how these ops will be propagated to isl?

ftynse:
(quick answer, I'll give more details when not on the phone)

We need to differentiate two cases for bitwise operations, similarly to modulo.  If they appear in tensor _subscripts_, those are not affine and should be overapproximated.  Outer operations should be handled as function calls essentially.

ftynse:
(quick answer, I'll give more details when not on the phone)

We need to differentiate two cases for bitwise operations, similarly to modulo.  If they appear in tensor _subscripts_, those are not affine and should be overapproximated.  Outer operations should be handled as function calls essentially.

prigoyal:
thanks. I will wait to get more details to understand better the cases to distinguish.

In the meantime, propagation from halide to isl doesn't work by default. here is the codegen for one example:

```
extern "C" {
__global__ void f(int32* pc, const int32* pa, const int32* pb) {
  int b0 = blockIdx.x; int b1 = blockIdx.y; int b2 = blockIdx.z;
  int t0 = threadIdx.x; int t1 = threadIdx.y; int t2 = threadIdx.z;
  int32 (*c) = reinterpret_cast<int32 (*)>(pc);
  const int32 (*a) = reinterpret_cast<const int32 (*)>(pa);
  const int32 (*b) = reinterpret_cast<const int32 (*)>(pb);
  c[0] = shift_left(a[0], b[0]);
}
}
```
Note the `shift_left` is the halide representation but I don't know what these operations are called in isl. I would appreciate any pointers.

ftynse:
isl must be agnostic to these operations.  Essentially, the level we are interested about in isl is _below_ those.  At that level, there is no difference between `A[i] + B[i]` and `A[i] << B[i]`.  We are only interested by _what data_ is accessed, not which operations are performed on it.  (This get different when expressions appear in the _subscripts_ because it affects _what data_ is accessed, but let's punt on it for now.)

In this particular case, the simplest solution is to inject `#define shift_left(a,b) ((a) << (b))` in the code to make Halide's function work.  I don't like this solution because it makes `shift_left` a reserved name in TC, but we already do so for something else if I'm not mistaken.  Alternatively, somewhere on isl/Halide codegen boundary, we could rewrite that expression to use `<<`.  Unfortunately, I cannot give you a pointer there immediately.

ftynse:
Now, if operations appear in subscripts, like `O(i << j) = A(i)`, isl must care about those because it changes _what data_ in `O` is accessed.  isl in particular and the polyhedral model in general is mostly designed for handling _affine_ expressions (that is, sums of constants and products of constants with variables).  There are some extensions allowed, which we call quasi-affine expressions.  More specifically, division and modulus by a constant is allowed.  Whenever the subscript expression is not affine, the common approach in the polyhedral is to over-approximate it: instead of saying "i-th iteration accesses O[i]", we say "i-th iteration _may_ access any element of O".

Bitiwise operations in general are not affine, so you would need that over-approximation. It should be rather straightforward to implement in `makeIslAffFrom...` functions, there are already some non-affine cases. Shifts with constant RHS are a special case.  They can be transformed to products (left shift) and divisions (right shift) with a constant factor, and that is (quasi)affine.

prigoyal:
thanks @ftynse, the bitwise operators work now for the simple test case added to check correctness.

can you provide what more test cases should I test this on (especially over-approximation, affine/non-affine scenarios)? I believe that would help me understand more how to address those cases by digging more into the codebase :) thanks a lot for your guidance.

skimo-openhub:
add me back when there is any progress

PR #382, opened 2018-05-02
ScheduleTreeElemBand::drop: do not preserve tuple identifier

skimo-openhub:
When some dimensions get removed from a space, then it is clearly
not the same space anymore, so it is dangerous to try and pretend
that it is.  It is also fairly arbitrary to try and save the identifier,
but not the internal structure of the space.
There may be specific situations where a user is changing the space
and where the tuple identifier should be preserved, but this should
then be the done by the caller and not by a generic method such as
ScheduleTreeElemBand::drop.


ftynse:
Hmm, this function was introduced for domain set manipulations AFAIR.  Range spaces of the schedule don't have tuple ids anyway...

PR #383, opened 2018-05-02
 Codegen::makeLoopIterators clean-ups

skimo-openhub:


tc/core/polyhedral/codegen.h
```diff
@@ -27,14 +28,13 @@ class Codegen {
 
  public:
   // Create a list of isl ids to be used as loop iterators when building the
-  // AST.
+  // AST from the schedule tree with the given root.
   //
   // Note that this function can be scrapped as ISL can generate some default
```
ftynse: Looking at this comment you did not touch (sorry for that), is there indeed such a function to generate a list of ids with the same prefix and different numeric suffixes?  I would expect it in PPCG.  If so, let's use that function.  Otherwise, let's kill the incorrect comment since we are touching the function.

skimo-openhub:
On Wed, May 02, 2018 at 09:20:19AM -0700, ftynse wrote:
> > @@ -27,14 +28,13 @@ class Codegen {
>  
>   public:
>    // Create a list of isl ids to be used as loop iterators when building the
> -  // AST.
> +  // AST from the schedule tree with the given root.
>    //
>    // Note that this function can be scrapped as ISL can generate some default
> 
> Looking at this comment you did not touch (sorry for that), is there indeed such a function to generate a list of ids with the same prefix and different numeric suffixes?  I would expect it in PPCG.  If so, let's use that function.  Otherwise, let's kill the incorrect comment since we are touching the function.

The isl AST generator should generate names automatically
if the user didn't provide any, as you mention in that comment.
So, there isn't any function that needs to be called.
We can simply not set any names ourselves.
Is that what you prefer now?

skimo

ftynse:
Not sure. It just feels like, at some point in the prehistory, we were using or wanted to use the user pointers of loop ids to communicate some information, and this is discouraged now.

Accepting this PR as is.

skimo-openhub:
On Thu, May 03, 2018 at 08:34:41AM +0000, ftynse wrote:
> Not sure. It just feels like, at some point in the prehistory, we were using or wanted to use the user pointers of loop ids to communicate some information, and this is discouraged now.

Let's just remove it then.

skimo

skimo-openhub:
On Thu, May 03, 2018 at 01:45:13AM -0700, skimo-openhub wrote:
> On Thu, May 03, 2018 at 08:34:41AM +0000, ftynse wrote:
> > Not sure. It just feels like, at some point in the prehistory, we were using or wanted to use the user pointers of loop ids to communicate some information, and this is discouraged now.
> 
> Let's just remove it then.

While writing the commit message for that, I realized that
it can be useful to set the names explicitly, in particular
to avoid name clashes.  That's what PPCG does.
Is there some other mechanism for avoiding name clashes?

skimo

skimo-openhub:
On Thu, May 03, 2018 at 02:19:30AM -0700, ftynse wrote:
> I don't think there is one.  It may be even worse, I don't think `c0`, `c1`, etc. are not allowed as tensor names so we can have a clash there.

Euhm... that's what I meant with name clashes...

OK, so this function may still turn out to be useful.
I'll just merge this PR then.

skimo

skimo-openhub:
On Thu, May 03, 2018 at 02:24:57AM -0700, skimo-openhub wrote:
> On Thu, May 03, 2018 at 02:19:30AM -0700, ftynse wrote:
> > I don't think there is one.  It may be even worse, I don't think `c0`, `c1`, etc. are not allowed as tensor names so we can have a clash there.
> 
> Euhm... that's what I meant with name clashes...

Ah... I see what you mean.  PPCG doesn't prevent all name clashes
either.

skimo

PR #384, opened 2018-05-02
test_cpu.sh: run test/test_cuda_mapper

ftynse:
Despite the name, this test does not require CUDA to be present in the
system to run.  It only includes syntactic tests on CUDA code generator
and can be run by CPU only.  Run it in test_cpu.sh, including on
CircleCI by default.


ftynse:
thanks @prigoyal 
I will merge when Jenkins tests pass too

PR #387, opened 2018-05-03
[c++ api] add support parameters functions

nicolasvasilache:
This is stacked on top of 9e8a8b4

PR #388, opened 2018-05-03
[C++ API] Extend CpuMappingOptions

nicolasvasilache:
This is stacked on 9e8a8b4

This changeset adds supporting functions and objects that will allow an
end-to-end mock CPU example to compile and run, in the context of the global
C++ API refactoring.

This changeset introduces extra methods in CpuMappingOptions to allow it to
be used as a template type with generic function names.

This changeset also moves the implementation of CPU backend specific options
files.


tc/core/cpu/cpu_mapping_options.h
```diff
@@ -22,15 +22,59 @@
...
   /// Construct from a serialized protocol buffer message.
-  inline explicit CpuMappingOptions(const std::string& str);
+  explicit CpuMappingOptions(const std::string& str);
+
+  std::string toProtobufSerializedString() const;
 
-  inline bool operator==(const CpuMappingOptions& options);
+  /// Set mappings
+  CpuMappingOptions& genericMappingOptions(const MappingOptions& options);
+  ///@}
```
ftynse: this looks unmatched

nicolasvasilache:
@ftynse  yes I updated the description: this is stacked on 9e8a8b4
I'll rebase now that it has been merged.

nicolasvasilache:
#387 which is rebased on this passed, merging

PR #389, opened 2018-05-03
[c++api] uniformize cuda mapping options

nicolasvasilache:
Stacked on top of 44b4b2d

This changeset moves the implementation of CUDA backend specific options
files.


nicolasvasilache:
green CI but needs a trivial rebase, I'll do this and merge when accepted without waiting for another round of CI

PR #391, opened 2018-05-03
[C++ API] Move impl of CudaMappingOptionsPrinter to .cc

nicolasvasilache:
This changeset moves the implementation of CUDA backend specific options printer
files.


nicolasvasilache:
@ftynse previous commit introduced CpuMappingOptionsPrinter and puts impl in .cc
This commit moves the CudaMappingOptionsPrinter impl to .cc too (was in -inl.h previously)

PR #399, opened 2018-05-04
Update from facebook 1ccd0b4e67b3

mingzhe09088:
This PR was generated by the sync script to sync TC from fbcode to github.

### DIFF
--- a/tc/c2/tc_op.h
+++ b/tc/c2/tc_op.h
@@ -41,14 +41,11 @@ class TcOp : public Operator<Context> {
         cudaMappingOptions_(tc::CudaMappingOptions::makeNaiveMappingOptions()),
         gradCudaMappingOptions_(
             tc::CudaMappingOptions::makeNaiveMappingOptions()) {
-    is_backward_ =
-      OperatorBase::GetSingleArgument<bool>("is_backward_", false);
-    tc_ =
-      OperatorBase::GetSingleArgument<std::string>(
-          is_backward_ ? "tcGradDef" : "tcDef", "ERROR");
-    tcName_ =
-      OperatorBase::GetSingleArgument<std::string>(
-          is_backward_ ? "tcGradName" : "tcName", "ERROR");
+    is_backward_ = OperatorBase::GetSingleArgument<bool>("is_backward_", false);
+    tc_ = OperatorBase::GetSingleArgument<std::string>(
+        is_backward_ ? "tcGradDef" : "tcDef", "ERROR");
+    tcName_ = OperatorBase::GetSingleArgument<std::string>(
+        is_backward_ ? "tcGradName" : "tcName", "ERROR");
     compiled_ = false;
     checkSizes_ = OperatorBase::GetSingleArgument<bool>("checkSizes", false);
     ArgumentHelper args(operator_def);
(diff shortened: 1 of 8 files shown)


tc/core/polyhedral/schedule_tree.cc
```diff
@@ -343,7 +343,9 @@ vector<const ScheduleTree*> ScheduleTree::collectDFSPreorder(
 vector<const ScheduleTree*> ScheduleTree::collectDFSPreorder(
     const ScheduleTree* tree,
     detail::ScheduleTreeType type) {
-  auto filterType = [type](const ScheduleTree* t) { return t->type_ == type; };
+  auto filterType = [type](const ScheduleTree* t) {
```
nicolasvasilache: The commit message is completely decorrelated from the changes which themselves should not have happened.

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -17,6 +17,8 @@
 #include <string>
 #include <vector>
 
+#include <google/protobuf/text_format.h>
```
nicolasvasilache: Changes to this file should land in a first commit with a proper commit message.

For the python part, depending on whether or not the test can run in OSS, it should either be completely omitted if it doesn't run (i.e. it is dead code) or it should be also added to the CI in the same PR.

Commit message is completely off.

test_python/test_c2.py
```diff
@@ -35,10 +35,17 @@ def matmul(float(M,N) A, float(N,K) B) -> (output) {
 }
 """
 
+MATMUL_GRAD_LANG = """
```
nicolasvasilache: The cpp portions of the commit make sense.

For the python part, depending on whether or not the test can run in OSS, it should either be completely omitted if it doesn't run (i.e. it is dead code) or it should be also added to the CI in the same PR.

Commit message is off.

In any case a C++ test should be added to test_caffe2.cc which tests gradient from C++, otherwhise this is dead code.

test_python/test_c2.py
```diff
@@ -35,10 +35,17 @@ def matmul(float(M,N) A, float(N,K) B) -> (output) {
...
+def matmul_grad(float(M, N) A, float(N, K) B, float(M, K) O_grad) -> (A_grad, B_grad) {
+    A_grad(i, kk) +=! O_grad(i, j) * B(kk, j)
```
nicolasvasilache: Please see out [TC coding guidelines](https://github.com/facebookresearch/TensorComprehensions/blob/master/docs/source/coding_conventions.rst).

tc/c2/tc_op.h
```diff
@@ -41,14 +41,11 @@ class TcOp : public Operator<Context> {
...
             tc::CudaMappingOptions::makeNaiveMappingOptions()) {
-    is_backward_ =
-      OperatorBase::GetSingleArgument<bool>("is_backward_", false);
-    tc_ =
-      OperatorBase::GetSingleArgument<std::string>(
-          is_backward_ ? "tcGradDef" : "tcDef", "ERROR");
-    tcName_ =
-      OperatorBase::GetSingleArgument<std::string>(
-          is_backward_ ? "tcGradName" : "tcName", "ERROR");
+    is_backward_ = OperatorBase::GetSingleArgument<bool>("is_backward_", false);
```
nicolasvasilache: These formatting changes should be squashed in the offending commit.

nicolasvasilache:
#400 is the manual replacement of this automatic PR

nicolasvasilache:
That will be fixed in the last commit of #400 when C++ gradient works. The
convention can be either:

1. Lower camel case in tc_op.h and underscore separated strings in args
(consistent with caffe2 style from python)
2. Underscore separated everywhere in tc_op.h consistent with caffe2 style.

Since tc_op is in the caffe2 namespace and ideally would live in the caffe2
repo at some point, I would go for the latter unless there are strong
objections.

PR #400, opened 2018-05-05
Update from facebook 1ccd0b4e67b3

nicolasvasilache:
This is a 1-off manual sync from fbcode to github.
It aims at replacing #399 and additionally introduces a C++ unit test for gradient.
There are followup items in #402 to put the TC caffe2 tests in a better place for integration into fbcode.

Even after #402 there are still a few followup cleanups to address but this moves the gradient in the right direction.

### DIFF
--- a/tc/c2/convolution_op.h
+++ b/tc/c2/convolution_op.h
@@ -63,8 +63,6 @@ class TcConvolutionOp : public TcOp<T, Context, Engine> {
 
     this->tc_ = tc::makeConvolution2DTc(strideH, strideW);
     this->tcName_ = tc::CONVOLUTION2D_TC_NAME;
-    this->gradTc_ = tc::makeConvolution2DGradTc(strideH, strideW);
-    this->gradTcName_ = tc::CONVOLUTION2D_GRAD_TC_NAME;
   }
 
   ~TcConvolutionOp() override {}
@@ -72,8 +70,6 @@ class TcConvolutionOp : public TcOp<T, Context, Engine> {
  protected:
   void setupNaiveCudaMappingOptions() override {
     this->cudaMappingOptions_ =
-        tc::CudaMappingOptions::makeConvolutionMappingOptions();
-    this->gradCudaMappingOptions_ =
         tc::CudaMappingOptions::makeConvolutionMappingOptions();
   }
 };
--- a/tc/c2/tc_op.h
+++ b/tc/c2/tc_op.h
@@ -83,12 +83,8 @@ class TcOp : public Operator<Context> {
         inputVoidPtrs_.push_back(inputDLTensors_.back()->data);
       }
 
-      auto parsedTcs = tc::detail::parse(tc_);
-      CHECK_EQ(parsedTcs.count(tcName_), 1u)
-          << "attempting to access undefined function " << tcName_;
-
-      auto outTensorInfo = tc::detail::inferOutputTensorInfo(
-          parsedTcs.at(tcName_), rawInputDLTensors_);
+      auto outTensorInfo =
+          tc::inferOutputTensorInfo(tc_, tcName_, rawInputDLTensors_);
       prepareOutputs(outTensorInfo);
 
       // now create the outputDLTensors
@@ -100,7 +96,7 @@ class TcOp : public Operator<Context> {
 
       // compile
       executor_ = tc::compile<tc::CudaBackend>(
-          parsedTcs.at(tcName_), rawInputDLTensors_, cudaMappingOptions_);
+          tc_, tcName_, rawInputDLTensors_, cudaMappingOptions_);
       compiled_ = true;
     }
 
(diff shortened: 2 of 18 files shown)


nicolasvasilache:
> Because we told it not to care. 

Ok, does this mean you prefer to not apply the first commit "Apply TC coding conventions" or are you fine with this commit ?

> And since line breaks in TC mark the end of statement, we really don't want to split string literals unless we have ; or some string continuation symbol in TC

[This](https://github.com/facebookresearch/TensorComprehensions/pull/401/files#diff-f0df12aba9b032b995bfa66dba1d7c49) passes for me.

tensor_comprehensions/pybinds/pybind_options.cc
```diff
@@ -55,37 +55,49 @@ PYBIND11_MODULE(mapping_options, m) {
             }
             throw std::runtime_error("Invalid option passed");
           }),
-          "Initialize the mapping options from one of the following:\n 1. naive\n 2. pointwise\n 3. mlp\n 4. conv\n 5. group_conv\n 6. single_thread")
+          "Initialize the mapping options from one of the following:\n"
+          "1. naive\n 2. pointwise\n 3. mlp\n 4. conv\n 5. group_conv\n "
```
skimo-openhub: This drops a space in from of "1.".

Also, I would split the string after every "\n" (here and below).

nicolasvasilache:
> 625fdb8: why add it if it is not functional?

Because this is a synchronization PR to bring the internal and github repos to the same state.
I cannot just destroy people's internal tests.
Granted there should be an OSS PR to activate those tests in CI.
cc @salexspb @mingzhe09088

tc/c2/convolution_op.h
```diff
@@ -63,8 +63,6 @@ class TcConvolutionOp : public TcOp<T, Context, Engine> {
 
     this->tc_ = tc::makeConvolution2DTc(strideH, strideW);
     this->tcName_ = tc::CONVOLUTION2D_TC_NAME;
-    this->gradTc_ = tc::makeConvolution2DGradTc(strideH, strideW);
-    this->gradTcName_ = tc::CONVOLUTION2D_GRAD_TC_NAME;
```
skimo-openhub: Why is this removal part of adding support for gradient computations?
Did this code do anything useful before?
If not, it should be removed first in a separate commit.

This change also removes the single uses of `CONVOLUTION2D_GRAD_TC_NAME` and `makeConvolution2DGradTc` so they should be removed together with these lines.

tc/c2/tc_op.h
```diff
@@ -102,12 +101,8 @@ class TcOp : public Operator<Context> {
...
       }
 
-      auto parsedTcs = tc::detail::parse(tc_);
-      CHECK_EQ(parsedTcs.count(tcName_), 1u)
-          << "attempting to access undefined function " << tcName_;
-
-      auto outTensorInfo = tc::detail::inferOutputTensorInfo(
-          parsedTcs.at(tcName_), rawInputDLTensors_);
+      auto outTensorInfo =
+        tc::inferOutputTensorInfo(tc_, tcName_, rawInputDLTensors_);
```
skimo-openhub: independent change?

skimo-openhub:
On Mon, May 07, 2018 at 12:49:34PM +0000, Nicolas Vasilache wrote:
> > 625fdb8: why add it if it is not functional?
> 
> Because this is a synchronization PR to bring the internal and github repos to the same state.
> I cannot just destroy people's internal tests.

I think I'm missing some context here.
Why can't the internal repo have some extra test that
is not available in the github repo?

Or is the purpose to abolish the internal repo completely
such that only one repo is left?

skimo

tc/c2/tc_op.h
nicolasvasilache: spurious change actually, thanks!

nicolasvasilache: my bad, I misunderstood until I saw it in github, please ignore ..

tc/c2/tc_op.h
```diff
@@ -57,10 +57,9 @@ class TcOp : public Operator<Context> {
       setupNaiveCudaMappingOptions();
     }
 
-    if (args.HasArgument("gradCudaMappingOptions")) {
-      gradCudaMappingOptions_ =
-          tc::CudaMappingOptions(args.GetSingleArgument<std::string>(
-              "gradCudaMappingOptions", "ERROR"));
+    if (args.HasArgument("gradMappingOptions")) {
+      gradCudaMappingOptions_ = tc::CudaMappingOptions(
+          args.GetSingleArgument<std::string>("gradMappingOptions", "ERROR"));
```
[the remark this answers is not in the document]
nicolasvasilache: Introduced a new commit that drops all the dead code related to gradients before introducing @mingzhe09088 's implementation.
As a consequence there is no more name change but introduction with the proper name.

tc/c2/convolution_op.h
nicolasvasilache: addressed by introducing a commit that removes all the dead code related to inexistent gradient support. Then @mingzhe09088 's commit introduces proper C2 gradient support.

PR #401, opened 2018-05-05
Add test for TC with arbitrary newlines

nicolasvasilache:


test/cuda/test_corner_cases.cc
```diff
@@ -282,6 +282,38 @@ TEST(TestCornerCases, E23) {
       at::Scalar(d[0]).toFloat());
 }
 
+TEST(TestCornerCases, E24) {
```
skimo-openhub: Please also explain what corner case is being tested in a comment.

nicolasvasilache:
@caffe2bot retest this please

test/cuda/test_corner_cases.cc
```diff
@@ -282,6 +282,44 @@ TEST(TestCornerCases, E23) {
       at::Scalar(d[0]).toFloat());
 }
 
+// This tests that the TC parser functions in the presence of arbitrary newlines
```
ftynse: parser -> _parses_

skimo-openhub:
On Mon, May 07, 2018 at 08:59:57PM +0000, ftynse wrote:
> > @@ -282,6 +282,44 @@ TEST(TestCornerCases, E23) {
>        at::Scalar(d[0]).toFloat());
>  }
>  
> +// This tests that the TC parser functions in the presence of arbitrary newlines
> 
> parser -> _parses_

I think the original is correct (where "functions" is a verb).
If you take "functions" to be a noun, then you'd also have to drop
"the" in front of "TC".

skimo

PR #403, opened 2018-05-07
Add group normalization example

nicolasvasilache:
This adds a simple group normalization example, the original TC is courtesy of @BIT-silence who wrote an internal python test based on https://arxiv.org/pdf/1803.08494.pdf.

This PR puts it in our C++ examples template and reworks the TC to simulate inlining which brings about 20% perf empirically. I also hardcoded the result of a longer tuning run (100 candidate / 25 generations) .

On my Pascal GPU this runs in 1.67 ms.

This is not yet tested or benchmarked against @BIT-silence's implementation in Caffe2.


tc/examples/group_normalization.cc
```diff
@@ -0,0 +1,165 @@
...
+        .mapToThreads(48, 6)
+        .mapToBlocks(256, 32)
+        .useSharedMemory(true)
+        .usePrivateMemory(true)
+        .unrollCopyShared(false);
+
+template <typename Backend>
+at::Tensor makeATenTensor(at::ArrayRef<long int> sizes);
+
+template <>
```
ftynse: This code keeps proliferating, I suggested to move it somewhere in the commit that introduced the first version.  I understand we want self-contained examples, but having a local `.h` is self-contained enough for C++ code IMO.

tc/examples/group_normalization.cc
```diff
@@ -0,0 +1,165 @@
...
+#include "tc/core/cuda/cuda_tc_executor.h"
+#include "tc/core/flags.h"
+
+DEFINE_string(proto_path, "", "Filename to load and store proto cache ");
+DEFINE_bool(
+    use_best_options,
+    false,
+    "Start from hardcoded best options; if false start from naive options ");
+
+auto previouslyTunedBestOptions =
```
ftynse: Can we add a comment for which GPU model this was found?

tc/examples/group_normalization.cc
```diff
@@ -0,0 +1,165 @@
...
+    mean(n, g) +=! I(n, g, r_d, r_h, r_w)
+     var(n, g) +=! I(n, g, r_d, r_h, r_w) * I(n, g, r_d, r_h, r_w)
+
+# Literal translation from paper (https://arxiv.org/pdf/1803.08494.pdf)
+#    mean(n, g)    = mean(n, g) * <scale>
+#    var(n, g)  = rsqrt(var(n, g) * <scale> - mean(n, g) * mean(n, g) + 1e-5)
+#    O(n, g, d, h, w) =
+#        gamma(g, d) * (I(n, g, d, h, w) - mean(n, g)) * var(n, g) + beta(g, d)
+
+# Simulate forced inlining and write it as follows results in better schedules
```
ftynse: we found it depended on sizes and/or GPU ;)

tc/examples/group_normalization.cc
```diff
@@ -0,0 +1,165 @@
...
+  while (true) {
+    auto pos = tc.find(std::string("<scale>"));
+    if (pos == std::string::npos) {
+      break;
+    }
+    tc = tc.replace(pos, std::string("<scale>").size(), std::to_string(scale));
+  }
+
+  // 4. Run autotuning with evolutionary search starting from a naive option.
+  auto naiveOptions = FLAGS_use_best_options
```
ftynse: maybe just `options`, they are naive only if flag is set

PR #408, opened 2018-05-07
Fix fbcode issues

nicolasvasilache:
This list of PRs fixes issues found when sync'ing github -> fbcode.
In particular I tested that the python test_c2.py and the benchmark pass.
Compilation in opt mode is still WIP but this is currently progress.


nicolasvasilache:
Jenkins seems to be in limbo again, these are simple compilation-only changes, landing and we'll test it live.

PR #410, opened 2018-05-08
finish handling of single node schedule trees

skimo-openhub:
Prior to 653ecbb0 (halide2isl::makeScheduleTree: use isl::schedule
for constructing tree, Thu Mar 22 11:32:03 2018 +0100),
makeScheduleTree refused to construct a schedule tree
that consists of a single (domain) node.
Since then, inputs that result in such a tree are handled
by makeScheduleTree, but makeScheduleConstraints still
assumed that a tree contains at least two nodes.
Remove that assumption, allowing TestCornerCases.E4
to be enabled.

### DIFF
--- a/tc/core/polyhedral/scop.cc
+++ b/tc/core/polyhedral/scop.cc
@@ -332,8 +332,6 @@ isl::schedule_constraints makeScheduleConstraints(
     const Scop& scop,
     const SchedulerOptionsView& schedulerOptions,
     isl::union_set restrictDomain = isl::union_set()) {
-  auto firstChildNode = scop.scheduleRoot()->child({0});
-
   auto constraints = isl::schedule_constraints::on_domain(scop.domain())
                          .set_validity(scop.dependences)
                          .set_proximity(scop.dependences)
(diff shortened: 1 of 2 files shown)


tc/core/polyhedral/scop.cc
```diff
@@ -332,17 +332,17 @@ isl::schedule_constraints makeScheduleConstraints(
...
                          .set_coincidence(scop.dependences);
   if (restrictDomain) {
     constraints = constraints.intersect_domain(restrictDomain);
   }
-  if (auto contextNode =
-          firstChildNode->elemAs<detail::ScheduleTreeElemContext>()) {
+  auto root = scop.scheduleRoot();
+  if (auto contextNode = root->numChildren() > 0
+          ? root->child({0})->elemAs<detail::ScheduleTreeElemContext>()
+          : nullptr) {
```
nicolasvasilache: looks like you can still dereference a nullptr?

nicolasvasilache: never mind, confused myself over nothing.
Can we write this instead:
```
auto contextNode = root->numChildren() > 0
          ? root->child({0})->elemAs<detail::ScheduleTreeElemContext>()
          : nullptr;
if (contextNode) {
  ...
}
```

PR #415, opened 2018-05-10
Apparent regression in full tile separation and reduction matching(Volta)

nicolasvasilache:
```
./test/cuda/test_tc_mapper_bugs --gtest_filter="*GroupNorm*" --debug_tc_mapper=1 --logtostderr=1
```

prints in particular:

```
if (t1 <= 3 && t0 <= 1) {
  for (int c0 = 0; c0 <= 3; c0 += 1) {
    acc_1 = 0.000000f;
    for (int c3 = 0; c3 <= 5; c3 += 1) {
      for (int c4 = 0; c4 <= 5; c4 += 1) {
        acc_1 = (acc_1 + I[c0][t1][t0][c3][c4]);
      }
    }
    __tc::CubReduceAlongX<6,6,1,__tc::ReductionOp::Sum>(&mean[c0][t1], acc_1);
  }
}
```

This then naturally deadlocks on execution (on Volta)

Here is a gist of the full log:
https://gist.github.com/nicolasvasilache/ece8a0e58c0a7ed03588485576960bcf

Note that the log shows:
I0509 20:20:18.553678 75489 mapped_scop.cc:295] Full/partial tile separation not needed because all tiles are full.

then immediately after:
```
...
I0509 20:20:18.790910 75489 mapped_scop.cc:1017] After mapping to threads
...
[D, G, H, N, W] -> { S_0[mean_s0_n, mean_s0_g] : D = 2 and G = 4 and H = 6 and N = 4 and W = 6 and 0 <= mean_s0_n <= 3 and 0 <= mean_s0_g <= 3 }
context([D, G, H, N, W, t1, t0, t2, b2, b1, b0] -> { [] : t2 = 0 and b2 = 0 and b1 = 0 and 0 <= t1 <= 11 and 0 <= t0 <= 5 and 0 <= b0 <= 7 })
mapping_filter(ids(t1, t0, ) [D, G, H, N, W, t1, t0] -> { S_0[mean_s0_n, mean_s0_g] : (-t1 + mean_s0_n) mod 12 = 0 and (-t0 + mean_s0_g) mod 6 = 0 and 0 <= t1 <= 11 and 0 <= t0 <= 5 })
```

This is likely related to / coming from 548c636725d7ae3e620143476246a5b1edb4fc0b (cc @ftynse)
but it seems we didn't have a test case until now.
Going to confirm I see similar behavior on Pascal

### DIFF
--- a/test/cuda/test_tc_mapper_bugs.cc
+++ b/test/cuda/test_tc_mapper_bugs.cc
@@ -774,6 +774,51 @@ TEST(Convolution, NestedExpressions) {
   CHECK_EQ(at::Scalar(B[10]).toFloat(), 1);
 }
 
+TEST(GroupNorm, ReductionDeadlockBug) {
+  auto group_norm = "group_norm";
+  auto TC = std::string(R"TC(
+def group_norm(
+    float(N, G, D, H, W) I, float(G, D) gamma, float(G, D) beta)
+    -> (O, mean, var)
+{
+    mean(n, g) +=! I(n, g, r_d, r_h, r_w)
+     var(n, g) +=! I(n, g, r_d, r_h, r_w) * I(n, g, r_d, r_h, r_w)
+    O(n, g, d, h, w) = gamma(g, d)
+      * ( I(n, g, d, h, w) - mean(n, g) * 1.0 )
+      * rsqrt( var(n, g) * 1.0
+            - mean(n, g) * mean(n, g) * 1.0 * 1.0
+            + 1e-5)
+      + beta(g, d)
+}
+  )TC");
+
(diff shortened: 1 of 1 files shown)


nicolasvasilache:
Interestingly, there is **no deadlock on Pascal**
Results are probably wrong though, we need a check too.

ftynse:
The problem with reduction below a branch appears before 548c636725d7ae3e620143476246a5b1edb4fc0b, which is reasonable because that commit is essentially code motion (construct the mapping filter internally to schedule tree node construction, rather than externally before calling that construction).

ftynse:
I backtracked a little bit, but have not finished.  The problem still appears on f54c5dd, if somebody wants to bisect further.

PR #418, opened 2018-05-11
insert reduction synchronization outside thread mapping

skimo-openhub:
Regular synchronization should never appear underneath a thread mapping
since the synchronization should be performed by all threads and
the mapping to threads may leave some thread instances unmapped.
Inserting reduction synchronization was apparently deemed safe
because the partial tile separation makes sure only complete
blocks are mapped to reductions.
However, by having the synchronization inside the mapping,
the isl AST generator may generate tests outside this synchronization
that involve thread identifiers (even if it is known to
the user that those same conditions could be represented
without involving thread identifiers, in combination with
other constraints in the code).
Insert the synchronization outside the mapping to prevent
this from happening.  This also means that the reduction
member no longer needs to be split off, such that
the thread mapping now always corresponds to a single band.

Note that while the partial tile separation makes sure
that only complete blocks are mapped to reductions,
multiple such complete blocks may still get mapped
by the thread mapping, including in the parallel
directions.  The current reduction handling does not
support this as it stores the partial reductions
in a single (per-thread) scalar variable.
The band mapped to threads therefore needs to be tiled
first such that it contains exactly one complete block
in the parallel directions.

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -390,7 +390,7 @@ size_t MappedScop::mapToThreads(detail::ScheduleTree* band) {
   mapThreadsBackward(band);
 
   if (isReduction) {
-    splitOutReductionAndInsertSyncs(band, nMappedThreads - 1);
+    splitOutReductionAndInsertSyncs(band);
   }
 
   return numThreads.view.size();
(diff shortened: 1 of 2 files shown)


nicolasvasilache:
I am having trouble understanding this sentence:

> Note that while the partial tile separation makes sure
> that only complete blocks are mapped to reductions,
> it may still map multiple blocks, including in the parallel
> directions.

Trying to decompose I read "partial tile separation may still map multiple blocks" which I cannot make sense of.

skimo-openhub:
On Fri, May 11, 2018 at 09:54:58AM -0700, Nicolas Vasilache wrote:
> I am having trouble understanding this sentence:
> 
> > Note that while the partial tile separation makes sure
> > that only complete blocks are mapped to reductions,
> > it may still map multiple blocks, including in the parallel
> > directions.
> 
> Trying to decompose I read "partial tile separation may still map multiple blocks" which I cannot make sense of.

Is this better?

    Note that while the partial tile separation makes sure
    that only complete blocks are mapped to reductions,
    multiple such complete blocks may still get mapped
    by the thread mapping, including in the parallel
    directions.

skimo

PR #421, opened 2018-05-13
Add support for readonly loads

nicolasvasilache:
This is rebased on another PR, please only look at the last commit.

CUDA supports a faster readonly cache for data that is never written.
This commits adds support in cuda/codegen to generate such accesses.
This is achieved by simply inspecting the schedule tree for all readonly
tensor references and saving their isl::id in a set that is passed in the
context.

A potential issue to address is that the Halide-related emitMappedTensorAccess
function only knows about tensor names but not isl::id. This should be fine
because a string is an id in Halide.

Another potential issue to address is that
`TensorReferenceGroup::accessedBySubtree` seemingly requires a `MappingFilter`
but it seems it would be useful if it also worked on the root (cc @ftynse).

The corresponding CudaMappingOptions are also added to control triggering of
the option.

A simple test is also added.


nicolasvasilache:
duplicate

PR #423, opened 2018-05-14
Benchmarks

nicolasvasilache:
Rebased on top of d139214aacc1563a9f343f1df6e37591d0d35fec, please ignore the first 3 commits.


ftynse:
Rebased on master after removing irrelevant commits (which got merged in separate PRs).

One thing I don't understand about your rework of benchmarks: why do you declare and define the options in headers?  Are they expected to be reused?  How? (I don't expect us recommending to `#include "../benchmarks/convolution.h"` and using some very long and obscure global variable name)  If not, I would ditch the headers and have options declared and defined in each test.  It already works that way in tests.

nicolasvasilache:
@ftynse I would claim the current structuring serves 2 purposes: (1) keep the benchmark itself readable and put the "configuration" in the .h, (2) as a stepping stone towards trying to have a centralized location where we could hold those configs and build a sparsely populated database without shipping a proto.

Even without (2), I would argue that separating the logic of the benchmark from the configuration of the best options found by tuning makes sense.

ftynse:
I would buy this better if it avoided ridiculously similar tests in the code and the dangers of accidentally using "group_convolution_N_32_F_32_C_32_KW_3_options" in "TEST_group_convolution_N_32_F_23_C_32_KW_3".

nicolasvasilache:
@caffe2bot retest this please

PR #424, opened 2018-05-14
[DO NOT MERGE YET] experimental benchmarks drop caffe2

nicolasvasilache:
Same as pr/benchmarks but trying to use the caffe2 from inside pytorch which comes with extra issues around proto.


ftynse:
Useful parts cherry-picked in #423, #437.  Caffe2 is now taken from conda.  Closing.

PR #427, opened 2018-05-14
drop checkValidIslSchedule

skimo-openhub:
This function dates back from a time in prehistory when
internal isl data structures (schedule trees) were
being created from the outside, subverting the internal checks
in isl.  Part of the removed comments still mentions
the possibility of constructing an invalid isl schedule.
Since then, the isl schedule tree construction
has been changed to use the proper interface.
There is therefore no longer any need to check that the constructed
isl schedule tree would be valid since isl should perform
the appropriate checks.  If any of those checks are missing
from isl then they should be added there.

The removed function is too low-level, incomplete (FIXMEs) and
has a high risk of getting out of sync with isl.
More generally, it does not make sense for TC to guess
whether the result of a conversion to isl schedule trees
would be valid if it does not control the conversion.
It may make sense to add a function that checks
whether it is possible to convert a TC schedule tree
to an isl schedule tree instead.


ftynse:
ok, let's reintroduce parts of the checks to TC tree -> isl tree conversion if the error messages in isl tree construction are not clear enough

PR #429, opened 2018-05-16
Add argument RuntimeInformation to uncheckedRun and run, so CUDA stre…

mingzhe09088:
Currently, TC executor and C2 are using different CUDA streams. This has lead to a synchronization issue. To resolve this issue, this PR passes the CUDA stream id from C2 to TC.

### DIFF
--- a/tc/c2/tc_op.h
+++ b/tc/c2/tc_op.h
@@ -119,11 +119,15 @@ class TcOp : public Operator<Context> {
       compiled_ = true;
     }
 
+    // Get CUDA stream id from C2
+    tc::CudaBackend::RuntimeInformation info;
+    info.stream = context_.cuda_stream();
+
     // run
     if (!check_sizes_) {
-      executor_->uncheckedRun(input_void_ptrs_, output_void_ptrs_);
+      executor_->uncheckedRun(input_void_ptrs_, output_void_ptrs_, info);
     } else {
-      executor_->run(raw_input_dl_tensors_, raw_output_dl_tensors_);
+      executor_->run(raw_input_dl_tensors_, raw_output_dl_tensors_, info);
     }
     return true;
   }
--- a/tc/core/cpu/cpu_tc_executor.cc
+++ b/tc/core/cpu/cpu_tc_executor.cc
@@ -52,7 +52,8 @@ CpuCompilationResult CpuBackend::compileWithTcMapper(
 
 void CpuTcExecutor::uncheckedRun(
     const std::vector<const void*>& inputs,
-    const std::vector<void*>& outputs) const {
+    const std::vector<void*>& outputs,
+    typename CpuBackend::RuntimeInformation info) const {
   LOG(ERROR) << "NYI: CpuTcExecutor::uncheckedRun";
 }
 
--- a/tc/core/cuda/cuda_backend.h
+++ b/tc/core/cuda/cuda_backend.h
@@ -44,7 +44,11 @@ struct CudaCompilationResult {
  * synchronization information of a kernel.
  */
 struct CudaRuntimeInformation {
-  cudaStream_t stream{0};
+ public:
+  CudaRuntimeInformation() : stream(0) {}
+
+ public:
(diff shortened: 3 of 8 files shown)


tc/c2/tc_op.h
```diff
@@ -119,11 +119,15 @@ class TcOp : public Operator<Context> {
       compiled_ = true;
     }
 
+    // Get CUDA stream id from C2
+    tc::CudaBackend::RuntimeInformation info;
+    info.stream = {context_.cuda_stream()};
```
nicolasvasilache: those braces aren't required, just  `info.stream = context_.cuda_stream()`

tc/core/cpu/cpu_tc_executor.cc
```diff
@@ -52,7 +52,8 @@ CpuCompilationResult CpuBackend::compileWithTcMapper(
 
 void CpuTcExecutor::uncheckedRun(
     const std::vector<const void*>& inputs,
-    const std::vector<void*>& outputs) const {
+    const std::vector<void*>& outputs,
+    typename CpuBackend::RuntimeInformation info) const {
```
nicolasvasilache: I'm conflicted here:
- by value makes sense because of asynchrony and not wanting to hold reference to local object that may go away. - OTOH the local object is used to launch the kernel synchronously on the CPU so const reference + the above is not a problem for GPUs but may be a problem in the future for other architectures.

So I'm tempted to leave by value, I think this at least deserves some documentation in the .h

tc/core/cuda/cuda_tc_executor.h
```diff
@@ -39,7 +39,8 @@ class CudaTcExecutor : public TcExecutor<CudaBackend> {
   /// doesn't then segfault will likely occur.
   void uncheckedRun(
       const std::vector<const void*>& inputs,
-      const std::vector<void*>& outputs) const;
+      const std::vector<void*>& outputs,
+      typename CudaBackend::RuntimeInformation info = {}) const;
```
nicolasvasilache: I would explicitly initialize to `{0}`, default initialization will prob not work in the future when we add more runtime info

tc/core/tc_executor.h
```diff
@@ -88,7 +88,8 @@ class TcExecutor {
   /// advanced aliasing) properties of the input and output tensors.
   void run(
       const std::vector<const DLConstTensor*>& inputs,
-      const std::vector<const DLTensor*>& outputs) const;
+      const std::vector<const DLTensor*>& outputs,
+      typename Backend::RuntimeInformation info = {}) const;
```
nicolasvasilache: I would explicitly initialize to `{0}`, default initialization will prob not work in the future when we add more runtime info

tc/core/cuda/cuda_backend.h
```diff
@@ -45,6 +45,8 @@ struct CudaCompilationResult {
  */
 struct CudaRuntimeInformation {
   cudaStream_t stream{0};
+
+  CudaRuntimeInformation(): stream(0) {}
```
nicolasvasilache:
```
struct CudaRuntimeInformation {
public:
 CudaRuntimeInformation() : stream(0) {}
public:
  cudaStream_t stream;
};
```
as per our coding conventions (I think)

tc/core/cuda/cuda_tc_executor.h
```diff
@@ -37,9 +37,14 @@ class CudaTcExecutor : public TcExecutor<CudaBackend> {
   /// No tensor-related information can be checked so it is the user's
   /// responsibility to ensure that shapes and strides match. If the user
   /// doesn't then segfault will likely occur.
+  /// Pass by const reference is good for GPU now, but it may be a problem
```
nicolasvasilache: Please drop this sentence: `Pass by const reference is good for GPU now, but it may be a problem`

mingzhe09088:
@nicolasvasilache  I've updated the PR based on your comments. Seems like the build is pending. Is there any issue with the OSS CI?

PR #431, opened 2018-05-21
Fix Halide reduction detection

nicolasvasilache:
This PR is @math-fehr 's commit to fix reduction detection in Halide.
Please update the commit message to explain the fix and let's land it once tests have passed.

### DIFF
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -559,6 +559,7 @@ std::vector<Reduction> findReductions(const Stmt& s) {
           auto n = initVars[op->name].size();
           p.dims.resize(vars.size() - n);
           std::iota(p.dims.begin(), p.dims.end(), n);
+          result.insert(make_pair(op->name, reductions[op->name]));
         } else {
           reductions.erase(op->name);
         }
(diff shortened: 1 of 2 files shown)


math-fehr:
I cannot modify the pr, but here is my explanation for the fix:
To find a reduction, the algorithm is visiting all halide Block, and add a reduction if it finds a for loop doing
a reduction initialization, followed by a for loop doing the reduction update.
However, in the previous code, when visiting halide Block to find reductions, what followed the reduction 
initialization in the current halide Block was casted to a halide For, which worked only when there is
exactly one halide For following the initialization.
This meant that reductions where matched only when the update was the last for loop in the current
halide Block during the visit.

ftynse:
@math-fehr  than create a new one please, I'll close this one

nicolasvasilache:
Could you send a new PR with your explanations as a commit message?

I extracted your changes from my benchmarks stack but we can also abandon
this and use what you send.

Also could you please add a simple mean  test where you string compare the
expected cub matching?

Thanks!

math-fehr:
Yes, I'm working on it.
However when I'm adding the test, it won't pass, and I'm trying to find why.
For the moment I tried to add a modified version of Reduction1D in test_cuda_mapper, the problem
might be related to the parametric aspect of the code.

nicolasvasilache:
@math-fehr ISL scheduling will be affected by full JIT'ing of parameters as well as the `fixParametersBeforeScheduling(...)` option, maybe look there?

PR #432, opened 2018-05-21
Add support for __ldg

nicolasvasilache:
This PR is rebased on top of `Fix Halide reduction detection`.

CUDA supports a faster readonly cache for data that is never written.
This commits adds support in cuda/codegen to generate such accesses.
This is achieved by simply inspecting the schedule tree for all readonly
tensor references and saving their isl::id in a set that is passed in the
context.

~~A potential issue to address is that the Halide-related emitMappedTensorAccess
function only knows about tensor names but not isl::id. This should be fine
because a string is an id in Halide.~~

> in TC, isl::id are reference-counted strings

~~Another potential issue to address is that
`TensorReferenceGroup::accessedBySubtree` seemingly requires a `MappingFilter`
but it seems it would be useful if it also worked on the root (cc @ftynse).~~

> do not use that function here

The corresponding CudaMappingOptions are also added to control triggering of
the option.

A simple test is also added.

This will be used in the benchmarks.

### DIFF
--- a/tc/core/polyhedral/cuda/codegen.cc
+++ b/tc/core/polyhedral/cuda/codegen.cc
@@ -365,8 +365,32 @@ void emitReductionInit(
 }
 
 namespace {
+// RAII-style wrapper around CodegenStatementContext that wraps the output
+// streamed to context.ss into an "__ldg()" intrinsic, if the tensor
+// with the given identifier is known to be read-only.
+struct LdgWrapper {
+ public:
+  LdgWrapper(const CodegenStatementContext& context, isl::id id)
+      : readOnly_(context.readOnlySet.count(id) > 0), out_(context.ss) {
+    if (readOnly_) {
+      out_ << "__ldg(&";
+    }
+  }
+
+  ~LdgWrapper() {
+    if (readOnly_) {
+      out_ << ")";
+    }
+  }
+
+ private:
+  bool readOnly_;
+  std::ostream& out_;
+};
+
 template <typename AFF>
 void emitAccess(AFF access, const CodegenStatementContext& context) {
+  LdgWrapper ldgWrapper(context, access.get_tuple_id(isl::dim_type::out));
   context.ss << context.build().access_from(access).to_C_str();
 }
 } // namespace
@@ -584,6 +608,8 @@ void emitMappedTensorAccess(
 
   // Not promoted, emitting just the mapped subscript.
   if (!promotionInfo.groupId) {
+    auto ctx = context.scop().domain().get_ctx();
+    LdgWrapper ldgWrapper(context, isl::id(ctx, name));
     context.ss << name;
     for (auto e : subscripts) {
       context.ss << "[";
@@ -681,6 +707,29 @@ size_t& nAstNodes() {
   return n;
 }
 
+namespace {
+// Collect ids of tensors that are only read on the Scop.
+std::unordered_set<isl::id, isl::IslIdIslHash> gatherReadOnlySet(
+    const MappedScop& mscop) {
+  std::unordered_set<isl::id, isl::IslIdIslHash> readOnlySet;
+
+  if (!mscop.useReadOnlyCache) {
+    return readOnlySet;
+  }
+
+  const auto& scop = mscop.scop();
+
+  auto read = scop.reads.universe().range();
+  auto written = scop.writes.universe().range();
+  auto readOnly = read.subtract(written);
+  for (auto s : readOnly.get_set_list()) {
+    readOnlySet.emplace(s.get_tuple_id());
+  }
+
+  return readOnlySet;
+}
+} // namespace
+
 string emitCudaKernel(
     const std::string& specializedName,
     const MappedScop& mscop) {
@@ -745,7 +794,9 @@ string emitCudaKernel(
   auto root = mscop.schedule();
   astBuild = astBuild.set_iterators(Codegen::makeLoopIterators(root));
   auto astNode = astBuild.node_from(schedule);
-  AstPrinter(CodegenContext(ss, mscop, nodeInfoMap)).emit(astNode);
+
+  AstPrinter(CodegenContext(ss, mscop, nodeInfoMap, gatherReadOnlySet(mscop)))
+      .emit(astNode);
   ss << "}" << endl;
 
   return ss.str();
--- a/docs/source/mapping_options.rst
+++ b/docs/source/mapping_options.rst
@@ -75,6 +75,8 @@ The following options are currently available:
 
 * :code:`.unrollCopyShared(<boolean>)`: Also unroll the copies to and from shared memory introduced by the :code:`TC` mapper. If :code:`unroll` value is not provided, has no effect.
 
+* :code:`.useReaOnlyCache(<boolean>)`: Emit loads to the readonly cache when appropriate.
+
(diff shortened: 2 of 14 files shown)


skimo-openhub:
> A potential issue to address is that the Halide-related emitMappedTensorAccess
> function only knows about tensor names but not isl::id. This should be fine
> because a string is an id in Halide.

For the use in TC, an isl::id is just a (reference counted) string.
Please remove/reformulate this paragraph since it can only lead to confusion.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -584,12 +592,27 @@ void emitMappedTensorAccess(
 
   // Not promoted, emitting just the mapped subscript.
   if (!promotionInfo.groupId) {
+    // Almost certainly not the proper way but how do we get to a proper
+    // isl::id here?
```
skimo-openhub: Just construct it directly from the name (or use `string`s as keys of the `readOnlySet`)

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -681,6 +704,38 @@ size_t& nAstNodes() {
...
+      auto groupMap = TensorReferenceGroup::accessedBySubtree(t, scop);
+      for (const auto& kvp : groupMap) {
+        if (isReadOnly.count(kvp.first) == 0) {
+          isReadOnly[kvp.first] = true;
+        }
+        for (const auto& group : kvp.second) {
+          isReadOnly[kvp.first] &= group->isReadOnly();
+        }
+      }
+    }
```
skimo-openhub: This looks like you are abusing a function for something that it was not meant for.
(Not entirely sure because `accessedBySubtree` is not documented.)

Try `scop.reads.universe().range().subtract(scop.writes.universe().range())` instead.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -681,6 +697,24 @@ size_t& nAstNodes() {
   return n;
 }
 
+// Collect ids of tensors that are only read on the Scop.
+std::unordered_set<isl::id, isl::IslIdIslHash> gatherReadOnlySet(
+    const MappedScop& mscop) {
+  auto root = mscop.schedule();
```
skimo-openhub: unused variable

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -681,6 +697,24 @@ size_t& nAstNodes() {
   return n;
 }
 
+// Collect ids of tensors that are only read on the Scop.
+std::unordered_set<isl::id, isl::IslIdIslHash> gatherReadOnlySet(
```
skimo-openhub: Put in anonymous namespace.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -681,6 +697,24 @@ size_t& nAstNodes() {
...
+    const MappedScop& mscop) {
+  auto root = mscop.schedule();
+  const auto& scop = mscop.scop();
+  std::unordered_set<isl::id, isl::IslIdIslHash> readOnlySet;
+
+  if (mscop.useReadOnlyCache) {
```
skimo-openhub: I tend to prefer return early instead.
That is,
```
if (!mscop.useReadOnlyCache) {
  return readOnlySet;
}
```
although I guess in this case the body is small enough

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -681,6 +697,24 @@ size_t& nAstNodes() {
...
+    auto readOnlyAcceses =
+        scop.reads.universe().range().subtract(scop.writes.universe().range());
```
skimo-openhub: These are not accesses, so the variable name is not appropriate.
I'd also break it up a bit to make it easier to read.
```
auto read = scop.reads.universe().range();
auto written = scop.writes.universe().range();
auto readOnlyAcceses = read.subtract(written);
```

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -584,12 +592,20 @@ void emitMappedTensorAccess(
...
+    }
     context.ss << name;
     for (auto e : subscripts) {
       context.ss << "[";
       emitHalideExpr(e, context);
       context.ss << "]";
     }
+    if (readOnly) {
+      context.ss << ")";
+    }
```
skimo-openhub: Hmm... could we avoid the code duplication by having a `CodegenStatementContext` method that takes an id and an access string and that wraps it in `__ldg` when appropriate?

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -365,8 +365,32 @@ void emitReductionInit(
 }
 
 namespace {
+// RAII-style wrapper around CodegenStatementContext that wraps the output
+// streamed to context.ss into an "__ldg()" intrinsic, if the tensor
+// with the given identifier is known to be read-only.
+struct LdgWrapper {
+ public:
+  LdgWrapper(const CodegenStatementContext& context, isl::id id)
+      : readOnly_(context.readOnlySet.count(id) > 0), out_(context.ss) {
```
skimo-openhub: Hmm... this looks like an overly complicated way of wrapping the access as
it requires the user to print the access inside a scope.
IMO, it would be more transparent if the user would print to a string first (in the case
where it's not a single string already) and then have that wrapped.
But anyway, at least the duplication is gone.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -681,6 +707,29 @@ size_t& nAstNodes() {
...
 }
 
+namespace {
+// Collect ids of tensors that are only read on the Scop.
+std::unordered_set<isl::id, isl::IslIdIslHash> gatherReadOnlySet(
+    const MappedScop& mscop) {
+  std::unordered_set<isl::id, isl::IslIdIslHash> readOnlySet;
+
+  if (!mscop.useReadOnlyCache) {
+    return readOnlySet;
```
skimo-openhub: I fixed this.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -365,8 +365,32 @@ void emitReductionInit(
 }
 
 namespace {
+// RAII-style wrapper around CodegenStatementContext that wraps the output
+// streamed to context.ss into an "__ldg()" intrinsic, if the tensor
+// with the given identifier is known to be read-only.
```
skimo-openhub: I extended the documentation.

tc/core/polyhedral/cuda/codegen.cc
ftynse: I actually prefer returning a clearly empty container with `return {};`, otherwise you have to scan the code upwards to see there is nothing inside.

tc/core/polyhedral/cuda/codegen.cc
ftynse: That would require creating a new stringstream. printing to it, then taking a string from that stream and passing it to a function that prints into another stringstream.  The latter stream still being accessible in the context where the former must be used, that looked confusing.

tc/core/polyhedral/cuda/codegen.cc
ftynse: :+1:

tc/core/polyhedral/cuda/codegen.cc
skimo-openhub: Well, CI didn't like it.

PR #441, opened 2018-05-23
Fix reduction detection

math-fehr:
Fix reduction detection bug.

In a halide Block, if there was something after a reduction update, the reduction wasn't matched.


ftynse:
Supersedes #431

tc/core/halide2isl.cc
```diff
@@ -548,20 +551,16 @@ std::vector<Reduction> findReductions(const Stmt& s) {
...
-          auto n = initVars[op->name].size();
-          p.dims.resize(vars.size() - n);
-          std::iota(p.dims.begin(), p.dims.end(), n);
-        } else {
-          reductions.erase(op->name);
-        }
+      } else if (isValidReductionUpdate(op)) {
+        auto& p = reductions[op->name];
+        CHECK(p.init.defined())
+            << "Missing reduction init or (unsupported) multiple updates";
```
skimo-openhub: The error message should be updated.
The original code could end up removing the init in a previous update,
but the new code no longer removes reductions.

tc/core/halide2isl.cc
```diff
@@ -578,7 +577,9 @@ std::vector<Reduction> findReductions(const Stmt& s) {
 
   std::vector<Reduction> result;
   for (auto p : finder.reductions) {
-    result.push_back(p.second);
+    if (p.second.init.defined() && p.second.update.defined()) {
```
skimo-openhub: `init` is always defined if `update` is.

tc/core/tc2halide.cc
```diff
@@ -843,7 +843,9 @@ HalideComponents translateDef(const lang::Def& def, bool throwWarnings) {
     using IRMutator2::visit;
```
skimo-openhub: Please add a note to the documentation to explain why the "Temporary hack" is needed.
Alternatively, remove the hack by exploiting #448.

test/test_cuda_mapper.cc
```diff
@@ -851,6 +851,41 @@ def fun(float(N) I) -> (O) {
   EXPECT_TRUE(code.find("O[0] = (O") != std::string::npos);
 }
 
+/*
+ * Check that a 2D mean produce a reduction library call.
```
skimo-openhub: I think something is missing after "2d mean", perhaps "TC"?
Also "produce" -> "produces".

ftynse:
> The original code would only detect reductions if init and update are the only two statements in a block. The new code only detects them if they are the first two statements in a block. That's still not the general case.

There's a pitfall. From what I understand, Halide's Blocks are functional lists (there's `first` and `rest` fields, where `rest` may be a block).  So if you have 4 loops one after another, it is represented as
```
Block {
  For
  Block {
    For
    Block {
      For
      For
    }
  }
}
```

as we observed in practice, rather than as
```
Block {
  For
  For
  For
  For
}
```
as one might expect.  Given that the matching recursively visits all blocks, it should match the two "middle" loops as well.  I think this mandates a test though (@math-fehr)

skimo-openhub:
On Fri, May 25, 2018 at 05:37:44AM -0700, ftynse wrote:
> > The original code would only detect reductions if init and update are the only two statements in a block. The new code only detects them if they are the first two statements in a block. That's still not the general case.
> 
> as one might expect.  Given that the matching recursively visits all blocks, it should match the two "middle" loops as well.  I think this mandates a test though (@math-fehr)

FuseReductions matches on statement names "s0" and "s1".
I'm pretty sure that's not going to work if the init and the update
are not the first two statements.
Anyway, let's just get rid of all that init statement stuff instead.

skimo

ftynse:
Can we please have a test for reduction in the middle, @math-fehr ?
If it does not work, let's have it disabled, open an issue and enable it when finally fixed.

math-fehr:
There is now three tests, testing if the library call is inserted when there is:
- An instruction modifying the tensor after the reduction
- An instruction modifying the tensor before the reduction
- An instruction modifying the tensor before and after the reduction

As for now, only the first test is enabled. Most surely the two other tests will be fixed with #448, so I'm waiting this pr to be merged to enable these tests.

ftynse:
I rebased this on top of #448 locally, only the tests remain then.  However, the two disabled tests still fail.  Probably because of #454.  There is a piece of code in #448 that attempts to schedule non-reduction statements before reduction, but fails because of dependences.

PR #442, opened 2018-05-23
More reduction detection tests.

math-fehr:
Add more tests for the non-parametric case of reduction detection

### DIFF
--- a/test/test_cuda_mapper.cc
+++ b/test/test_cuda_mapper.cc
@@ -926,6 +926,78 @@ def fun(float(N, K) I, float(N) O0) -> (O) {
 )TC");
 }
 
+/*
+ * Check that a 2D mean with these parameters does not produce a library call.
+ * The call is not produced because the band is tiled by 32 and 512 threads are
+ * mapped to the band.
+ * In practice, check that the library call does not appear in the code.
+ */
+TEST_F(PolyhedralMapperTest, Mean2DNonParametric_512threads) {
+  string tc = R"TC(
+def fun(float(36864, 1024) I) -> (O) {
+    O(n) +=! I(n, r_n)
+    O(n) = O(n) / (1024)
+}
+)TC";
+  auto mappingOptions =
+      DefaultOptions()
+          .outerScheduleFusionStrategy(tc::FusionStrategy::Preserve3Coincident)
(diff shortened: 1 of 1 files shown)


tc/core/halide2isl.cc
```diff
@@ -573,11 +574,12 @@ std::vector<Reduction> findReductions(const Stmt& s) {
     // For each init node, the names of its outer For nodes.
     std::map<std::string, std::vector<std::string>> initVars;
     std::map<std::string, Reduction> reductions;
+    std::map<std::string, Reduction> result;
```
skimo-openhub: Why do you introduce a different field here?
The reductions are already collected in `reductions`.
Now, you may want to adjust the conditions under which reductions are
removed from consideration, but you should do it in a controlled way.

tc/core/tc2halide.cc
```diff
@@ -852,6 +854,17 @@ HalideComponents translateDef(const lang::Def& def, bool throwWarnings) {
...
         body = mutate(Block::make(first->body, body));
+        if (restAsBlock) {
+          return Block::make(
+              For::make(
+                  first->name,
+                  first->min,
+                  first->extent,
+                  first->for_type,
+                  first->device_api,
+                  body),
```
skimo-openhub: This looks like code duplication.
Can't you create the For first and then figure out if you need to put a block
around it?

test/test_cuda_mapper.cc
```diff
@@ -886,6 +886,78 @@ def fun(float(N, K) I) -> (O) {
   EXPECT_TRUE(code.find(kCUBReductionName) != std::string::npos);
 }
 
+/*
+ * Check that a 2D mean with fixed parameters mapped to 512 threads
+ * produce a reduction library call.
+ * In practice, check that the library call appears in the code.
+ */
+TEST_F(PolyhedralMapperTest, DISABLED_Mean2DNonParametric_512threads) {
```
skimo-openhub: This test looks incorrect.
The reduction band is the result of tiling by 32, so it can only be mapped
to 32 threads, while the number of threads in the X direction is set to 512.

ftynse:

ftynse:

PR #443, opened 2018-05-23
Adding two unit tests to add 2D tensors, with and without strides

protonu:
I added two unit tests. Both the tests add two 2D tensors. The first test has contiguous tensors, while the second unit test has strides.  I perform validation by looking for a for-loop with given conditions in the generated CUDA code (string). The test for the v1 (strides) passes for now, even though handling strides does not work.

### DIFF
--- a/test/cuda/test_tc_mapper.cc
+++ b/test/cuda/test_tc_mapper.cc
@@ -298,6 +298,42 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
 }
 
 ///////////////////////////////////////////////////////////////////////////////
+// TensorAddStrided
+//   O(n, m) += I0_view(n, m) * I1_view(n, m)
+///////////////////////////////////////////////////////////////////////////////
+TEST_F(TcCudaMapperTest, TensorAddStrided) {
+  N = 64;
+  M = 64;
+  at::Tensor I0 = at::CUDA(at::kFloat).rand({N, M});
+  at::Tensor I0_view =
+      I0.type().tensor().set_(*I0.storage(), 0, {N, M}, {1, 16});
+  at::Tensor I1 = at::CUDA(at::kFloat).rand({N, M});
+  at::Tensor I1_view =
+      I1.type().tensor().set_(*I1.storage(), 0, {N, M}, {1, 16});
+  std::vector<at::Tensor> inputs = {I0_view, I1_view};
+
+  static constexpr auto TC = R"TC(
+def tensoraddstrided(float(N, M) I0_view, float(N, M) I1_view) -> (O) {
+    O(n, m) += I0_view(n, m) + I1_view(n, m)
+}
+  )TC";
+
+  auto checkFun = [](const std::vector<at::Tensor>& ins,
+                     std::vector<at::Tensor>& outs) { return true; };
+  auto options = tc::CudaMappingOptions::makeNaiveMappingOptions();
+  auto name = "tensoraddstrided";
+  auto res = Check(TC, name, options, inputs, checkFun);
+  // This test should be modified  when strided tensors are handled
+  std::string expected =
+      "const float32 (*I0_view)[64] = "
+      "reinterpret_cast<const float32 (*)[64]>(pI0_view)";
+
+  ASSERT_NE(std::string::npos, res.second.find(expected))
+      << "In resulting code:\n"
+      << res.second << "\nfound unexpected: " << expected;
+}
+
+///////////////////////////////////////////////////////////////////////////////
 // Lookup Table
 //   O(b, n) +=! LUT(I(b, n), r_r)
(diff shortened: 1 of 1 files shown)


test/cuda/test_tc_mapper.cc
```diff
@@ -297,6 +297,67 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
   ::benchmarkKernelOptions(TC, name, inputs, options);
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// TensorAdd2D -- Version 0
+//   O(n, m) +=! I0(n, m) * I1(n, m)
+///////////////////////////////////////////////////////////////////////////////
+TEST_F(TcCudaMapperTest, TensorAdd2Dv0) {
```
nicolasvasilache: this test does not really test anything interesting at this point, I would remove it

test/cuda/test_tc_mapper.cc
```diff
@@ -297,6 +297,67 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
...
+          << "In resulting code:\n"
+          << res.second << "\nfound unexpected: " << expected;
+  ::benchmarkKernelOptions(TC, name, inputs, options);
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// TensorAdd2D  -- Version 1 with Strides
+//   O(n, m) +=! I0(n, m) * I1(n, m)
+///////////////////////////////////////////////////////////////////////////////
+TEST_F(TcCudaMapperTest, TensorAdd2Dv1) {
```
nicolasvasilache: TensorAddStrided

test/cuda/test_tc_mapper.cc
```diff
@@ -297,6 +297,67 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
...
+  N = 64; M=64;
+  at::Tensor I0 = at::CUDA(at::kFloat).rand({N,M});
+  at::Tensor I0_view = I0.type().tensor().set_(*I0.storage(), 0 /*offset*/,{N,M}/* size*/, {1,16}/*strides*/);
```
nicolasvasilache: you should run clang format on this or the CI will fail

test/cuda/test_tc_mapper.cc
```diff
@@ -297,6 +297,67 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
...
+  at::Tensor I1 = at::CUDA(at::kFloat).rand({N,M});
+  at::Tensor I1_view = I1.type().tensor().set_(*I1.storage(), 0 /*offset*/,{N,M}/* size*/, {1,16}/*strides*/);
+  //std::vector<at::Tensor> inputs = {I0, I1};
```
nicolasvasilache: no commented out code plz

test/cuda/test_tc_mapper.cc
```diff
@@ -297,6 +297,67 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
...
+}
+  )TC";  
+
+  auto checkFun = [](const std::vector<at::Tensor>& ins,
+		  std::vector<at::Tensor>& outs) { return true;};
+  auto options = tc::CudaMappingOptions::makeNaiveMappingOptions();
+  auto name = "tensoradd2dv1";
+  auto res = Check(TC, name, options, inputs, checkFun);
+  //This test should be modified  when strided tensors are handled
+  std::string expected = "for (int c2 = t1; c2 <= 31; c2 += 8)";
```
nicolasvasilache: this does not check strides, just that some loop is set
you want to look at the place where the case from pointer to multidimensional array occurs and create an assertion based on that.
At this time the cast will only take sizes into account and we will need it to take strides into account that do not yet exist.

test/cuda/test_tc_mapper.cc
```diff
@@ -297,6 +297,67 @@ def tensordot(float(N, C1, C2, H, W) I0, float(N, C2, C3, H, W) I1) -> (O) {
...
+  ASSERT_NE(std::string::npos, res.second.find(expected))
+          << "In resulting code:\n"
+          << res.second << "\nfound unexpected: " << expected;
+  ::benchmarkKernelOptions(TC, name, inputs, options);
```
nicolasvasilache: this does not seem right until strides are supported and the test can pass

test/cuda/test_tc_mapper.cc
protonu: Thanks! I see your point about asserting based on the cast to multi-dim array. I'll make that change.

protonu:
Adding a unit test for stride support. The current lack of support means this test will show incorrect strides. The check will be modified once support for strides is added.

PR #444, opened 2018-05-24
MappedScop::findBestSync: use proper mapping to threads and warps

skimo-openhub:
In particular, use mappings to threads and warps to determine
whether all dependences are within a thread or warp.
Do so instead of considering pairs of thread identifier parameters.
Duplicating thread identifier parameters is confusing at best and
relies on renaming parameters, which is frowned upon and which
will not be exported in the mainline isl C++ interface.
Duplicate parameters are confusing because there is only
one value for any given thread identifier at a given point
in the execution.
Furthermore, the filter assigning the thread mapping to parameters
is only relevant during AST generation.  Prior to that,
it is more natural to think in terms of the thread mapping itself.
This is what this commit does.
It is not only simpler, but also shorter.

It is not entirely clear why the intersection
with the context is needed, but this is what the original code does and
it is preserved by this commit.


tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -431,167 +431,107 @@ bool hasOuterSequentialMember(
...
-          setInFun.set_dim_name(isl::dim_type::param, dim, name + suffix);
+  auto space = isl::space(tree->ctx_, 0);
+  auto empty = isl::union_set::empty(space);
+  auto id = isl::id(tree->ctx_, kBlock);
+  space = space.named_set_from_params_id(id, nThread);
+  auto zero = isl::multi_val::zero(space);
+  auto domainToThread = isl::multi_union_pw_aff(empty, zero);
+
+  for (auto mapping : tree->collect(tree, ScheduleTreeType::MappingFilter)) {
+    auto mappingNode = mapping->elemAsBase<ScheduleTreeElemMappingFilter>();
```
ftynse: I don't think there's a need for `elemAsBase` here, just `elemAs` would suffice.
(Actually, this node type inheritance is a cognitive overhead...)

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -431,167 +431,107 @@ bool hasOuterSequentialMember(
...
-// This function returns a set because it might change in the future,
-// and take into account the blocks.
-isl::set getSameWarpConstraints(
-    isl::space space,
-    const std::string& suffix1,
-    const std::string& suffix2,
+/*
+ * Construct a mapping
+ *
+ *  block[x] -> warp[floor((x)/s_w)]
```
ftynse: please define what `s_w` is

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -431,167 +431,107 @@ bool hasOuterSequentialMember(
...
-  std::vector<std::pair<isl::id, unsigned>> mappingIds{
-      {TX, TX.mappingSize(block)},
-      {TY, TY.mappingSize(block)},
-      {TZ, TZ.mappingSize(block)}};
-
-  auto formula1 = getLinearizedThreadIdxFormula(space, block, suffix1);
-  auto formula2 = getLinearizedThreadIdxFormula(space, block, suffix2);
+  auto space = isl::space(ctx, 0);
+  auto id = isl::id(ctx, kBlock);
+  auto domain = space.named_set_from_params_id(id, block.view.size());
```
ftynse: I would use `domainSpace` and `warpSpace` here, or even `blockSpace` for the former.  There are too many places where `domain` is a `union_set` that you start assuming that type.

skimo-openhub:
@caffe2bot retest this please

PR #448, opened 2018-05-25
stop relying on reduction init statements

skimo-openhub:
Use dependence information instead.

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -215,9 +215,8 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
   // a single reduction for now.
   // Support for multiple reductions would require a check
   // that these reductions do not interfere with each other.
-  auto initsUpdates = reductionInitsUpdates(band->mupa_.domain(), scop());
-  auto inits = initsUpdates.first;
-  auto updates = initsUpdates.second;
+  auto domain = band->mupa_.domain();
+  auto updates = reductionUpdates(domain, scop());
   if (updates.n_set() != 1) {
     return false;
   }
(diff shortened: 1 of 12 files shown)


tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -230,12 +229,17 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
   if (!isReductionMember(member, updates, scop())) {
     return false;
   }
-  // Order the init statements (if any) before the update statements
+  // Order the other statements (if any) before the update statements
```
ftynse: Can't we have other statements that are not updates and should be ordered _after_ reduction updates?  Something like
```
O(i)  = 0 for i in 0:N     #order before
O(i) += A(i,j)
O(i)  = O(i) / M             #order after
```

with a schedule
```
band([{S1[i]->[i]; S2[i,j]->[i]; S3[i]->[i]}; {S1[i]->[0]; S2[i,j]->[j]; S3[i]->[M]}])
  sequence()
    filter(S1[i])
    filter(S2[i,j])
    filter(S3[i])
```

skimo-openhub:
On Fri, May 25, 2018 at 05:47:59AM -0700, ftynse wrote:
> > @@ -230,12 +229,17 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
>    if (!isReductionMember(member, updates, scop())) {
>      return false;
>    }
> -  // Order the init statements (if any) before the update statements
> +  // Order the other statements (if any) before the update statements
> 
> Can't we have other statements that are not updates and should be ordered _after_ reduction updates?  Something like

Sure. This commit doesn't pretend to handle all cases,
only that it handles more cases than before and (most
importantly) that it does not rely on explicit init statements.

skimo

ftynse:
Looks good.  Can you open issues for "reorder statements before/after the reduction based on dependences" and "use statement ids instead of tensor names in halide reduction matching", and mention them in the code to keep track of our observations?

ftynse:
addressed comments myself

ftynse:
Ordering before/after is now #454 

And cleaning up tc2halide is #455.

PR #457, opened 2018-05-29
tc2halide: properly tag reduction updates

skimo-openhub:
In particular, only tag reduction updates and only tag them once.
The original code would tag any non-initial write to a tensor
involved in a reduction and it would tag it possibly several times,
once for each reduction on that tensor.
Tag the reduction updates immediately to ensure that
only reduction updates get tagged and that they get tagged exactly once.

Multiply tagged expressions cause confusion in TC.
In particular, isSupportedReduction would fail to detect
the reduction and the code generation also assumes
that a reduction update is only tagged once.

### DIFF
--- a/tc/core/tc2halide.cc
+++ b/tc/core/tc2halide.cc
@@ -488,13 +488,15 @@ Expr reductionUpdate(Expr e) {
   return Call::make(e.type(), kReductionUpdate, {e}, Call::Intrinsic);
 }
 
+// Note that the function definitions created by translateComprehension may
+// contain kReductionUpdate intrinsics.  These may have to be removed
+// in order to be able to apply internal Halide analysis passes on them.
 void translateComprehension(
     const lang::Comprehension& c,
     const map<string, Parameter>& params,
     bool throwWarnings,
     map<string, Function>* funcs,
-    FunctionBounds* bounds,
-    vector<Function>* reductions) {
+    FunctionBounds* bounds) {
   Function f;
   auto it = funcs->find(c.ident().name());
   if (it != funcs->end()) {
@@ -589,8 +591,9 @@ void translateComprehension(
                                  << c.assignment()->range().text() << "\n";
   }
 
+  // Tag reductions as such
   if (c.assignment()->kind() != '=') {
-    reductions->push_back(f);
+    rhs = reductionUpdate(rhs);
   }
 
   // Bind any scalar params on the rhs to their parameter objects.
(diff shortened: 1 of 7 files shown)


skimo-openhub:
On Tue, May 29, 2018 at 03:54:20AM -0700, ftynse wrote:
> 
> typos "Multiply tagged expressions cause confusing in TC."  -> "Multiple tagged expressions cause confusion in TC".

Fixed.

skimo

tc/core/tc2halide.cc
```diff
@@ -593,8 +588,9 @@ void translateComprehension(
                                  << c.assignment()->range().text() << "\n";
   }
 
+  // Tag reductions as such
   if (c.assignment()->kind() != '=') {
-    reductions->push_back(f);
+    rhs = reductionUpdate(rhs);
```
abadams: This is way simpler than the old code. The only thing that gives me pause is that it means the reductionUpdate intrinsics now appear in the function definitions, rather than just the lowered stmt. This may break internal Halide analysis passes that run over the functions (e.g. predicting the range of values they could take on to bound things like the access to f in f(g(idx))), factoring associative reductions). I don't think we're using any of that machinery right now though. If we ever send this DAG of Funcs down the regular Halide lowering path instead (e.g. to hit some exotic architecture), then we'll need to run a mutator on the Function DAG that strips these intrinsics out. 

For this PR, probably just add a comment:

Note: wrapping the function definitions in this intrinsic will hinder some of the analysis passes in Halide's standard lowering path. If we ever use that, or more pieces of that, we should revisit this.

skimo-openhub:
On Tue, May 29, 2018 at 09:05:06AM -0700, Andrew Adams wrote:
> abadams commented on this pull request.
> >    if (c.assignment()->kind() != '=') {
> -    reductions->push_back(f);
> +    rhs = reductionUpdate(rhs);
> 
> This is way simpler than the old code. The only thing that gives me pause is that it means the reductionUpdate intrinsics now appear in the function definitions, rather than just the lowered stmt. This may break internal Halide analysis passes that run over the functions (e.g. predicting the range of values they could take on to bound things like the access to f in f(g(idx))), factoring associative reductions). I don't think we're using any of that machinery right now though. If we ever send this DAG of Funcs down the regular Halide lowering path instead (e.g. to hit some exotic architecture), then we'll need to run a mutator on the Function DAG that strips these intrinsics out. 
> 
> For this PR, probably just add a comment:
> 
> Note: wrapping the function definitions in this intrinsic will hinder some of the analysis passes in Halide's standard lowering path. If we ever use that, or more pieces of that, we should revisit this.

OK, please double-check.

skimo

PR #460, opened 2018-05-29
Tc check

### LINKED ISSUE #351: Implement CHECK macros as exceptions

ttheodor:
closes #351


skimo-openhub:
On Tue, May 29, 2018 at 08:37:19AM -0700, Theodoros Theodoridis wrote:
> closes #351 

I haven't looked at it in detail yet, but this looks great.

skimo

tc/core/check.cc
```diff
@@ -0,0 +1,39 @@
...
+  if (not additionalMsg_.empty()) {
+    ss << ": " << additionalMsg_;
+  }
+  throw std::runtime_error(ss.str());
+}
+
+Checker tc_check(bool condition, const char* filename, uint64_t lineno) {
+  if (condition) {
+    return Checker(condition, {}, {});
+  }
```
skimo-openhub: Why this special case?  Is it to avoid calling `makeLocation`?
Why not simply store `filename` and `lineno`?
Also, why does only this variant have such a special case?

tc/core/check.h
```diff
@@ -0,0 +1,111 @@
...
+      additionalMsg_.pop_back();
+    }
+    return *this;
+  }
+
+ private:
+  bool condition_;
+  std::string location_;
+  std::string baseErrorMsg_;
+  std::string additionalMsg_;
```
skimo-openhub: Why do you need to make a distinction between these two types of messages?

test/test_check.cc
```diff
@@ -0,0 +1,177 @@
...
+    std::stringstream expected;
+    expected << "Check failed [" << __FILE__ << ':' << __LINE__ + 2 << ']'
+             << ": 1+1=3";
+    ASSERT_THROW_WHAT(TC_CHECK(false) << "1+1=3", expected.str());
+  }
+}
+
+TEST(CHECK, Vector) {
+  std::stringstream expected;
+  expected << "Check failed [" << __FILE__ << ':' << __LINE__ + 3 << ']'
```
skimo-openhub: This looks fragile and probably compiler dependent.
Why can you assume that the reported line number is the last line of
the macro invocation?  Better make sure `ASSERT_THROW_WHAT` call
is within a single line in these tests.

test/test_check.cc
```diff
@@ -0,0 +1,177 @@
...
+  }
+}
+
+TEST(CHECK, LT) {
+  ASSERT_NO_THROW(TC_CHECK_LT(1, 2));
+  ASSERT_NO_THROW(TC_CHECK_LT(std::string("aaa"), std::string("baa")));
+  {
+    std::stringstream expected;
+    expected << "Check failed [" << __FILE__ << ':' << __LINE__ + 2
+             << "] 1 not less than " << 1;
```
skimo-openhub: Why is the second "1" outside the string (and not the first one)?
Similarly below.

tc/autotuner/genetic_search.cc
```diff
@@ -131,11 +131,11 @@ void dropInvalidConfigurations(GeneticSearch::Population& population) {
...
-  CHECK_LT(numberElites, maxPopulationSize);         \
-  CHECK(mutationRate >= 0 and mutationRate <= 100)   \
+  TC_CHECK_LT(numberElites, maxPopulationSize);         \
+  TC_CHECK(mutationRate >= 0 and mutationRate <= 100)   \
       << "the mutation rate (" << mutationRate       \
       << ") should be in the [0,100] interval";      \
-  CHECK(crossOverRate >= 0 and crossOverRate <= 100) \
+  TC_CHECK(crossOverRate >= 0 and crossOverRate <= 100) \
       << "the crossover (" << crossOverRate          \
       << ") rate should be in the [0,100] interval";
```
skimo-openhub: reformat

tc/autotuner/msg
```diff
@@ -0,0 +1 @@
+#259 switched from TC function names to using whole TCs (canonicalized) as part of compilation cache keys. However, the final step of the autotuner, that is retrieving the best found options, was not updated and it used just the function name. As a result, an empty optional<CudaMappingOptions> was always returned.
```
skimo-openhub: what's this?

tc/core/cuda/cuda_rtc.cc
```diff
@@ -20,6 +20,7 @@
 #include <cuda_runtime.h>
 #include <nvrtc.h>
 
+#include "tc/core/check.h"
```
skimo-openhub: Why only here (and a couple of other places)? Shouldn't every file that uses these calls have this include?

tc/core/tensor-inl.h
```diff
@@ -18,9 +18,10 @@
 #include <algorithm>
 #include <sstream>
 
-#include <glog/logging.h>
+//#include <glog/logging.h>
```
skimo-openhub: remove instead of commenting out

tc/core/check.cc
ttheodor: If the `condition` holds then the strings are not used, it's an "optimization". 

It was present in the rest of `tc_check_*` but I removed it because it didn't feel very useful. I forgot to remove this one.

tc/core/check.h
ttheodor: There's no need. I'll merge them.

test/test_check.cc
ttheodor: No reason. I'll fix it.

test/test_check.cc
ttheodor: Do we still have format checking enabled on CI? If yes, then it will complain.

tc/autotuner/msg
ttheodor: I must have accidentally committed this.

ttheodor:
> Υes.  Complain about what?

This:

> This looks fragile and probably compiler dependent.
> Why can you assume that the reported line number is the last line of
> the macro invocation? Better make sure ASSERT_THROW_WHAT call
> is within a single line in these tests.

skimo-openhub:
On Wed, May 30, 2018 at 06:26:23AM -0700, Theodoros Theodoridis wrote:
> > Υes.  Complain about what?
> 
> This:
> 
> > This looks fragile and probably compiler dependent.
> > Why can you assume that the reported line number is the last line of
> > the macro invocation? Better make sure ASSERT_THROW_WHAT call
> > is within a single line in these tests.

Why would it complain if you arrange for all ASSERT_THROW_WHAT calls
to stay within a line?

skimo

tc/core/check.h
ttheodor: On a second thought there is one use case:
```
TC_CHECK_X(...) << "A" << "B";
```

What I initially implemented will print the following:
```
Check failed [file:lineno] "what failed": AB
```
The `"what failed"` depends on the type of check, e.g. for `TC_CHECK_EQ(1,2)` it would be `1 not equal to 2`. This part of the output is always present.

Extra information (passed with `<<`) is optional. The reason I keep track of it in a separate string is to figure out if and where the separator (`:`) and the space before it  should be printed. 

There are two options:
1. The optional/extra information is maintained on a separate string.
2. The space and separator is appended on the first call to `operator<<` and there is extra state that keeps track of it (e.g., a boolean).

skimo-openhub:
On Thu, May 31, 2018 at 05:51:53AM -0700, Theodoros Theodoridis wrote:
> Extra information (passed with `<<`) is optional. The reason I keep track of it in a separate string is to figure out if and where the separator (`:`) and the space before it  should be printed. 

OK.

skimo

ttheodor:
All comments were addressed.

ttheodor:
Being able to specify the exception type makes a lot of sense. How about  adding "overloads" with an extra exception type parameter?  For example:
```
CHECK_LT(i,size, std::out_of_range);
```

nicolasvasilache:
@ttheodor sounds good to me but let's do it as a followup PR? No point in delaying landing this nice improvement.

ttheodor:
@nicolasvasilache since this was still unmerged I went ahead and implemented the checks with custom exception types.

nicolasvasilache:
@ttheodor  you should be able to rebase and land this now

ttheodor:
Rebased

nicolasvasilache:
Landing this now without extra reviews because it has been accepted multiple times and ends up being delayed over and over since everything conflicts with it.

tc/core/mapping_options-inl.h
```diff
@@ -161,7 +161,7 @@ MappingOptionsView& MappingOptionsView::scheduleFusionStrategy(
 
 MappingOptionsView& MappingOptionsView::scheduleFusionStrategy(
     const std::string& str) {
-  FusionStrategy fs;
```
skimo-openhub: Who complains about this and why?

ftynse: see #500, I did not understand that either.

nicolasvasilache: see #504 (it should show in the CI)

PR #461, opened 2018-05-29
halide2isl: findReductions: support multiple reductions on the same tensor

### LINKED ISSUE #455: Identify reductions by statement name rather than by tensor name
In particular, this should allow for reductions with multiple updates.

skimo-openhub:
The tensor name was being used as a key in a map that was used
to match init and update nodes and to store the results.
Since 33a3bb65 (halide2isl: do not require init statement
for reduction detection, Sun May 27 12:04:46 2018 +0200),
these init statements are no longer considered, so the map
is no longer needed.
Collect the reductions directly in a vector instead.

Closes #455

### DIFF
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -529,11 +529,10 @@ std::vector<Reduction> findReductions(const Stmt& s) {
           }
         }
         if (dims.size() > 0) {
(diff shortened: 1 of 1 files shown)


skimo-openhub:
Jenkins doesn't appear to be make any progress, but it was green before a formatting change.

PR #463, opened 2018-05-31
schedule_print.cc: print each map of an extension on a new line

ftynse:
This was omitted in the original commit in the prehistory that made
union_* members of schedule nodes print members of a union on separate
lines.  Since there is no isl_union_map_get_map_list, wrap the union_map
to obtain a union_set, get the list of sets and unwrap them to get the
maps back.


skimo-openhub:
On Thu, May 31, 2018 at 01:03:48PM +0000, ftynse wrote:
> This was omitted in the original commit in the prehistory that made
> union_* members of schedule nodes print members of a union on separate
> lines.  Since there is no isl_union_map_get_map_list, wrap the union_map

Aren't you about to merge that in?
Why not simply add an export and use it?

skimo

PR #465, opened 2018-05-31
Test valid strides before compiling

protonu:
…it test with incorrect strides which looks for a new exception.

### DIFF
--- a/tc/core/compiler.cc
+++ b/tc/core/compiler.cc
@@ -18,6 +18,7 @@
 #include <sstream>
 #include <string>
 
+#include "tc/core/exceptions.h"
 #include "tc/core/flags.h"
 #include "tc/core/halide_utils.h"
 #include "tc/core/tensor.h"
@@ -35,6 +36,19 @@ std::vector<TensorInfo> inferOutputTensorInfo(
 }
 
 namespace detail {
+
+inline void helpThrowInvalidStride(
+    const lang::TreeRef param,
+    const DLConstTensor* inputsInfo) {
+  lang::ErrorReport ep = lang::ErrorReport(param);
+  std::ostringstream err_c;
+  err_c << ep.what() << "compilation aborted: invalid strides in tensor "
+        << std::endl;
+  err_c << *(inputsInfo);
+  err_c << std::endl;
+  throw InvalidStrideException(err_c.str());
+}
+
 void checkInputsCompliant(
     const tc2halide::HalideComponents& halideComponents,
     const std::vector<const DLConstTensor*>& inputsInfo) {
@@ -66,6 +80,20 @@ void checkInputsCompliant(
       throw lang::ErrorReport(halideComponents.getDef().params()[i])
           << "expected a tensor with " << hdim << " dimensions but found "
           << dldim << " dimensions.";
+    }
+    auto dlstrides = inputsInfo[i]->strides;
+    auto dlsizes = inputsInfo[i]->shape;
+    if (dldim) {
+      if (dlstrides[dldim - 1] < 1) {
+        helpThrowInvalidStride(
+            halideComponents.getDef().params()[i], inputsInfo[i]);
+      }
+      for (size_t j = 0; j < dldim - 1; ++j) {
+        if (dlstrides[j] < dlstrides[j + 1] * dlsizes[j + 1]) {
+          helpThrowInvalidStride(
+              halideComponents.getDef().params()[i], inputsInfo[i]);
+        }
+      }
     }
   }
 }
--- a/tc/core/compiler.h
+++ b/tc/core/compiler.h
@@ -117,6 +117,7 @@ std::unique_ptr<typename Backend::ExecutorType> compile(
 std::vector<TensorInfo> inferOutputTensorInfo(
     lang::TreeRef tcDefinition,
     const std::vector<const DLConstTensor*> inputs);
+
 } // namespace detail
 } // namespace tc
 
--- a/tc/core/tensor.cc
+++ b/tc/core/tensor.cc
@@ -107,6 +107,48 @@ bool TensorInfo::operator!=(const TensorInfo& t) const {
   return !(*this == t);
 }
 
+std::ostream& operator<<(std::ostream& os, const TensorInfo& t) {
+  auto ndim = t.shape.size();
+  if (ndim == 0) {
+    return os;
+  }
+  os << "Dimensions: [";
+  if (ndim) {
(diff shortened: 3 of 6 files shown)


tc/core/tensor-inl.h
```diff
@@ -63,6 +63,25 @@ void DLTensorDeleter::operator()(const DLTensorType* t) {
   delete t;
 };
 
+template <typename DLTensorType>
+std::string printSizesAndStrides(const DLTensorType* t) {
```
nicolasvasilache: please make this an `operator<<` overload as I had originally suggested

tc/core/compiler.cc
```diff
@@ -67,6 +67,25 @@ void checkInputsCompliant(
           << "expected a tensor with " << hdim << " dimensions but found "
           << dldim << " dimensions.";
     }
+    auto dlstrides = inputsInfo[i]->strides;
+    auto dlsizes = inputsInfo[i]->shape;
+    lang::ErrorReport ep =
```
nicolasvasilache: no reason to do this in the critical path, you prob want a helper function that throws and call it once you have detected the faulty case

tc/core/compiler.cc
```diff
@@ -67,6 +67,25 @@ void checkInputsCompliant(
...
+        lang::ErrorReport(halideComponents.getDef().params()[i]);
+    std::ostringstream err_c;
+    err_c << ep.what() << "invalid strides for compilation in tensor "
+          << std::endl;
+    err_c << printSizesAndStrides(inputsInfo[i]);
+    err_c << std::endl;
+    if (dldim) {
+      for (size_t j = 0; j < dldim - 1; ++j) {
+        if (dlstrides[j] < dlstrides[j + 1] * dlsizes[j + 1]) {
+          throw InvalidStrideException(err_c.str());
```
nicolasvasilache: call helper function from here

tc/core/compiler.cc
```diff
@@ -67,6 +67,25 @@ void checkInputsCompliant(
...
+        }
+        if (dlstrides[dldim - 1] < 1) {
+          throw InvalidStrideException(err_c.str());
```
nicolasvasilache: call helper function from here

tc/core/tensor-inl.h
```diff
@@ -63,6 +63,25 @@ void DLTensorDeleter::operator()(const DLTensorType* t) {
...
+  std::ostringstream sizesandshapes;
+  sizesandshapes << "Dimensions of the tensor are: ";
```
nicolasvasilache: This is probably too much spew. 
How about `os << "sizes: "` ... `os << " strides: "` ... etc
Would be good to also print the base pointer, the offset and probably the type too

tc/core/compiler.cc
```diff
@@ -67,9 +67,37 @@ void checkInputsCompliant(
...
           << dldim << " dimensions.";
     }
+    auto dlstrides = inputsInfo[i]->strides;
+    auto dlsizes = inputsInfo[i]->shape;
+    if (dldim) {
+      for (size_t j = 0; j < dldim - 1; ++j) {
+        if (dlstrides[j] < dlstrides[j + 1] * dlsizes[j + 1]) {
+          helpThrowInvalidStride(halideComponents, inputsInfo, i);
+        }
+        if (dlstrides[dldim - 1] < 1) {
```
nicolasvasilache: this is independent of j, let's hoist it out

tc/core/compiler.h
```diff
@@ -117,6 +119,12 @@ std::unique_ptr<typename Backend::ExecutorType> compile(
 std::vector<TensorInfo> inferOutputTensorInfo(
     lang::TreeRef tcDefinition,
     const std::vector<const DLConstTensor*> inputs);
+
+void helpThrowInvalidStride(
```
nicolasvasilache: this should definitely not be exposed to the high-level API (at least not with this name and signature)

tc/core/tensor.cc
```diff
@@ -107,6 +107,48 @@ bool TensorInfo::operator!=(const TensorInfo& t) const {
...
+    }
+    os << std::endl;
+    os << "Strides: ";
+    if (t.strides.size()) {
+      for (size_t i = 0; i < ndim; ++i) {
+        os << t.strides[i] << "  ";
+      }
+    }
+  }
+  os << std::endl << "Type:";
```
nicolasvasilache: You don't want to force all these newlines, you want to give your caller the option to add a newline or not.
You want to print something like:
```
type: float base: 0x... offset: 0 sizes: [3, 4, 5] strides: [20 ,5, 1]
```

tc/core/compiler.cc
```diff
@@ -67,9 +67,37 @@ void checkInputsCompliant(
...
+      }
+    }
   }
 }
 
+void helpThrowInvalidStride(
+    const tc2halide::HalideComponents& halideComponents,
+    const std::vector<const DLConstTensor*>& inputsInfo,
+    size_t i) {
+  auto dlstrides = inputsInfo[i]->strides;
```
nicolasvasilache: dead code

tc/core/compiler.cc
```diff
@@ -67,9 +67,37 @@ void checkInputsCompliant(
...
+  auto dlsizes = inputsInfo[i]->shape;
```
nicolasvasilache: dead code

tc/core/compiler.cc
```diff
@@ -67,9 +67,37 @@ void checkInputsCompliant(
...
+        if (dlstrides[dldim - 1] < 1) {
+          helpThrowInvalidStride(halideComponents, inputsInfo, i);
+        }
+      }
+    }
   }
 }
 
+void helpThrowInvalidStride(
+    const tc2halide::HalideComponents& halideComponents,
```
nicolasvasilache: this signature looks unnecessarily convoluted to me, why not:
```
void helpThrowInvalidStride(TreeRef param, const DLConstTensor* info) {...}
```

tc/core/compiler.h
```diff
@@ -19,6 +19,8 @@
 #include <string>
 #include <vector>
 
+#include "tc/core/exceptions.h"
```
nicolasvasilache: why? this is useless

tc/core/compiler.h
```diff
@@ -19,6 +19,8 @@
...
+#include "tc/core/halide_utils.h"
```
nicolasvasilache: not worth it, just pass a TreeRef, or better yet do not expose the function and increase the compiler API

PR #467, opened 2018-06-01
More python bindings

nicolasvasilache:


tc/proto/mapping_options.proto
```diff
@@ -44,13 +44,12 @@ message MappingOptionsProto {
...
   // Unroll innermost loops to until at most "unroll" copies.  If not provided,
   // do not unroll.
-  optional uint64 unroll = 7;
-  // Apply tiling to imperfectly nested loops if possible.  Same tilesizes are
-  // used.
-  required bool tile_imperfectly_nested = 8;
-  // reserved 5, 6, 9 to 13; can only activate with proto3
-  // Reserved: 5, 6, 9-13 -> factored out into CudaMappingOptionsProto
-  required bool match_library_calls = 14;
+  optional uint64 unroll = 5;
```
ftynse: I think we did not want to reuse the numbers between different versions of proto, otherwise it attempts to match incompatible messages.

tensor_comprehensions/pybinds/tc.cc
```diff
@@ -287,24 +287,16 @@ PYBIND11_MODULE(tc, m) {
...
-      .def(
-          "unrollCopyShared",
-          &tc::CudaMappingOptions::unrollCopyShared,
-          "Also unroll the copies to and from shared memory. If an unroll "
-          "value is not provided, has no effect")
-      .def(
-          "useReadOnlyCache",
-          &tc::CudaMappingOptions::useReadOnlyCache,
-          "Use the readonly cache (i.e. emit __ldg loads)")
+          "toString",
```
ftynse: python has magic methods `__str__` and `__repr__` for this

tc/proto/mapping_options.proto
nicolasvasilache: That's in a world where proto is stable enough, I'd claim we're still in alpha stage and we don't save any protos.

tensor_comprehensions/pybinds/tc.cc
nicolasvasilache: They just print things like:
```
<method-wrapper '__str__' of tensor_comprehensions.tc.CudaMappingOptions object at 0x7f715d879420>
```

We want a human readable form of the options instead which is what `toString` does (and it was already present before btw)

ftynse: that's why you want to _define_ `__str__` rather than some `toString`, it will print what you return (and so will `str(yourobject)`)

tc/proto/mapping_options.proto
ftynse: I would go for a minimally intrusive change in this case, meaning don't change the numbers unless you have to

tensor_comprehensions/pybinds/tc.cc
nicolasvasilache: ah yes of course, I am rusty with python, thanks!

PR #469, opened 2018-06-02
emitCudaKernel: store parameter values in specialized scop and take them from there

skimo-openhub:
There is little point in trying to extract these values from a context
when those very same values were used to create the context.
Simply use the original known values.

Perform some additional clean-ups along the way.

### DIFF
--- a/tc/core/cuda/cuda_tc_executor.cc
+++ b/tc/core/cuda/cuda_tc_executor.cc
@@ -91,8 +91,7 @@ CudaCompilationResult CudaBackend::compileWithTcMapper(
   LOG_IF(INFO, FLAGS_debug_tc_mapper) << "Mapped schedule:" << std::endl
                                       << *(mappedScop->schedule());
 
-  auto parameters =
-      mappedScop->scop().getParameterValues(globalParameterContext);
+  auto parameters = mappedScop->scop().getParameterValues(pvm);
   auto specializedName = specializeKernelName(tcName, parameters);
 
   // This updates the launch bounds with the actual result from compilation
--- a/tc/core/cuda/cuda_tc_executor.cc
+++ b/tc/core/cuda/cuda_tc_executor.cc
@@ -78,8 +78,7 @@ CudaCompilationResult CudaBackend::compileWithTcMapper(
   auto scop = polyhedral::Scop::makeScop(
       isl::with_exceptions::globalIslCtx(), halideComponents);
   auto globalParameterContext = scop->makeContextFromInputs(inputs);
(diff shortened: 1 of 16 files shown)


skimo-openhub:
On Sun, Jun 03, 2018 at 11:19:00PM -0700, ftynse wrote:
> ftynse requested changes on this pull request.

What change do you want exactly?
Do you want globalParameterContext to be generated on the fly?

> As much as I like the cleanup, this change touches a style/policy-level question.  Until now, one of the key criteria for TC code was to avoid duplicate state.  It looks like this PR _introduces_ duplicate state between `parameterValues`, `globalParameterContext` and an eventual context node in the schedule tree.  I suggest we discuss a broader style issue so that everyone is on the same page and does not end up undoing each other's work.

This is similar to the "duplicate" state in a ScheduleTreeElemMappingFilter
where the filter_ is derived from the mapping, except that in this
case the globalParameterContext is not derived from parameterValues
alone.

skimo

ftynse:
> This is similar to the "duplicate" state in a ScheduleTreeElemMappingFilter where the filter_ is derived from the mapping.

Yes, and I'd like to get rid of that too, along with the extra complexity node type inheritance.  There is no need to do that anymore, we can just construct an isl filter node from tc mapping node, and have the `getFilterUnionSet` method in the latter for internal uses if necessary.

skimo-openhub:
On Mon, Jun 04, 2018 at 07:49:52AM -0700, ftynse wrote:
> > This is similar to the "duplicate" state in a ScheduleTreeElemMappingFilter where the filter_ is derived from the mapping.
> 
> Yes, and I'd like to get rid of that too, along with the extra complexity node type inheritance.  There is no need to do that anymore, we can just construct an isl filter node from tc mapping node, and have the `getFilterUnionSet` method in the latter for internal uses if necessary.

No argument there, but as I wrote in my message (not in your rewrite
above), this situation is slightly different.

skimo

ftynse:
> except that in this case the globalParameterContext is not derived from parameterValues alone.

what else does it need?

Generally, I'd prefer not having both `globalParameterContext` and `parameterValues` in the scop.

skimo-openhub:
On Mon, Jun 04, 2018 at 07:59:40AM -0700, ftynse wrote:
> > except that in this case the globalParameterContext is not derived from parameterValues alone.
> 
> what else does it need?
> 
> Generally, I'd prefer not having both `globalParameterContext` and `parameterValues` in the scop.

It also needs the list of all parameters to mark them non-negative,
in particular when no parameters have been fixed yet, but also when
only a subset of the paramters get fixed.

skimo

PR #470, opened 2018-06-04
Update installation instructions

nicolasvasilache:
This PR updates the installation instructions and the documentation.
In particular it kills the deprecated portions of the doc and moves the contents of BUILD.md to the sphinx docs.

Addresses #497 #498 #499


BUILD.md
```diff
@@ -1,5 +1,5 @@
 # Important notice
-***In order to uniformize and simplify the build system we had to make choices. TC is currently only officially supported on Ubuntu 16.04 with gcc 5.4.0, cuda 9.0 and cudnn 7.***
```
ftynse: Why removing cuda/cudnn versions?  Pytorch will result in dynamic linking errors with other versions of cuda, and caffe2 (if built) will complain about cudnn (and you need _7.1_ specifically).

BUILD.md
```diff
@@ -103,4 +80,41 @@ python -c 'import tensor_comprehensions'
 Run Python tests:
 ```
 ./test_python/run_test.sh
-```
\ No newline at end of file
+```
+
+At this point if things work as expected you can venture installing as follows (always a good idea to record installed files for easy removal):
```
ftynse: Comma after "at this point", and after the if-clause please.

BUILD.md
skimo-openhub: The document now recommends installing a version of pytorch that corresponds to the system version of cuda, so it makes sense to remove this version here.

batch_renormalization.py
```diff
@@ -0,0 +1,383 @@
+import time
```
ftynse: spurious file

PR #472, opened 2018-06-04
Support strided access in shared memory promotion

ftynse:
Depends on #477 

Original implementation of memory promotion ignored strides in accesses
to simplify the code.  isl recently introduced support for extracting
stride information from sets, making stride manipulation easy in TC.
Introduce special handling for strided support into TensorReferenceGroup
and related classes.  An access is strided if the access function has
the shape

`  (a_i - offset_i) = 0 mod stride_i`

where stride_i is a constant and offset_i is some affine expression on
the iteration domain.  Use isl to compute offsets and strides in access
relations.  Use this information to promote to shared memory only those
tensor elements that are actually read in case of strided accesses.
This decreases the amount of shared memory used by a kernel with such
accesses.

This also prepares for the introduction of shared memory promotion where
accesses of each thread individually are strided with the stride equal
to the number of threads.

Note that references accessing disjoint sets of elements with strides
are not grouped even if their non-strided footprints overlap, e.g.
`A[2*i]` and `A[2*i + 1]` belong to different groups.  This may decrease the
benefit of coalesced reads when copying between global and shared
memory.  At the same time, it also decreases the required shared memory
size making it to promote one of the references in case where a group
with two references would not fit.  The profitability of such grouping
requires further exploration.


tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -773,5 +774,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
...
+/*
+ * Extract a mapping from the domain elements active at "tree"
+ * to identifiers "ids", where all branches in "tree"
+ * are assumed to have been mapped to these identifiers.
+ * The result lives in a space of the form "tupleId"["ids"...].
+ */
+isl::multi_union_pw_aff extractDomainToIds(
+    const detail::ScheduleTree* tree,
+    const std::vector<mapping::MappingId>& ids,
+    isl::id tupleId = isl::id()) {
```
skimo-openhub: Why did you make the tuple id optional?
I was looking for other uses you have of this function later in this series,
but I can't find any, so maybe I should ask, why is this commit (d60c22af500fe82b1e06efd8197a5708b384a0ce) in this series?

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -799,11 +799,18 @@ isl::multi_union_pw_aff extractDomainToIds(
...
+      if (mappingNode->mapping.count(id) == 0) {
+        hasIds = false;
+        break;
+      }
       auto idMap = mappingNode->mapping.at(id);
       list = list.add(idMap);
     }
+    if (!hasIds) {
+      continue;
+    }
```
skimo-openhub: This changes the behavior, without modifying the documentation.
It also changes the behavior of extractMappingToThreads, which will
now silently ignore unmapped branches.

tc/external/detail/islpp-inl.h
```diff
@@ -38,10 +38,14 @@ inline isl::aff operator*(isl::aff A, isl::val V) {
   return V * A;
 }
 
+inline isl::aff operator/(isl::aff A, isl::val V) {
+  isl::aff T(isl::local_space(A.get_space().domain()), V);
```
skimo-openhub: This is needlessly complicated.  Let me send out some patches in my queue.

tc/external/detail/islpp-inl.h
```diff
@@ -192,6 +192,17 @@ inline isl::map operator<=(isl::aff_map A, isl::aff B) {
...
 
+///////////////////////////////////////////////////////////////////////////////
+// Operations on isl::multi_aff
+///////////////////////////////////////////////////////////////////////////////
+inline isl::multi_aff operator/(isl::multi_aff left, isl::multi_val right) {
+  auto resultList = isl::aff_list(left.get_ctx(), left.size());
+  for (int i = 0, n = left.size(); i < n; ++i) {
+    resultList = resultList.add(left.get_aff(i) / right.get_val(i));
+  }
+  return isl::multi_aff(left.get_space(), resultList);
```
skimo-openhub: Use `left.scale_down(right)`.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -135,13 +135,12 @@ std::unique_ptr<TensorReferenceGroup> TensorReferenceGroup::makeSingleton(
   return group;
 }
 
-isl::set ScopedFootprint::footprint(isl::set domain) const {
-  auto space = add_range(domain.get_space(), size());
-  auto accessed = isl::map::universe(space).intersect_domain(domain);
+isl::set TensorReferenceGroup::approximateFootprint() const {
+  auto accessed = scopedAccesses();
```
skimo-openhub: I don't understand this code.
You apparently start out from the actual accesses and then you "intersect"
that with an overapproximation of those accesses.
Doesn't that just give you the actual accesses again?

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -96,6 +96,11 @@ std::pair<isl::val, isl::aff> outputRangeSingle(isl::map access) {
 // TODO: replace this with a call to isl_map_range_box_hull when available.
 ScopedFootprint outputRanges(isl::map access) {
   int nSubscripts = access.dim(isl::dim_type::out);
+
+  if (access.get_space().domain().is_wrapping()) {
+    access = access.domain_factor_domain();
+  }
```
skimo-openhub: This is too much magic.
It should be the callers responsibility to ensure consistency.
See f183f2ed242b76960975944eba7576f09b326193 in #477.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -158,25 +188,61 @@ isl::set TensorReferenceGroup::approximateFootprint() const {
...
+
 isl::multi_aff ScopedFootprint::lowerBounds() const {
+  return createMA(
+      *this, [](const ScopedFootprintDim& dim) { return dim.lowerBound; });
+}
+
+isl::multi_aff ScopedFootprint::offsets() const {
+  return createMA(
+      *this, [](const ScopedFootprintDim& dim) { return dim.offset; });
+}
```
skimo-openhub: Why not store the offsets and lower bounds as `isl::multi_aff`s in the first place?

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -393,13 +459,21 @@ isl::multi_aff TensorReferenceGroup::promotion() const {
...
 
-  // lower bounds space is S -> P; which we transform into [S -> O] -> P
-  auto lowerBounds = approximation.lowerBounds().pullback(
-      isl::multi_aff::domain_map(accessSpace));
-  auto promotion = isl::multi_aff::range_map(accessSpace)
-                       .reset_tuple_id(isl::dim_type::out) -
-      lowerBounds;
-  return promotion;
+  // Construct a projection multi-aff in [S -> O] -> S
+  // for futher precomposition.
```
skimo-openhub: further

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -393,13 +459,21 @@ isl::multi_aff TensorReferenceGroup::promotion() const {
...
+  auto originalSpaceInserter = isl::multi_aff::domain_map(accessSpace);
+
+  // Lower bounds and offsets space is S -> P; transform into [S -> O] -> P.
+  auto lowerBounds =
+      approximation.lowerBounds().pullback(originalSpaceInserter);
+  auto offsets = approximation.offsets().pullback(originalSpaceInserter);
+
+  // Create identity mapping from original to promoted,
+  // resetting the tuple name for the promoted tensor.
```
skimo-openhub: Why does the tuple name need to be reset?
If you want to make as distinction between original and promoted tensor,
then it's better to assign an explicit name than to simply remove it.

tc/core/polyhedral/memory_promotion.h
```diff
@@ -35,15 +35,25 @@ enum class AccessType : short { Read, Write };
 // ScheduleTree*), the user is responsible for maintaining the correspondance
 // between schedule tree positions and footprints.
 // Overapproximates one dimension by its lower bound, affine function of
-// parameters and schedule dimensions visible around the scope, and by a
-// constant size.
+// parameters and schedule dimensions visible around the scope, by a
+// constant size, and by a pair offset/stride for strided accesses.  If the
+// access is not strided, then "offset" is a zero expression and "stride" is 1.
+// The lowerBound and the size are computed after removing the eventual stride.
```
skimo-openhub: The stride is computed first, so it's not "eventual".
Did you mean "potential"?

tc/core/polyhedral/memory_promotion.h
```diff
@@ -35,15 +35,25 @@ enum class AccessType : short { Read, Write };
...
+      isl::aff offset,
+      isl::val stride)
+      : lowerBound(lowerBound), size(size), offset(offset), stride(stride) {}
 
  public:
   isl::aff lowerBound;
   isl::val size;
+
+  isl::aff offset;
+  isl::val stride;
```
skimo-openhub: Why do you need the offsets and strides on each dimension individually?
Wouldn't it be better to store them together?

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -508,6 +508,110 @@ TEST_F(MatMulBias, RegisterPromotionSharedPreference) {
...
+
+      EXPECT_EQ(oneGroup->approximation[1].stride, isl::val(ctx, groupIStride));
+
+      auto expectedOffset =
+          isl::aff::zero_on_domain(ref->scopedAccess.domain().get_space()) +
+          groupIConstOffset;
+      // Convert to pw_aff because it has is_equal whereas a simple aff only has
+      // is_plain_equal that fails here.
+      EXPECT_TRUE(isl::pw_aff(oneGroup->approximation[1].offset)
+                      .is_equal(isl::pw_aff(expectedOffset)))
```
skimo-openhub: I don't think you should be checking for a specific offset.  The offset is implementation dependent.  You could check that the offsets are equal modulo the stride.

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -508,6 +508,110 @@ TEST_F(MatMulBias, RegisterPromotionSharedPreference) {
...
+  auto mscop = makeScopAndCheck(tc, {{"N", 42}, {"M", 420}}, 32, 2, -1);
+  ASSERT_TRUE(mscop.get() != nullptr);
+  auto& scop = mscop->scop();
+
+  // Additionally check that copies look fine.
+  scop.promoteEverythingAt({0, 0, 0});
+  auto code = std::get<0>(mscop->codegen("strided"));
+  EXPECT_TRUE(
+      code.find("I[32 * b1 + c2][64 * b0 + 2 * c3 + 1]") != std::string::npos)
+      << "expected only read elements to be copied";
```
skimo-openhub: Why does this test show that _only_ the elements read are copied?

tc/core/polyhedral/schedule_transforms.cc
ftynse: I will need it later for promotion.

tc/external/detail/islpp-inl.h
ftynse: it would be great if we were aware of that queue somehow

tc/core/polyhedral/memory_promotion.cc
ftynse: Because there is a `ScopedFootprintDim` class; changing/removing that class is extra work that does not bring us anywhere closer to the goal of this PR.

tc/core/polyhedral/memory_promotion.cc
ftynse: That names is not yet known at this point.

tc/core/polyhedral/memory_promotion.h
ftynse: Because we already have bounds and sizes for each dimension individually.  Storing some things individually and some other things collectively is confusing.  They are stored individually for cases when only _some_ dimensions can be approximated, which is useful when promoting indirect accesses.

skimo-openhub:
On Tue, Jun 05, 2018 at 06:06:01AM -0700, ftynse wrote:
> > @@ -773,5 +774,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
>    parent->insertChild(childPos, std::move(seq));
>  }
>  
> +/*
> + * Extract a mapping from the domain elements active at "tree"
> + * to identifiers "ids", where all branches in "tree"
> + * are assumed to have been mapped to these identifiers.
> + * The result lives in a space of the form "tupleId"["ids"...].
> + */
> +isl::multi_union_pw_aff extractDomainToIds(
> +    const detail::ScheduleTree* tree,
> +    const std::vector<mapping::MappingId>& ids,
> +    isl::id tupleId = isl::id()) {
> 
> I will need it later for promotion.

Then add it in a follow-up PR.

skimo

skimo-openhub:
On Tue, Jun 05, 2018 at 06:09:13AM -0700, ftynse wrote:
> >  isl::multi_aff ScopedFootprint::lowerBounds() const {
> +  return createMA(
> +      *this, [](const ScopedFootprintDim& dim) { return dim.lowerBound; });
> +}
> +
> +isl::multi_aff ScopedFootprint::offsets() const {
> +  return createMA(
> +      *this, [](const ScopedFootprintDim& dim) { return dim.offset; });
> +}
> 
> Because there is a `ScopedFootprintDim` class; changing/removing that class is extra work that does not bring us anywhere closer to the goal of this PR.

I don't see the connection.  Why does ScopedFootprintDim need to be removed
for you to be able to store the new information directly in ScopedFootprint?

skimo

skimo-openhub:
On Tue, Jun 05, 2018 at 01:11:54PM +0000, ftynse wrote:
> >  
>   public:
>    isl::aff lowerBound;
>    isl::val size;
> +
> +  isl::aff offset;
> +  isl::val stride;
> 
> Because we already have bounds and sizes for each dimension individually.  Storing some things individually and some other things collectively is confusing.  They are stored individually for cases when only _some_ dimensions can be approximated, which is useful when promoting indirect accesses.

Wouldn't you approximate them by the actual size then?
That is, why would some dimensions not be approximatable?
Also, what would you do with ScopedFootprint in such cases?

skimo

tc/core/polyhedral/memory_promotion.cc
ftynse: good point

skimo-openhub:
On Tue, Jun 05, 2018 at 06:06:30AM -0700, ftynse wrote:
> > @@ -38,10 +38,14 @@ inline isl::aff operator*(isl::aff A, isl::val V) {
>    return V * A;
>  }
>  
> +inline isl::aff operator/(isl::aff A, isl::val V) {
> +  isl::aff T(isl::local_space(A.get_space().domain()), V);
> 
> it would be great if we were aware of that queue somehow

It's tricky because some of those changee conflict with each other,
but I've separated them out for you (except those that depend
on isl changes).

skimo

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -158,25 +188,61 @@ isl::set TensorReferenceGroup::approximateFootprint() const {
...
   return accessed.range();
 }
 
+namespace {
+template <typename Func>
+isl::multi_aff createMA(
+    const std::vector<ScopedFootprintDim>& dims,
+    Func accessor) {
+  if (dims.size() == 0) {
+    throw promotion::PromotionNYI("promotion for scalars");
```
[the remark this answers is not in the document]
ftynse: it should be clear from it taking a vector of `ScopedFootprintDim` but ok

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -77,9 +77,10 @@ std::unique_ptr<TensorReferenceGroup> TensorReferenceGroup::makeSingleton(
   return group;
 }
 
-isl::set ScopedFootprint::footprint(isl::set domain) const {
```
skimo-openhub: Remove declaration too.

skimo-openhub:
> This also prepares for the introduction of shared memory promotion where

Do you mean _register_ promotion?

> accesses of each thread individually are strided with the stride equal
> to the number of threads.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -31,10 +31,50 @@ namespace polyhedral {
...
+    isl::map relation,
+    isl::multi_val strides,
+    isl::multi_aff offsets) {
+  CHECK_EQ(strides.size(), offsets.size());
+
+  auto space = relation.get_space();
+  auto stridesMA = isl::multi_aff::identity(space.range().map_from_set());
+  stridesMA = stridesMA / strides;
+
+  return relation.sum(isl::map(offsets.neg())).apply_range(isl::map(stridesMA));
```
skimo-openhub: Note: if you take the preimage over the scaling (rather than applying the anti-scaling)
then you don't need the second conversion.  (That's how it's done in PPCG.)

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -31,10 +31,50 @@ namespace polyhedral {
...
+}
+
+// Compute a box approximation of the range of the given relation,
+// including the lower bounds, the box sizes, and the strides.
+// If the range has strides, remove them first.
 ScopedFootprint outputRanges(isl::map access) {
+  auto ctx = access.get_ctx();
+  int nSubscripts = access.dim(isl::dim_type::out);
```
skimo-openhub: Hmm... I'm trying to get rid of those.
Anyway, should be easy enough to fix.  I'll send a follow-up commit.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -79,15 +119,21 @@ std::unique_ptr<TensorReferenceGroup> TensorReferenceGroup::makeSingleton(
 
 isl::set TensorReferenceGroup::approximateFootprint() const {
   auto scopedDomain = scopedAccesses().domain();
-  auto space = box.get_space();
+  auto space = approximation.box.get_space();
```
skimo-openhub: Huh?
This looks like it belongs to the previous commit....
...indeed, 5d48a371948b9bfb0c14859db93d2569fc327547 doesn't compile.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -79,15 +119,21 @@ std::unique_ptr<TensorReferenceGroup> TensorReferenceGroup::makeSingleton(
...
+    auto stride = approximation.stride(i);
+    auto strideOffset = approximation.strideOffset(i);
+    auto size = approximation.size(i);
     auto rhs = isl::aff(lspace, isl::dim_type::set, i);
-    isl::map partial = (isl::aff_map(dimLowerBound) <= rhs) &
-        (isl::aff_map(dimLowerBound + size(i)) > rhs);
+    auto lowerBound = offset * stride + strideOffset;
+    auto upperBound = (offset + size) * stride + strideOffset;
+    auto partial =
+        (isl::aff_map(lowerBound) <= rhs) & (isl::aff_map(upperBound) > rhs);
```
skimo-openhub: Hmm.... if you need to spell it out like that (with a loop), then this
probably means there is some functionality missing in isl, but we
can clean it up later.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -314,11 +360,24 @@ isl::multi_aff TensorReferenceGroup::promotion() const {
...
+  auto offsets = approximation.strideOffsets.pullback(originalSpaceInserter);
+
+  // Create promotion starting by identity in [S -> O] -> O.
+  auto original = isl::multi_aff::range_map(accessSpace);
+  auto promotion =
+      (original - offsets) / approximation.strideValues - lowerBounds;
+
+  // The target space is that of the promoted tensor, but the group name is not
+  // yet known.  Just reset the tuple id of the target space to obtain
+  // [S -> O] -> P where P has empty tuple id.
```
skimo-openhub: What's the harm in keeping the original name until the group name is known?

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -504,6 +504,113 @@ TEST_F(MatMulBias, RegisterPromotionSharedPreference) {
...
+              .is_equal(isl::pw_aff(expectedOffset.mod(stride))))
+          << oneGroup->approximation.strideOffset(1) << "\n"
+          << expectedOffset;
+    }
+    return mscop;
+  }
+};
+
+// Check that strides are effectively handled in memory promotion.  In
+// particular, check that array elements that are jumped over
```
skimo-openhub: Sentence doesn't look like it's finished.

tc/core/polyhedral/memory_promotion.cc
ftynse: I'd argue that the spaces are conceptually different (array in global memory vs array in some other memory), so they should not have matching tuples.  OTOH, using an unnamed tuple makes it look like it's the same space for all groups.

Previous implementation did not have the tuple name for the "promoted" space, hence the comment about lower bounds living in "S -> P", which is no longer the case.

skimo-openhub:
On Thu, Jun 07, 2018 at 01:31:33AM -0700, ftynse wrote:
> > +  // for further precomposition.
> +  auto originalSpaceInserter = isl::multi_aff::domain_map(accessSpace);
> +
> +  // Lower bounds and offsets space is S -> O; transform into [S -> O] -> O.
> +  auto lowerBounds =
> +      approximation.lowerBounds().pullback(originalSpaceInserter);
> +  auto offsets = approximation.strideOffsets.pullback(originalSpaceInserter);
> +
> +  // Create promotion starting by identity in [S -> O] -> O.
> +  auto original = isl::multi_aff::range_map(accessSpace);
> +  auto promotion =
> +      (original - offsets) / approximation.strideValues - lowerBounds;
> +
> +  // The target space is that of the promoted tensor, but the group name is not
> +  // yet known.  Just reset the tuple id of the target space to obtain
> +  // [S -> O] -> P where P has empty tuple id.
> 
> I'd argue that the spaces are conceptually different (array in global memory vs array in some other memory), so they should not have matching tuples.

Makes sense.

> OTOH, using an unnamed tuple makes it look like it's the same space for all groups.

Then how about using a temporary name that reflects that this
is a promoted version of (part of) a specific tensor?

skimo

ftynse:
> Then how about using a temporary name that reflects that this is a promoted version of (part of) a specific tensor? 

This probably deserves a refactoring at a higher level... The group ids are currently managed by a `Scop`.  Because promotion can be applied to different, potentially nested, subtrees, group ids must be unique in a `Scop`.  However, it looks reasonable that a group should _have_ an id rather than be associated with it in some map inside `Scop`.  This becomes too big for this PR.

skimo-openhub:
On Thu, Jun 07, 2018 at 01:56:48AM -0700, ftynse wrote:
> > Then how about using a temporary name that reflects that this is a promoted version of (part of) a specific tensor? 
> 
> This probably deserves a refactoring at a higher level... The group ids are currently managed by a `Scop`.  Because promotion can be applied to different, potentially nested, subtrees, group ids must be unique in a `Scop`.  However, it looks reasonable that a group should _have_ an id rather than be associated with it in some map inside `Scop`.  This becomes too big for this PR.

Can you leave the resetting out from this commit then?
It wasn't in there before this PR.

skimo

ftynse:
> Can you leave the resetting out from this commit then? It wasn't in there before this PR.

Yes, I did that and updated the documentation accordingly.

Issue #473, opened 2018-06-04
Do not store mapping options proto after tuning unless explicitly asked to


ftynse:
alternatively, do not automatically pick up those options.  It appears that TC currently stores and picks up options from `/tmp` silently.  This is a confusing behavior that makes it look like the autotuner works much better in some cases than it actually does.  It is very problematic if the behavior of the mapper changed and the previously stored options are used.

nicolasvasilache:
This is done implicitly in the `autotuner.tune` when passing a file, we should make it explicit.
Either have 2 tuning calls, but even better just expose a simple API for cache loading and storing that we can reuse and compose a load call with tune explicitly if we want to start from an existing state.

nicolasvasilache:
Atm we have implicit load and implicit store if a non-empty filename is passed (i.e. either "reinforce behavior" either "forget the world behavior").

However I turn this, it seems much more error prone to break the symmetry.
The only logical conclusion is to totally decouple tuning from storage in the C++ API, so tune will not take a filename anymore and whoever wants to reinforce from C++ will have to explicitly load from file, tune and then store to file.

Note that in the python API we most likely want to hide that complexity so passing a filename will be synonymous with "reinforce behavior".

Shout if you disagree on the claim that having an implicit assymmetric behavior will be quite more confusing.

ftynse:
I don't think asymmetry is a problem. This is essentially saving and
loading files. I doubt there is a potential user who is not familiar with
the concept of files.

Implicitness is a problem. Leaky abstratctions are a problem.  HCI 101:
users build a mental model of how your program works based on the state
they can observe (and almost nobody reads the docs past the first example).
When their mental model does not match how your program works, errors and
frustration appear. Explicit files are a much safer bet for the mental
model than saving something somewhere under the hood.

What happens if the user wants to overwrite the saved options with newly
generated ones, say after changing TC version? Well, they have to remember
that there is some file with long name in /tmp and delete it...  What
happens if somebody cleaned /tmp between runs, or worse, overwrote that
file with some other options?  If your goal is to make the use of TC tuner
unambiguous, this "implicit reinforce" is going in the opposite direction.

I agree that there should be a separate functionality for saving and
loading options. This is a clean separation of concerns. The tuner now
returns the list of options, and there is no segfault,  let the caller
decide what they want to do with it.

I disagree that there should be a different python API. We should take the
opportunity given by current refactoring and clean that up, make the API as
[...]

PR #476, opened 2018-06-04
Improve autotuner UX (both C++ and python land)

nicolasvasilache:
This PR serves the purpose of running the autotuner more smoothly on a script sent by a user:

```
wget -P /tmphttps://gist.githubusercontent.com/nicolasvasilache/d069e701229cb07f4b03a4c7980ab737/raw/1eef01140cae9a1bac1e3d771bb9008827e752be/batch_renormalization.py
python setup.py install --prefix=/tmp
python /tmp/batch_renormalization.py
```

I have been hitting Ctrl + C when perf looked "reasonable", user experience feels pretty good to me and we seem able to recover from existing protos.

```
Copying tensor_comprehensions.egg-info to /tmp/lib/python3.6/site-packages/tensor_comprehensions-0.1.1-py3.6.egg-info
running install_scripts
[INFO]: Autotuning cache will be saved to: /tmp/calc_mean_std_2_256_32_32_6.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 44/661/1335
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_r_d_256_256_256_256_6.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 8/9/34
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_O_2_256_32_32_256_256_256_256_256_256.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 20/83/3784
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_running_mean_std_256_256_256_256_6.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 8/9/37
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_xHat_grad_256_2_256_32_32.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 11/31/65
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_mean_std_grad_2_256_32_32_256_256_256_2_256_32_32.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 69/281/1745
Iteration 1     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 65/183/316
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_xHat_2_256_32_32_256_256_256_256.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 12/57/70
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_weight_bias_grad_2_256_32_32_2_256_32_32.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 70/220/2303
[INFO]: Tuned kernel options found, using those options
[INFO]: Autotuning cache will be saved to: /tmp/calc_I_grad_2_256_32_32_256_256_256_2_256_32_32_256_256.options
Iteration 0     Jobs(Compiled, Evaluated)/total  (100, 100)/100   (best/median/worst)us: 16/60/276
[INFO]: Tuned kernel options found, using those options
```

In the process, I had to debug a parameters issue in the second kernel. 
To print things properly it was also necessary to add operator<< to CpuMappingOptions.

### DIFF
--- a/tc/autotuner/autotuner-inl.h
+++ b/tc/autotuner/autotuner-inl.h
@@ -89,24 +89,28 @@ void TuningHarness<Backend>::doCompile(SearchStrategy& searchStrategy) {
     }
     std::unique_ptr<typename Backend::ExecutorType> pExecutor(nullptr);
     auto pConf = searchStrategy.population.at(current).get();
-    auto options = makeOptions<Backend>(baseMapping_, *pConf);
-    try {
-      if (FLAGS_debug_tuner) {
-        std::stringstream ssInfo;
-        typename Backend::MappingOptionsCppPrinter infoPrinter(ssInfo);
-        infoPrinter << options;
-        LOG(INFO) << "[COMPILE] Start compilation @:" << current;
-        LOG_LINE_BY_LINE(INFO, ssInfo);
+    if (not stopRequested_) {
+      auto options = makeOptions<Backend>(baseMapping_, *pConf);
+      try {
+        if (FLAGS_debug_tuner) {
+          std::stringstream ssInfo;
+          typename Backend::MappingOptionsCppPrinter infoPrinter(ssInfo);
+          infoPrinter << options;
+          LOG(INFO) << "[COMPILE] Start compilation @:" << current;
+          LOG_LINE_BY_LINE(INFO, ssInfo);
+        }
+        pExecutor =
+            tc::compile<Backend>(tcTree_, inputs_.begin()->second, options);
+        LOG_IF(INFO, FLAGS_debug_tuner) << "[COMPILE] Done compilation";
+      } catch (const std::exception& e) {
+        LOG(WARNING) << "[TUNER][COMPILE] failed compilation: " << e.what();
+        std::stringstream ssWarning;
+        typename Backend::MappingOptionsCppPrinter warningPrinter(ssWarning);
+        warningPrinter << options;
+        LOG_LINE_BY_LINE(WARNING, ssWarning);
+        pConf->invalid = true;
       }
-      pExecutor =
-          tc::compile<Backend>(tcTree_, inputs_.begin()->second, options);
-      LOG_IF(INFO, FLAGS_debug_tuner) << "[COMPILE] Done compilation";
-    } catch (const std::exception& e) {
-      LOG(WARNING) << "[TUNER][COMPILE] failed compilation: " << e.what();
-      std::stringstream ssWarning;
-      typename Backend::MappingOptionsCppPrinter warningPrinter(ssWarning);
-      warningPrinter << options;
-      LOG_LINE_BY_LINE(WARNING, ssWarning);
+    } else {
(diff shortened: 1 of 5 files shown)


nicolasvasilache:
cc @SSS135 who posted about this [here](https://gist.github.com/SSS135/d218f81dad12a0e5bab7665c1b5777ec)

ftynse:
The changes to `operator<<` don't seem to belong into "fix Ctrl+C", but ok.

ftynse:
> we seem able to recover from existing protos.

I find this behavior confusing as explained in #473.  In the worst case, dump those protos into the working directory, not `/tmp` to at least make them visible to the user.  If the tuning does not terminate the process on SIGINT, it can just as well return the best option found so far to make it usable from the code without some under-the-carpet communication through some temporary files.

nicolasvasilache:
> I also wouldn't call it useless. It does terminate the tuning, but indeed takes time. 

The current flow is to send SIGINT, wait for some (lenghty) amount fo time, then crash because of the double join rendering the termination utterly useless. Arguable we could separate the double join from the rest but I'd call that splitting hairs. This PR makes it actually usable 

> An alternative solution I have been proposing was to run the entire tuning as a separate process. 

Sure, we've been kicking this around for months. If you have the cycles go for it, this is not the intent of this PR. When we implement your suggestion, let's be sure the API is simple enough that we can implement it as an external service too if needed.

nicolasvasilache:
> The changes to operator<< don't seem to belong into "fix Ctrl+C", but ok.

I don't think we have strict rules wrt what can be included in a PR. This came up as part of investigating batch renormalization, the commits are well separated so that should satisfy our guidelines.

nicolasvasilache:
> I find this behavior confusing

This is happening when implicitly loading a proto during tuning if a filename is passed. I'll send a followup PR that can either `tune` (from scratch) or `tuneFromCheckpoint` (or something similar). Not in this PR since this is the current intended behavior.

ftynse:
> I don't think we have strict rules wrt what can be included in a PR. This came up as part of investigating batch renormalization, the commits are well separated so that should satisfy our guidelines.

Then let's turn this into policy.  I had to change my PRs repeatedly because of it.

PR #477, opened 2018-06-05
outputRanges: use isl::map::get_range_simple_fixed_box_hull

skimo-openhub:
The computation is fairly low-level.  Let isl take care of it.
TensorReferenceGroup::promotion no longer needs to fiddle
with the output space because the space is preserved
by isl::map::get_range_simple_fixed_box_hull.

### DIFF
--- a/tc/core/polyhedral/memory_promotion.cc
+++ b/tc/core/polyhedral/memory_promotion.cc
@@ -108,8 +108,7 @@ bool TensorReferenceGroup::isReadOnly() const {
 }
 
 isl::set TensorReferenceGroup::promotedFootprint() const {
-  auto space =
-      scopedAccesses().get_space().range().reset_tuple_id(isl::dim_type::set);
+  auto space = scopedAccesses().get_space().range();
   auto sizes = approximationSizes();
   if (sizes.size() != space.dim(isl::dim_type::set)) {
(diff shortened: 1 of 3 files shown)


ftynse:
Let me see how this combines with #472 first, there are other space-related modifications.

skimo-openhub:
On Tue, Jun 05, 2018 at 01:33:18AM -0700, ftynse wrote:
> Let me see how this combines with #472 first, there are other space-related modifications.

Well... one of the main goals of these changes was to make it easier
for you to do #472, but you apparently decided not to use them...

skimo

ftynse:
Well, I prefer to do things rather than to wait until something else gets done before I can even start thinking about what I want.  We already work 3 times slower than we used to.  Otherwise I can just stop coding.

skimo-openhub:
On Tue, Jun 05, 2018 at 02:17:40AM -0700, ftynse wrote:
> Well, I prefer to do things rather than to wait until something else gets done before I can even start thinking about what I want.  We already work 3 times slower than we used to.  Otherwise I can just stop coding.

AFAIK, you worked on your changes last week, while I made the extent
changes available to you (in test/extent) at the end of the week before.
All that changed since then is the commit id of the isl submodule.

skimo

PR #481, opened 2018-06-05
isAlmostIdentityReduction: drop FIXME

skimo-openhub:
This was in my queue, but I'll probably replace it with a rewrite of the reduction detection.

### DIFF
--- a/tc/core/polyhedral/reduction_matcher.cc
+++ b/tc/core/polyhedral/reduction_matcher.cc
@@ -90,11 +90,8 @@ bool pwAffInvolvesOnlyInputDim(isl::pw_aff pa, int redDimIdx) {
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


tc/core/polyhedral/reduction_matcher.cc
```diff
@@ -90,11 +90,8 @@ bool pwAffInvolvesOnlyInputDim(isl::pw_aff pa, int redDimIdx) {
   return true;
 }
 
-// Does pa have the form S(...) -> [(K*r)] where S is either a reduction init
+// Does pa have the form S(...) -> [f(r)] where S is either a reduction init
```
ftynse: actually, a "reduction init" does not exist anymore so this function is not detecting it

skimo-openhub:
Superseded by #520.

PR #489, opened 2018-06-07
register promotion: promote below any node in the tree

### LINKED ISSUE #490: Register promotion at arbitrary subtrees
Currently, register promotion is performed below the band mapped to threads.  We want to be able to perform below any node where it wouldn't break the tree structure (e.g., cannot promote below a sequence node because its children must be filters).

### LINKED ISSUE #505: extractDomainToThread: take into account filters

ftynse:
Introduce promoteToRegistersBelow that takes a schedule tree node as a
scoping point and attempts register promotion below that node.  Rewrite
promoteToRegistersBelowThreads to keep the existing behavior.

Independently of the promotion scope, tensor reference groups must be
computed per-thread (and per-block).  Include the thread (and block)
mapping in the domain of the outer schedule when computing the groups.

Since registers are private to a thread, different threads should not 
access the same value within the given scope.  Inside the promotion
scope, there may be multiple subtrees with different thread mappings.
Furthermore, these mappings are not included in the prefix (scope)
schedule.  Always include thread schedule when checking if accesses are 
performed from one thread only.  This schedule takes into account
potentially different mappings.

Relax the restriction for a promoted group to contain only one element.
Instead, check that all only the dimensions that either correspond to
unrolled loops or are mapped to blocks or threads appear in the access
subscripts.  During code generation, the corresponding loop iterators
are either replaced by constants due to unrolling or become implicit in
block/thread-specific groups (i.e., each thread in each block accesses
different elements so it can use the same "address" for them).

Compute reuse within threads by appending thread schedule to the prefix
schedule before checking for reuse.

Closes #505
Closes #490

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -509,7 +509,7 @@ isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
     ids.emplace_back(mapping::ThreadId::makeId(i));
   }
   auto tupleId = isl::id(tree->ctx_, kBlock);
-  return extractDomainToIds(tree, ids, tupleId);
+  return extractDomainToIds(scop_->scheduleRoot(), tree, ids, tupleId);
 }
 
 isl::multi_union_pw_aff MappedScop::blockMappingSchedule(
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -467,24 +467,6 @@ constexpr auto kBlock = "block";
 constexpr auto kWarp = "warp";
 
 /*
- * Extract a mapping from the domain elements active at "tree"
- * to the thread identifiers, where all branches in "tree"
- * are assumed to have been mapped to thread identifiers.
- * "nThread" is the number of thread identifiers.
- * The result lives in a space of the form block[x, ...].
- */
-isl::multi_union_pw_aff extractDomainToThread(
-    const detail::ScheduleTree* tree,
-    size_t nThread) {
-  std::vector<mapping::MappingId> ids;
-  for (size_t i = 0; i < nThread; ++i) {
-    ids.emplace_back(mapping::ThreadId::makeId(i));
-  }
-  auto tupleId = isl::id(tree->ctx_, kBlock);
-  return extractDomainToIds(tree, ids, tupleId);
-}
-
-/*
  * Construct a mapping
  *
  *  block[x] -> warp[floor((x)/warpSize)]
@@ -517,6 +499,16 @@ isl::multi_aff constructThreadToWarp(
   return isl::multi_aff(mapSpace, isl::aff_list(aff));
 }
 } // namespace
+
+isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
+    const detail::ScheduleTree* tree) const {
+  std::vector<mapping::MappingId> ids;
+  for (size_t i = 0; i < numThreads.view.size(); ++i) {
+    ids.emplace_back(mapping::ThreadId::makeId(i));
+  }
+  auto tupleId = isl::id(tree->ctx_, kBlock);
+  return extractDomainToIds(tree, ids, tupleId);
+}
 
 Scop::SyncLevel MappedScop::findBestSync(
     detail::ScheduleTree* st1,
@@ -708,7 +700,7 @@ void MappedScop::insertBestSyncInSeq(detail::ScheduleTree* seq) {
 
   auto outer = hasOuterSequentialMember(scop_->scheduleRoot(), seq);
 
-  auto domainToThread = extractDomainToThread(seq, numThreads.view.size());
+  auto domainToThread = threadMappingSchedule(seq);
   auto threadToWarp = constructThreadToWarp(seq->ctx_, 32, numThreads);
   auto domainToWarp = domainToThread.apply(threadToWarp);
 
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -476,28 +476,12 @@ constexpr auto kWarp = "warp";
 isl::multi_union_pw_aff extractDomainToThread(
     const detail::ScheduleTree* tree,
     size_t nThread) {
-  using namespace polyhedral::detail;
-
-  auto space = isl::space(tree->ctx_, 0);
-  auto empty = isl::union_set::empty(space);
-  auto id = isl::id(tree->ctx_, kBlock);
-  space = space.named_set_from_params_id(id, nThread);
-  auto zero = isl::multi_val::zero(space);
-  auto domainToThread = isl::multi_union_pw_aff(empty, zero);
-
-  for (auto mapping : tree->collect(tree, ScheduleTreeType::MappingFilter)) {
-    auto mappingNode = mapping->elemAs<ScheduleTreeElemMappingFilter>();
-    auto list = isl::union_pw_aff_list(tree->ctx_, nThread);
-    for (size_t i = 0; i < nThread; ++i) {
-      auto threadId = mapping::ThreadId::makeId(i);
-      auto threadMap = mappingNode->mapping.at(threadId);
-      list = list.add(threadMap);
-    }
-    auto nodeToThread = isl::multi_union_pw_aff(space, list);
-    domainToThread = domainToThread.union_add(nodeToThread);
+  std::vector<mapping::MappingId> ids;
+  for (size_t i = 0; i < nThread; ++i) {
+    ids.emplace_back(mapping::ThreadId::makeId(i));
   }
-
-  return domainToThread;
+  auto tupleId = isl::id(tree->ctx_, kBlock);
+  return extractDomainToIds(tree, ids, tupleId);
 }
 
 /*
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -1068,7 +1068,7 @@ std::unique_ptr<MappedScop> MappedScop::makeWithOuterBlockInnerThreadStrategy(
 
   // 9. Promote to registers below the loops mapped to threads.
   if (cudaOptions.proto().use_private_memory()) {
-    promoteToRegistersBelowThreads(mappedScop->scop(), -1ull);
+    promoteToRegistersBelowThreads(*mappedScop, -1ull);
   }
 
   LOG_IF(INFO, FLAGS_debug_tc_mapper)
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -461,6 +461,8 @@ bool hasOuterSequentialMember(
   return false;
 }
 
+// Name of the space of blocks inside the grid
+constexpr auto kGrid = "grid";
 // Name of the space of threads inside a block
 constexpr auto kBlock = "block";
 // Name of the space of warps
@@ -508,6 +510,16 @@ isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
   }
   auto tupleId = isl::id(tree->ctx_, kBlock);
   return extractDomainToIds(tree, ids, tupleId);
+}
+
+isl::multi_union_pw_aff MappedScop::blockMappingSchedule(
+    const detail::ScheduleTree* tree) const {
+  std::vector<mapping::MappingId> ids;
+  for (size_t i = 0; i < numBlocks.view.size(); ++i) {
+    ids.emplace_back(mapping::BlockId::makeId(i));
+  }
+  auto tupleId = isl::id(tree->ctx_, kGrid);
+  return extractDomainToIds(scop_->scheduleRoot(), tree, ids, tupleId);
 }
 
 Scop::SyncLevel MappedScop::findBestSync(
--- a/tc/core/polyhedral/cuda/mapped_scop.h
+++ b/tc/core/polyhedral/cuda/mapped_scop.h
@@ -193,6 +193,13 @@ class MappedScop {
   isl::multi_union_pw_aff threadMappingSchedule(
       const detail::ScheduleTree* tree) const;
 
+  // Extract a mapping from the domain elements active at "tree"
+  // to the block identifiers, where all branches in "tree"
+  // are assumed to have been mapped to block identifiers.
+  // The result lives in a space of the form grid[x, ...].
+  isl::multi_union_pw_aff blockMappingSchedule(
+      const detail::ScheduleTree* tree) const;
+
  private:
   // Insert the optimal combination of synchronizations in the sequence
   void insertBestSyncInSeq(detail::ScheduleTree* seq);
--- a/tc/core/polyhedral/cuda/mapped_scop.h
+++ b/tc/core/polyhedral/cuda/mapped_scop.h
@@ -186,6 +186,13 @@ class MappedScop {
       size_t nChildren,
       bool hasOuterSequentialMember);
 
+  // Extract a mapping from the domain elements active at "tree"
+  // to the thread identifiers, where all branches in "tree"
+  // are assumed to have been mapped to thread identifiers.
+  // The result lives in a space of the form block[x, ...].
+  isl::multi_union_pw_aff threadMappingSchedule(
+      const detail::ScheduleTree* tree) const;
+
  private:
   // Insert the optimal combination of synchronizations in the sequence
   void insertBestSyncInSeq(detail::ScheduleTree* seq);
--- a/tc/core/polyhedral/cuda/mapped_scop.h
+++ b/tc/core/polyhedral/cuda/mapped_scop.h
@@ -190,6 +190,8 @@ class MappedScop {
   // to the thread identifiers, where all branches in "tree"
   // are assumed to have been mapped to thread identifiers.
   // The result lives in a space of the form block[x, ...].
+  //
+  // Note: this function ignores statements introduced by extension nodes.
   isl::multi_union_pw_aff threadMappingSchedule(
       const detail::ScheduleTree* tree) const;
 
@@ -197,6 +199,8 @@ class MappedScop {
   // to the block identifiers, where all branches in "tree"
   // are assumed to have been mapped to block identifiers.
   // The result lives in a space of the form grid[x, ...].
+  //
+  // Note: this function ignores statements introduced by extension nodes.
   isl::multi_union_pw_aff blockMappingSchedule(
       const detail::ScheduleTree* tree) const;
 
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -48,14 +48,16 @@ bool isMappingIdType(const mapping::MappingId& id) {
 }
 
 /*
- * Is "tree" a mapping filter that maps a thread identifier?
+ * Is "tree" a mapping filter that maps identifiers of the type provided as
+ * template argument?
  */
-bool isThreadMapping(const detail::ScheduleTree* tree) {
+template <typename MappingType>
+bool isMappingTo(const detail::ScheduleTree* tree) {
   using namespace detail;
 
   if (auto filterNode = tree->elemAs<ScheduleTreeElemMappingFilter>()) {
     for (auto& kvp : filterNode->mapping) {
-      if (isMappingIdType<mapping::ThreadId>(kvp.first)) {
+      if (isMappingIdType<MappingType>(kvp.first)) {
         return true;
       }
     }
@@ -100,7 +102,7 @@ void mapCopiesToThreads(MappedScop& mscop, bool unroll) {
     // Check that we are not mapping to threads below other thread mappings.
     std::unordered_set<mapping::ThreadId, mapping::ThreadId::Hash> usedThreads;
     for (auto n : node->ancestors(root)) {
-      if (isThreadMapping(n)) {
+      if (isMappingTo<mapping::ThreadId>(n)) {
         throw promotion::PromotionBelowThreadsException(
             "attempted to map memory copies to threads below "
             "another thread mapping");
@@ -257,7 +259,7 @@ const detail::ScheduleTree* findThreadMappingAncestor(
     const detail::ScheduleTree* root,
     const detail::ScheduleTree* node) {
   auto ancestors = node->ancestors(root);
-  ancestors = functional::Filter(isThreadMapping, ancestors);
+  ancestors = functional::Filter(isMappingTo<mapping::ThreadId>, ancestors);
   if (ancestors.size() < 1) {
     throw promotion::PromotionLogicError("missing MappingFilter");
   }
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -649,27 +649,55 @@ void promoteGreedilyAtDepth(
   mapCopiesToThreads(mscop, unrollCopies);
 }
 
-namespace {
-
 /*
- * Perform promotion to registers below the thread specific marker "marker"
- * in the schedule tree of "mscop".
+ * Perform promotion to registers below the node "scope" in the schedule tree
+ * of "mscop".  Throw if promotion would violate the well-formedness of the
+ * schedule tree, in particular in cases of promotion immediately below
+ * a set/sequence node or immediately above a thread-specific marker node.
  */
-void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* marker) {
+void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* scope) {
+  // Cannot promote below a sequence or a set node.  Promotion may insert an
+  // extension node, but sequence/set must be followed by filters.
+  if (scope->elemAs<detail::ScheduleTreeElemSequence>() ||
+      scope->elemAs<detail::ScheduleTreeElemSet>()) {
+    throw promotion::IncorrectScope("cannot promote under a sequence/set node");
+  }
+  // Cannot promote between a thread-mapped band and a thread-specific marker
+  // node because the latter is used to identify thread-mapped bands as
+  // immediate ancestors.
+  if (scope->numChildren() == 1 &&
+      scope->child({0})
+          ->elemAs<detail::ScheduleTreeElemThreadSpecificMarker>()) {
+    throw promotion::IncorrectScope(
+        "cannot promote above a thread-specific marker node");
+  }
+
   auto& scop = mscop.scop();
   auto root = scop.scheduleRoot();
-  auto threadMapping = mscop.threadMappingSchedule(root);
 
-  auto partialSched = prefixSchedule(root, marker);
-  // Pure affine schedule without (mapping) filters.
-  auto partialSchedMupa = partialScheduleMupa(root, marker);
-
-  // Because this function is called below the thread mapping marker,
-  // partialSched has been intersected with both the block and the thread
-  // mapping filters.   Therefore, groups will be computed relative to
-  // blocks and threads.
+  // Compute groups specific to threads and block by including the mappings
+  // into the domain of the partials schedule.
+  auto blockMapping = collectMappingsTo<mapping::BlockId>(scop);
+  auto mapping =
+      collectMappingsTo<mapping::ThreadId>(scop).intersect(blockMapping);
+  auto schedule = partialSchedule(scop.scheduleRoot(), scope);
   auto groupMap = TensorReferenceGroup::accessedWithin(
-      partialSched, scop.reads, scop.writes);
+      schedule.intersect_domain(mapping), scop.reads, scop.writes);
+
+  auto threadSchedule = mscop.threadMappingSchedule(mscop.schedule());
+  auto blockSchedule = mscop.blockMappingSchedule(mscop.schedule());
+
+  // Pure affine schedule without (mapping) filters.
+  auto partialSchedMupa = partialScheduleMupa(root, scope);
+  // Schedule with block mapping filter.
+  auto partialSched =
+      isl::union_map::from(partialSchedMupa).intersect_domain(blockMapping);
+  // The following promotion validity and profitability checks need to be
+  // performed with respect to the block mapping, so append the block schedule.
+  // If the partial schedule contains it already, it will just end up with
+  // identical dimensions without affecting the result of the checks.
+  partialSchedMupa = partialSchedMupa.flat_range_product(blockSchedule);
+
   for (auto& tensorGroups : groupMap) {
     auto tensorId = tensorGroups.first;
 
@@ -682,12 +710,15 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* marker) {
         continue;
       }
       if (!isPromotableToRegistersBelow(
-              *group, root, marker, partialSchedMupa, threadMapping)) {
+              *group, root, scope, partialSchedMupa, threadSchedule)) {
         continue;
       }
-      if (!hasReuseWithin(*group, partialSchedMupa)) {
+      // Check reuse within threads.
+      auto schedule = partialSchedMupa.flat_range_product(threadSchedule);
+      if (!hasReuseWithin(*group, schedule)) {
         continue;
       }
+
       // TODO: if something is already in shared, but reuse it within one
       // thread only, there is no point in keeping it in shared _if_ it
       // gets promoted into a register.
@@ -695,13 +726,11 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* marker) {
           Scop::PromotedDecl::Kind::Register,
           tensorId,
           std::move(group),
-          marker,
+          scope,
           partialSched);
     }
   }
 }
-
-} // namespace
 
 // Promote at the positions of the thread specific markers.
 void promoteToRegistersBelowThreads(MappedScop& mscop, size_t nRegisters) {
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -34,9 +34,17 @@ namespace tc {
 namespace polyhedral {
 namespace {
 
-bool isThreadId(const mapping::MappingId& id) {
-  return id == mapping::ThreadId::x() or id == mapping::ThreadId::y() or
-      id == mapping::ThreadId::z();
+/*
+ * Is "id" a mapping of the type provided as template argument?
+ */
+template <typename MappingType>
+bool isMappingIdType(const mapping::MappingId& id) {
+  for (size_t i = 0; i < MappingType::kMaxDim; ++i) {
+    if (id == MappingType::makeId(i)) {
+      return true;
+    }
+  }
+  return false;
 }
 
 /*
@@ -47,7 +55,7 @@ bool isThreadMapping(const detail::ScheduleTree* tree) {
 
   if (auto filterNode = tree->elemAs<ScheduleTreeElemMappingFilter>()) {
     for (auto& kvp : filterNode->mapping) {
-      if (isThreadId(kvp.first)) {
+      if (isMappingIdType<mapping::ThreadId>(kvp.first)) {
         return true;
       }
     }
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -468,11 +468,13 @@ void promoteToSharedGreedy(
   // both.
   size_t remainingMemory = maxMemory;
   for (auto bandNode : bands) {
-    auto groupMap = TensorReferenceGroup::accessedBySubtree(bandNode, scop);
+    auto activePoints = activeDomainPoints(root, bandNode);
     auto partialSched = partialSchedule(root, bandNode);
+
+    auto groupMap = TensorReferenceGroup::accessedWithin(
+        partialSched.intersect_domain(activePoints), scop.reads, scop.writes);
     // Pure affine schedule without (mapping) filters.
     auto partialSchedMupa = partialScheduleMupa(root, bandNode);
-    auto activePoints = activeDomainPoints(root, bandNode);
 
     // Prepare groups for sorting, to have specified order necessary for
     // reproducibility and tests.
@@ -581,7 +583,12 @@ void promoteToRegistersBelowThreads(Scop& scop, size_t nRegisters) {
       auto mapSchedMupa = infixScheduleMupa(root, mapping, marker);
       auto partialSchedMupa = prefixSchedMupa.flat_range_product(mapSchedMupa);
 
-      auto groupMap = TensorReferenceGroup::accessedBySubtree(marker, scop);
+      // Because this function is called below the thread mapping marker,
+      // partialSched has been intersected with both the block and the thread
+      // mapping filters.   Therefore, groups will be computed relative to
+      // blocks and threads.
+      auto groupMap = TensorReferenceGroup::accessedWithin(
+          partialSched, scop.reads, scop.writes);
       for (auto& tensorGroups : groupMap) {
         auto tensorId = tensorGroups.first;
 
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -734,8 +734,6 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* scope) {
 
 // Promote at the positions of the thread specific markers.
 void promoteToRegistersBelowThreads(MappedScop& mscop, size_t nRegisters) {
-  using namespace tc::polyhedral::detail;
-
   auto& scop = mscop.scop();
   auto root = scop.scheduleRoot();
   auto markers = findThreadSpecificMarkers(root);
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -649,58 +649,70 @@ void promoteGreedilyAtDepth(
   mapCopiesToThreads(mscop, unrollCopies);
 }
 
+namespace {
+
+/*
+ * Perform promotion to registers below the thread specific marker "marker"
+ * in the schedule tree of "mscop".
+ */
+void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* marker) {
+  auto& scop = mscop.scop();
+  auto root = scop.scheduleRoot();
+  auto threadMapping = mscop.threadMappingSchedule(root);
+
+  auto partialSched = prefixSchedule(root, marker);
+  // Pure affine schedule without (mapping) filters.
+  auto partialSchedMupa = partialScheduleMupa(root, marker);
+
+  // Because this function is called below the thread mapping marker,
+  // partialSched has been intersected with both the block and the thread
+  // mapping filters.   Therefore, groups will be computed relative to
+  // blocks and threads.
+  auto groupMap = TensorReferenceGroup::accessedWithin(
+      partialSched, scop.reads, scop.writes);
+  for (auto& tensorGroups : groupMap) {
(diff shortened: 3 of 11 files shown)


ftynse:

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -474,29 +474,13 @@ constexpr auto kWarp = "warp";
...
-    auto mappingNode = mapping->elemAs<ScheduleTreeElemMappingFilter>();
-    auto list = isl::union_pw_aff_list(tree->ctx_, nThread);
-    for (size_t i = 0; i < nThread; ++i) {
-      auto threadId = mapping::ThreadId::makeId(i);
-      auto threadMap = mappingNode->mapping.at(threadId);
-      list = list.add(threadMap);
-    }
-    auto nodeToThread = isl::multi_union_pw_aff(space, list);
-    domainToThread = domainToThread.union_add(nodeToThread);
+    int nThreads) {
```
skimo-openhub: Why did you change the type of "nThreads"?

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -797,5 +798,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
...
+/*
+ * Extract a mapping from the domain elements active at "tree"
+ * to identifiers "ids", where all branches in "tree"
+ * are assumed to have been mapped to these identifiers.
+ * The result lives in a space of the form "tupleId"["ids"...].
+ */
+isl::multi_union_pw_aff extractDomainToIds(
+    const detail::ScheduleTree* tree,
+    const std::vector<mapping::MappingId>& ids,
+    isl::id tupleId = isl::id()) {
```
skimo-openhub: Why is the `tuplId` optional?
I couldn't find any uses of `extractDomainToIds` in this PR where it is not specified.

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -517,6 +499,16 @@ isl::multi_aff constructThreadToWarp(
 }
 } // namespace
 
+isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
+    const detail::ScheduleTree* tree) const {
+  std::vector<mapping::MappingId> ids;
+  for (int i = 0; i < numThreads.view.size(); ++i) {
```
skimo-openhub: warning: comparison between signed and unsigned integer expressions

tc/core/polyhedral/cuda/mapped_scop.h
```diff
@@ -186,6 +186,14 @@ class MappedScop {
       size_t nChildren,
       bool hasOuterSequentialMember);
 
+  // Extract a mapping from the domain elements active at "tree"
+  // to the thread identifiers, where all branches in "tree"
+  // are assumed to have been mapped to thread identifiers.
+  // "nThread" is the number of thread identifiers.
```
skimo-openhub: no longer relevant

skimo-openhub:
31a70bb507065e4464553a937470a12ffa5e4691:
> A subtree may have mapping filter nodes with different ids.  If this

Really?  what kind of ids are you thinking of?

Also, you are modifying `extractDomainToId` and ensuring that the modification
does not have any effect in the single call to `extractDomainToId` in the entire PR.
This very much looks like the introduction of dead code.

skimo-openhub:
31a70bb507065e4464553a937470a12ffa5e4691:
> Original implementation assumed that the given node is band and used a

*The* original implementation
*a* band

> CHECK macro to enforce that.  In practice, it may be useful to call
> partialScheduleMupa to obtain the partial schedule including the
> schedule of the current node if it is a band, and the prefix schedule
> otherwise.  Replace the CHECK by an early return of the result of
> prefixScheduleMupa.

The commit itself looks good.
(I'm not too fond of the `partialScheduleMupa` name, but that's a separate issue.
It's just that the (for me) confusing name means that it is not obvious to determine
whether this change make sense or not.)

skimo-openhub:
On Fri, Jun 08, 2018 at 09:30:39AM +0000, ftynse wrote:
> On 08/06/2018 11:22, skimo-openhub wrote:
> >
> > 31a70bb 
> > <https://github.com/facebookresearch/TensorComprehensions/commit/31a70bb507065e4464553a937470a12ffa5e4691>:
> >
> >     A subtree may have mapping filter nodes with different ids. If this
> >
> > Really? what kind of ids are you thinking of?
> >
> I'm calling it on the root of the tree and asking for thread ids. There 
> will be a mapping to blocks that will fail the check.

Ah.. you mean nodes with different _types_ of ids.
OK, that part makes sense now.
I'll review that commit again.

skimo

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -506,7 +506,10 @@ isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
     ids.emplace_back(mapping::ThreadId::makeId(i));
   }
   auto tupleId = isl::id(tree->ctx_, kBlock);
-  return extractDomainToIds(tree, ids, tupleId);
+  auto mapping = extractDomainToIds(tree, ids, tupleId);
+  // Check that all active elements got mapped.
+  CHECK(activeDomainPoints(schedule(), tree).is_subset(mapping.domain()));
```
skimo-openhub: Shouldn't this check be put in `extractDomainToIds` intead?

I know you say

> The check was originally introduced in extractDomainToThreads and was
> meant to ensure that all branches in the subtree are indeed mapped to
> threads.  Perform the check inside threadMappingSchedule (successor of
> extractDomainToThreads) instead by checking that all active domain
> points are present in the mapping.

but I don't think that's the case.  The check you removed doesn't check that all branches
are mapped, only that every mapping node found maps to the requested ids.
So, the check you introduce here may be useful, but it's a separate issue and
looks to me like it belongs to `extractDomainToIds`.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -823,11 +823,18 @@ isl::multi_union_pw_aff extractDomainToIds(
...
+      if (mappingNode->mapping.count(id) == 0) {
+        hasIds = false;
+        break;
+      }
       auto idMap = mappingNode->mapping.at(id);
       list = list.add(idMap);
     }
+    if (!hasIds) {
+      continue;
+    }
```
skimo-openhub: Hmm... I wonder if it would be clearer if you would simply collect all the elements with a matching id and then check that the resulting list is either empty (in which case you skip it) or of size equal to `ids.size()`.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -526,6 +526,21 @@ void promoteToSharedGreedy(
     scop.insertSyncsAroundCopies(bandNode);
   }
 }
+
+// Returns a union of all mapping filters to "MappingType" in "scop".
```
skimo-openhub: Nit: I think we either use imperative ("Return") or an explicit subject ("This function returns")

The union is unambiguously defined, so it should be "the union".

It doesn't seems like you are actually taking any union, though...

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -526,6 +526,21 @@ void promoteToSharedGreedy(
...
+template <typename MappingType>
+isl::union_set collectMappingsTo(const Scop& scop) {
+  auto root = scop.scheduleRoot();
+  auto domain = scop.domain();
+  auto mappingFilters = detail::ScheduleTree::collect(
+      root, detail::ScheduleTreeType::MappingFilter);
+  mappingFilters = functional::Filter(isMappingTo<MappingType>, mappingFilters);
```
skimo-openhub: Ha!  Why don't you do this in `extractDomainToIds` instead of the explicit `hasIds` stuff?

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -526,6 +526,21 @@ void promoteToSharedGreedy(
...
+  for (auto mf : mappingFilters) {
+    auto filterNode = mf->elemAs<detail::ScheduleTreeElemMappingFilter>();
+    domain = domain.intersect(filterNode->filter_);
```
skimo-openhub: Perhaps I'm missing something, but this doesn't seem to make sense.
You're looking at multiple mapping filters of the same type and then intersecting them.
Since multiple (thread) mapping filters have disjoint domains, the result will be empty if there is any branching.
I'll request changes for this one.

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -328,17 +328,38 @@ void addSingletonReferenceGroups(
...
+// "outerSchedule" are considered.  In particular, if this domain is
+// intersected with block and/or thread mapping, the reference groups are
+// computed inside one block and/or thread, even if "outerSchedule" does not
+// include band members mapped to blocks and/or threads.
+//
+// Tensor reference descriptors (TensorReference) contain information about
+// tensor elements accessed through the given reference within "outerSchedule".
+// Several references form a group (TensorReferenceGroup) if the same elements
+// may be accessed through these references, and at least one of the accesses
+// write to the element.  A group stores a rectangular overapproximation of the
```
skimo-openhub: writes

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -328,17 +328,38 @@ void addSingletonReferenceGroups(
...
+// set of accessed tensor elements (access footprint).  This overappoximation
+// can be used to create copies of the given tensor elements in another memory
+// space, i.e., to perform memory promotion.  If the domain of "outerSchedule"
+// included thread or block mapping, then the overappoximation is computed
+// per-block or per-thread.
+//
+// Returns a map between tensor ids and vectors of unique pointers to
```
skimo-openhub: Return

tc/core/polyhedral/memory_promotion.h
```diff
@@ -111,9 +111,10 @@ class TensorReferenceGroup {
   TensorReferenceGroup() {}
 
  public:
-  static TensorGroups accessedBySubtree(
-      const detail::ScheduleTree* tree,
-      const Scop& scop);
+  static TensorGroups accessedWithin(
```
skimo-openhub: Isn't the TC convention to document the declaration rather than the implementation?
(I don't mind, personally; I find it more convenient to have the documentation close to the implementation.)

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -50,6 +50,21 @@ class TestMapper : public ::testing::Test {
     return MappedScop::makeWithOuterBlockInnerThreadStrategy(
         std::move(scop), mappingOptions);
   }
+
+  // This mimics behavior of the old TensorReferenceGroup::accessedBySubtree.
+  // In particular, it takes the partial schedule until "tree" inclusive,
+  // intersecting its domain with all (mapping) filter ancestors and compute
```
skimo-openhub: computes

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -323,11 +323,138 @@ bool promotionImprovesCoalescing(
...
+ * one must consider per-reference access relation for single-valuedness as
+ * different references may have different values, but all of them remain
+ * independent of non-unrolled loop iterators.
+ */
+bool accessSubscriptsAreUnrolledLoops(
+    const TensorReferenceGroup& group,
+    const detail::ScheduleTree* root,
+    const detail::ScheduleTree* scope,
+    isl::multi_union_pw_aff prefixSchedule,
+    isl::multi_union_pw_aff threadSchedule) {
```
skimo-openhub: You didn't document the meaning of `threadSchedule`, but it seems
to be band schedule from which the thread mapping is derived.
I think you need the actual thread mapping here, i.e., the output of
`MappedScop::threadMappingSchedule`.
Concatenating the prefix schedule, the thread mapping schedule and the unrolled schedule dimensions, should give you the right discriminator.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -323,11 +323,138 @@ bool promotionImprovesCoalescing(
...
+      }
+    }
+
+    auto space = isl::space(leaf->ctx_, 0, unrolledDims.n())
+                     .align_params(subdomain.get_space());
+    auto unrolledDimsMupa = isl::multi_union_pw_aff(space, unrolledDims);
+
+    // It is possible that no loops are unrolled, in which case
+    // unrolledDimsMupa is zero-dimensional and needs an explicit domain
+    // to be convertible to a union_map.
```
skimo-openhub: Did you actually run into this problem?
You're concatenating with the other schedules first, so I wouldn't expect this to be an issue.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -323,11 +323,138 @@ bool promotionImprovesCoalescing(
...
+      }
+      schedule = schedule.intersect_domain(subdomain);
+
+      for (size_t i = 0, e = band->nMember(); i < e; ++i) {
+        if (!band->unroll_[i]) {
+          continue;
+        }
+        unrolledDims = unrolledDims.add(schedule.get_union_pw_aff(i));
+      }
+    }
```
skimo-openhub: I would put the collection of the unrolled schedule into a separate function.
`accessSubscriptsAreUnrolledLoops` is pretty long now.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -559,9 +667,10 @@ void promoteGreedilyAtDepth(
 }
 
 // Promote at the positions of the thread specific markers.
-void promoteToRegistersBelowThreads(Scop& scop, size_t nRegisters) {
+void promoteToRegistersBelowThreads(MappedScop& mscop, size_t nRegisters) {
```
skimo-openhub: Why do you need `mscop` here?
You only seem to be using it in `mscop.scop()`.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -797,5 +798,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
   parent->insertChild(childPos, std::move(seq));
 }
 
+/*
+ * Extract a mapping from the domain elements active at "tree"
```
nicolasvasilache: My comments are from legacy, so my apologies for raising this late.

First, I find it confusing to not see anything related to `activeDomainPoints`.
The name `extractMappingExpressions` feels more appropriate to me since you are really collecting the expressions for each mapping, irrespective of the sets of points.

Second, what happens in an imperfectly nested or full / partial tile world where you could generally have multiple mapping expressions to different subpieces of a given statement?
Doesn't this warrant some sanity checks and a comment?

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -651,6 +615,52 @@ void promoteToSharedGreedy(
...
+inline bool isPromotableToRegistersBelow(
+    const TensorReferenceGroup& group,
+    const detail::ScheduleTree* root,
+    const detail::ScheduleTree* scope,
+    isl::multi_union_pw_aff prefixSchedule,
+    isl::multi_union_pw_aff threadSchedule) {
+  return isThreadPrivate(group, prefixSchedule, threadSchedule) &&
+      accessSubscriptsAreUnrolledLoops(
+             group, root, scope, prefixSchedule, threadSchedule);
+}
```
skimo-openhub: Did you just rename `isPromotableToRegisterBelowThreads` to `isPromotableToRegistersBelow` and then split it up into two functions?
Surely, that could have been done in a separate commit.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -666,60 +676,92 @@ void promoteGreedilyAtDepth(
...
-          // gets promoted into a register.
-          scop.promoteGroup(
-              Scop::PromotedDecl::Kind::Register,
-              tensorId,
-              std::move(group),
-              marker,
-              partialSched);
-        }
-      }
-    }
```
skimo-openhub: Can you do the outlining in a separate commit?
It's difficult to see exactly what you changed when you combine the move and the changes in a single commit.

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -461,9 +461,45 @@ def fun(float(N,K) A, float(K,M) B, float(N,M) C) -> (O) {
...
+    // We don't know the exact name of the iterator, but it starts with c.
+    auto oWithIteratorPos = code.find("_O_0[c");
+    auto oWithThreadPos = code.find("_O_0[t1");
+
+    EXPECT_TRUE(oWithIteratorPos == std::string::npos)
+        << "accessing local arrays with iterators in subscripts makes "
+        << "these arrays placed in local memory instead of registers";
+    EXPECT_TRUE(oWithThreadPos == std::string::npos)
+        << "expected per-thread groups to be computed, i.e. thread "
+        << "identifiers should not appear in the subscripts";
```
skimo-openhub: How about checking the AST in `emitCuda` to make sure that every access
to a tensor promoted to registers has integer index expressions instead?

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -505,16 +537,103 @@ TEST_F(MatMulBias, RegisterPromotionSharedPreference) {
...
+  // We don't know the exact name of the iterator, but it starts with c.
+  auto oWithIteratorPos = code.find("_O_0[c");
+  auto oWithThreadPos = code.find("_O_0[t1");
+
+  EXPECT_TRUE(oWithIteratorPos == std::string::npos)
+      << "accessing local arrays with iterators in subscripts makes "
+      << "these arrays placed in local memory instead of registers";
+  EXPECT_TRUE(oWithThreadPos == std::string::npos)
+      << "expected per-thread groups to be computed, i.e. thread "
+      << "identifiers should not appear in the subscripts";
```
skimo-openhub: Isn't this `expectNoSymbolicSubscript`?

skimo-openhub:
On Fri, Jun 08, 2018 at 08:44:58AM -0700, Nicolas Vasilache wrote:
> > @@ -797,5 +798,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
>    parent->insertChild(childPos, std::move(seq));
>  }
>  
> +/*
> + * Extract a mapping from the domain elements active at "tree"
> 
> My comments are from legacy, so my apologies for raising this late.
> 
> First, I find it confusing to not see anything related to `activeDomainPoints`.
> The name `extractMappingExpressions` feels more appropriate to me since you are really collecting the expressions for each mapping, irrespective of the sets of points.

Either name sounds reasonable to me.

> Second, what happens in an imperfectly nested or full / partial tile world where you could generally have multiple mapping expressions to different subpieces of a given statement?

The function will collect the mappings over the subpieces.

> Doesn't this warrant some sanity checks and a comment?

There is a comment:

 * where all branches in "tree"
 * are assumed to have been mapped to these identifiers.

and Alex added a sanity check in the caller.

skimo

tc/core/polyhedral/schedule_transforms.cc
nicolasvasilache: (Moving from main thread to scoped location in github UI)
> First, I find it confusing to not see anything related to `activeDomainPoints`.

Either name sounds reasonable to me.

> Second, what happens in an imperfectly nested or full / partial tile world where you could generally have multiple mapping expressions to different subpieces of a given statement?

The function will collect the mappings over the subpieces.

> Doesn't this warrant some sanity checks and a comment?
There is a comment:
 * where all branches in "tree"
 * are assumed to have been mapped to these identifiers.

and Alex added a sanity check in the caller.

skimo

nicolasvasilache: > The function will collect the mappings over the subpieces.

@skimo-openhub I'm sorry but I don't see it, here is what the function computes on `MatMulBias.RegistersAtRoot`:

```
E0610 11:21:34.416007 61796 schedule_transforms.cc:842] [K, M, N] -> block[{ S_0[O_s0_i, O_s0_j] -> [(O_s0_j - 32*floor((O_s0_j)/32))]; S_1[O_s1_i, O_s1_j, O_s1_k] -> [(O_s1_j - 32*floor((O_s1_j)/32))]; S_2[O_s2_i, O_s2_j] -> [(O_s2_j - 32*floor((O_s2_j)/32))] }, { S_0[O_s0_i, O_s0_j] -> [(O_s0_i - 8*floor((O_s0_i)/8))]; S_1[O_s1_i, O_s1_j, O_s1_k] -> [(O_s1_i - 8*floor((O_s1_i)/8))]; S_2[O_s2_i, O_s2_j] -> [(O_s2_i - 8*floor((O_s2_i)/8))] }]
```

Since no set/domain information is included (no active domain), there is an assumption that a statement may appear only in a single subtree below the current node (or that all mapping expressions for a statement must agree). Otherwise you would get multiple entries for the same statement and they would be impossible to disambiguate. The reusable functions in schedule_transforms should be generally applicable; here it seems they can only be applied under assumptions. So I think this needs a check for the above assumption and a comment. We may want to review other functions in this schedule_transforms.cc.

> Either name sounds reasonable to me

I've thought about this more I really think domain and active element should go (there is no domain or active elements in the computation); it's really just about recovering the expressions of the mappings.

Lastly it would be better to use the github web UI for review otherwise these comments get decoupled from their place of occurrence.

tc/core/polyhedral/cuda/mapped_scop.h
```diff
@@ -186,6 +186,14 @@ class MappedScop {
...
+  // The result lives in a space of the form block[x, ...].
+  isl::multi_union_pw_aff threadMappingSchedule(
```
nicolasvasilache: Same comment as in the previous commit, there is no schedule involved here so I would call this `threadMappingExpressions` or something along that vein.

tc/core/polyhedral/cuda/mapped_scop.cc
nicolasvasilache: + 1 for putting it inside `extractDomainToIds` and combine with the previously suggested checks).

tc/core/polyhedral/schedule_transforms.cc
nicolasvasilache: +1 was going to suggest the same

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -328,17 +328,38 @@ void addSingletonReferenceGroups(
 }
 } // namespace
 
-TensorGroups TensorReferenceGroup::accessedBySubtree(
-    const ScheduleTree* tree,
-    const Scop& scop) {
+// Compute tensor reference groups encapsulating all tensor accesses within
+// "outerSchedule".  Only statement instances present in the domain of
+// "outerSchedule" are considered.  In particular, if this domain is
+// intersected with block and/or thread mapping, the reference groups are
```
nicolasvasilache: It seems to me that the comment about groups that are "private" to a mapping really come as a byproduct of the IR. Generic parameters are understood to take one value in the parameter domain whereas mapping parameters are understood to take all values in the set but are represented as generic parameters. So the consequence for this function (and other functions that don't treat mapping paremeters in a special way is that you get the "private" behavior. Does that sound like a reasonable description? Worth to mention in the comment?

tc/core/polyhedral/memory_promotion.h
nicolasvasilache: Is `static` a sign that this function (and more generally the notion of TensorGroup) is ready for hoisting out of mapped_scop? (not in this PR)

tc/core/polyhedral/memory_promotion.cc
```diff
@@ -461,14 +461,20 @@ isl::multi_aff dropDummyTensorDimensions(
   space = add_range(space, list.n());
   return isl::multi_aff(space, list);
 }
+
+inline void unrollAllMembers(detail::ScheduleTreeElemBand* band) {
```
nicolasvasilache: Make a TODO that we want to investigate this in the context of long compilation times?

tc/core/polyhedral/schedule_transforms.cc
skimo-openhub: On Sun, Jun 10, 2018 at 10:38:37AM -0700, Nicolas Vasilache wrote:
> > @@ -797,5 +798,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::unio
n_set filter) {
>    parent->insertChild(childPos, std::move(seq));
>  }
>
> +/*
> + * Extract a mapping from the domain elements active at "tree"
>
> > The function will collect the mappings over the subpieces.
>
> @skimo-openhub I'm sorry but I don't see it, here is what the function computes on `MatM
ulBias.RegistersAtRoot`:
>
> ```
> E0610 11:21:34.416007 61796 schedule_transforms.cc:842] [K, M, N] -> block[{ S_0[O_s0_i,
 O_s0_j] -> [(O_s0_j - 32*floor((O_s0_j)/32))]; S_1[O_s1_i, O_s1_j, O_s1_k] -> [(O_s1_j -
32*floor((O_s1_j)/32))]; S_2[O_s2_i, O_s2_j] -> [(O_s2_j - 32*floor((O_s2_j)/32))] }, { S_
0[O_s0_i, O_s0_j] -> [(O_s0_i - 8*floor((O_s0_i)/8))]; S_1[O_s1_i, O_s1_j, O_s1_k] -> [(O_
s1_i - 8*floor((O_s1_i)/8))]; S_2[O_s2_i, O_s2_j] -> [(O_s2_i - 8*floor((O_s2_i)/8))] }]
> ```

This in itself doesn't show much, but it seems you are right that
the mapping does not have enough information in general 
(in this case it does).

> Since no set/domain information is included (no active domain), there is an assumption t
hat a statement may appear only in a single subtree below the current node (or that all ma
pping expressions for a statement must agree). Otherwise you would get multiple entries fo
r the same statement and they would be impossible to disambiguate. The reusable functions
in schedule_transforms should be generally applicable; here it seems they can only be appl
ied under assumptions. So I think this needs a check for the above assumption and a commen
t. We may want to review other functions in this schedule_transforms.cc.

I was assuming that the mapping would contain information about the filters
because that is what I'm used to in PPCG, but it seems we do indeed 
not carry enough information in general.
I also misunderstood the check you wanted to add.
AFAIU now, you want to check that there is no point that gets mapped
more than once (rather than checking that each point gets mapped).

> > Either name sounds reasonable to me
>
> I've thought about this more I really think domain and active element should go

I don't understand.  Do you mean that these things should be removed?
Where should they be removed from?

> (there is no domain or active elements in the computation); it's really just about recov
ering the expressions of the mappings.
>
> Lastly it would be better to use the github web UI for review otherwise these comments g
et decoupled from their place of occurrence.

Ugh.  The github UI is horrible.  I'll just prepare an email and then paste it.

tc/core/polyhedral/cuda/mapped_scop.h
skimo-openhub: > Same comment as in the previous commit, there is no schedule involved here so I would ca
ll this `threadMappingExpressions` or something along that vein.

Do you mean that this is more like a "placement" than a "schedule"?
Usually, we don't make this distinction at this level.
I don't think "Expressions" adds any value.
The main point it to get the mapping to threads, so if you don't
like the "Schedule" bit, I'd just call it "threadMapping".

tc/core/polyhedral/schedule_transforms.cc
skimo-openhub: I opened up #505; I have an implementation too, but it exposes some issue in isl
that I need to send out first.

tc/core/polyhedral/cuda/mapped_scop.h
nicolasvasilache: Ah I see the terminology impedance mismatch. What ticked me off here is that I have built this mental model that functions involving the word schedule usually collect the mupa upt to the root. Similarly for "active elements", "active domain" and sometimes "domain" which I grew accustomed to seeing call `activeDomainPoints`. Should I revisit these expectations? threadMapping seems fine to me too.

tc/core/polyhedral/schedule_transforms.cc
nicolasvasilache: > AFAIU now, you want to check that there is no point that gets mapped more than once (rather than checking that each point gets mapped).

My concern with the information currently produced was that even totally disjoint sets of points getting mapped with different mappings would create ambiguities. So with the current information it's really: enforce <= 1 mapping filter per statement in the subtree. But what you write makes sense to

tc/core/polyhedral/cuda/mapped_scop.cc
ftynse: That function could be used precisely to check what is _not_ mapped by subtracting its result from the active points.  But i'll change for this PR.

ftynse: This is slightly annoying because it requires "extractDomainToIds" to take the root pointer which it does need otherwise.

tc/core/polyhedral/cuda/mapped_scop.h
ftynse: We represent the mapping as a `union_set` filter _and_ as a `multi_union_pw_aff`.  I actually use both in a function in the further commits blocked by this PR.  I wanted to make it clear which type is used for the mapping in a particular location, so I went for the explicit `Schedule` suffix.

>  that functions involving the word schedule usually collect the mupa upt to the root

Not necessarily, for example `infixScheduleMupa`.

tc/core/polyhedral/memory_promotion.h
ftynse: It never was in mapped scop.  This is a static constructor of an object that wraps `unique_ptr`s.  We always do it this way.

tc/core/polyhedral/memory_promotion.cc
ftynse: > That means that at the point where you introduced
> parameters for block/thread identifiers, you are
> considering one particular block/thread.

This is exactly how I see it, and that's why that function can compute groups specific to blocks and/or threads.  It _is_ considering one particular block/thread, it just can express everything given the id of that block thread so it returns the result in a compact form.  It is the same as parametric domains and what @nicolasvasilache calls generic parameters.  Parameters will have one fixed value, but the schedule is computed so that it remains valid for all possible values of those parameters.

It took me some time to wrap my head around it in context of the code generation, but it is actually easier to reason this way.

tc/core/polyhedral/cuda/mapped_scop.cc
skimo-openhub: > This is slightly annoying because it requires "extractDomainToIds" to take the root poin
ter which it does need otherwise.

What was the motivation for extracting out extractDomainToIds
in the first place?  The commit message says it could be used
for other types of mapping ids, but this PR doesn't do that.
It's then difficult to reason about what extractDomainToIds
should look like.

skimo

ftynse:
> 829e842 doesn't appear to be needed in this PR either

now it is

tc/core/polyhedral/cuda/mapped_scop.cc
ftynse: We seem to compile with `-Wno-sign-compare`

tc/core/polyhedral/cuda/mapped_scop.cc
ftynse: to have the `blockMappingSchedule` in the same way, it's now added

ftynse:
> isMappingTo looks a bit strange with a detail::ScheduleTree argument, but I guess it's referring to the template argument.

Yes, the template argument is non-deducible, so the call will always look like `isMappingTo<ThreadId>(...)`

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: Because a mapping to "t0" only is still a mapping to threads, but want that function to throw in case the caller asked for the mapping to "t0" _and_ "t1".

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse:
```
(tc_build) zinenko@citrouille:~/TensorComprehensions$ git grep "Returns " | wc -l
25
(tc_build) zinenko@citrouille:~/TensorComprehensions$ git grep "Return " | wc -l                                          
43
```
With such balance, I don't this there is even an implicit convention.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -323,11 +323,138 @@ bool promotionImprovesCoalescing(
...
+  auto domain = scop.domain();
+  auto mappingFilters = detail::ScheduleTree::collect(
+      root, detail::ScheduleTreeType::MappingFilter);
+  mappingFilters = functional::Filter(isMappingTo<MappingType>, mappingFilters);
+  for (auto mf : mappingFilters) {
+    auto filterNode = mf->elemAs<detail::ScheduleTreeElemMappingFilter>();
+    domain = domain.intersect(filterNode->filter_);
+  }
+  return domain;
+}
```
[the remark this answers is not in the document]
ftynse: This will lead to annoying rebase conflict management.  The second-to-last commit is artificially introduced.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: Indeed, this is due to an artifically introduced commit.  The old function was just erased in my original commit rather than going through progressive upates with extra code that just gets deleted by the subsequent commit.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: I did, hence the explanatory comment, although it may have been at some previous version of the code.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: not anymore

tc/core/polyhedral/schedule_transforms.cc
ftynse: I could not fix it because of the problem in isl.  I'm fine waiting for it to be fixed, or merging this and addressing in a separate PR

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: No, I wrote a completely new `isPromotableToRegistersBelow` (including the unrolling part that I separated in a different commit) and then called it from `isPromotableToRegistersBelowThreads`.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: That would be very annoying because this would be another artificial commit with potential problems and potentially new code that will be erased by the subsequent commit.

test/test_cuda_mapper_memory_promotion.cc
ftynse: This is a good idea, but for a separate PR.

skimo-openhub:
I'm going to edit this during my review.

skimo-openhub:
On Wed, Jun 13, 2018 at 11:11:04AM -0700, ftynse wrote:
> > 829e842 doesn't appear to be needed in this PR either
> 
> now it is

How?  I only see you adding a commit using that for
a MappedScop::blockMappingSchedule, but then you don't
use that function anywhere in this PR.

skimo

skimo-openhub:
On Wed, Jun 13, 2018 at 11:11:37AM -0700, ftynse wrote:
> > @@ -517,6 +499,16 @@ isl::multi_aff constructThreadToWarp(
>  }
>  } // namespace
>  
> +isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
> +    const detail::ScheduleTree* tree) const {
> +  std::vector<mapping::MappingId> ids;
> +  for (int i = 0; i < numThreads.view.size(); ++i) {
> 
> We seem to compile with `-Wno-sign-compare`

Yes.  Apparently someone thought that was a good idea and
then left it to me and Theo to clean up the mess.

skimo

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: I don't think any such distinction should be made.  A mapping node should always map to all (relevant) ids of a type.  It would be better to have a single way to check for a mapping node
than to have several ways of doing it slightly differently.  Anyway, I'll just leave it like this for now.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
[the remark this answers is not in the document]
skimo-openhub: I fixed it up.  It was worse than just moving code around.  The moved code was *different* from the original code.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: On Wed, Jun 13, 2018 at 11:15:26AM -0700, ftynse wrote:
> Indeed, this is due to an artifically introduced commit.  The old function was just eras
ed in my original commit rather than going through progressive upates with extra code that
 just gets deleted by the subsequent commit.

Hmm... from the looks of it, it seems like you did fix this.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: On Wed, Jun 13, 2018 at 11:15:48AM -0700, ftynse wrote:
> I did, hence the explanatory comment, although it may have been at some previous version
 of the code.

Hmm... I guess it doesn't hurt, but it may point to a bug in isl,
in which case we should fix that instead of working around it.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: >
> not anymore

I think it's still pretty long and it would be easier
to read if this was abstracted out.  I'll leave it like this for now.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: On Wed, Jun 13, 2018 at 11:17:18AM -0700, ftynse wrote:
> No, I wrote a completely new `isPromotableToRegistersBelow` (including the unrolling par
t that I separated in a different commit) and then called it from `isPromotableToRegisters
BelowThreads`.

Let's try and avoid reimplementing the same functionality.
I removed these changes.  If you would still like
to split up this function, then please do it 
in a separate PR.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: > That would be very annoying because this would be another artificial commit with potenti
al problems and potentially new code that will be erased by the subsequent commit.

Not if you do it properly.
(So not the way I initially pushed out.)

test/test_cuda_mapper_memory_promotion.cc
skimo-openhub: #517

ftynse:
> How? I only see you adding a commit using that for a MappedScop::blockMappingSchedule, but then you don't use that function anywhere in this PR.

Apparently it got lost in rebases and force-pushes.

ftynse:
> Yes. Apparently someone thought that was a good idea and then left it to me and Theo to clean up the mess.

I removed it locally and get plenty of warnings from caffe2 headers.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -798,5 +799,42 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
...
+ */
+isl::multi_union_pw_aff extractDomainToIds(
+    const detail::ScheduleTree* tree,
+    const std::vector<mapping::MappingId>& ids,
+    isl::id tupleId) {
+  using namespace polyhedral::detail;
+
+  auto space = isl::space(tree->ctx_, 0);
+  auto empty = isl::union_set::empty(space);
+  if (tupleId) {
```
skimo-openhub: Why did you introduce this test?
All callers specify an id so this very much looks like dead code.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -323,6 +323,24 @@ bool promotionImprovesCoalescing(
...
+isl::union_set collectMappingsTo(const Scop& scop) {
+  auto root = scop.scheduleRoot();
+  auto domain = scop.domain();
+  auto mappingFilters = detail::ScheduleTree::collect(
+      root, detail::ScheduleTreeType::MappingFilter);
+  mappingFilters = functional::Filter(isMappingTo<MappingType>, mappingFilters);
+  auto mapping = isl::union_set::empty(domain.get_space());
+  for (auto mf : mappingFilters) {
+    auto filterNode = mf->elemAs<detail::ScheduleTreeElemMappingFilter>();
+    mapping = mapping.unite(filterNode->filter_);
```
skimo-openhub: Doesn't this have the same issue as #505
(which you have since solved in `extractDomainToIds`)?

Actually, this looks like duplicate code wrt `extractDomainToIds`
(the mapping is represented differently, but conceptually, it's the same operation),
but I guess we can fix that up later.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -323,6 +323,24 @@ bool promotionImprovesCoalescing(
...
+  }
+  return domain.intersect(mapping);
```
skimo-openhub: Why do you need to intersect with the domain?

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -799,6 +799,20 @@ void orderAfter(ScheduleTree* root, ScheduleTree* tree, isl::union_set filter) {
...
+isl::union_set activeDomainPointsNoMapping(
+    const detail::ScheduleTree* root,
+    const detail::ScheduleTree* tree) {
+  auto domain = root->elemAs<detail::ScheduleTreeElemDomain>()->domain_;
+  for (auto t : tree->ancestors(root)) {
+    if (auto f = t->elemAs<detail::ScheduleTreeElemFilter>()) {
+      domain = domain.intersect(f->filter_);
+    }
+  }
+  return domain;
```
skimo-openhub: This looks like it will return an empty set if `tree` is inside an extension.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: > Doesn't this have the same issue as #505
>  (which you have since solved in extractDomainToIds)?

Good point.

> Actually, this looks like duplicate code wrt extractDomainToIds
> (the mapping is represented differently, but conceptually, it's the same operation),
> but I guess we can fix that up later.

Yes, there are increasingly many places that need mapping in both forms: mupa and union set.  It should be relatively easy to extract a standalone conversion function from mupa to union set and always return mupa.  Or introduce an opaque "mapping" object that can be viewed as both types.  Created #519

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: To have the domain be consistent with the domain of the function returned by `extractDomainToIds`

tc/core/polyhedral/schedule_transforms.cc
skimo-openhub: On Fri, Jun 15, 2018 at 01:33:06AM -0700, ftynse wrote:
> To have the domain be consistent with the domain of the function returned by `extractDom
ainToIds`

Mention that in the commit message and mention
the intersection in the comment.

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
skimo-openhub: On Fri, Jun 15, 2018 at 01:33:06AM -0700, ftynse wrote:
> To have the domain be consistent with the domain of the function returned by `extractDom
ainToIds`

Mention that in the commit message and mention
the intersection in the comment.

PR #500, opened 2018-06-11
Fix undefined variable errors

nicolasvasilache:
For some reason they only triggered for me when trying to rebase tc_check.
This is still relevant as an independent fix.


ftynse:
What was the exact error message?  Was it an error or an escalated warning?  This looks surprising that the variables were considered _undefined_.  You seem to have only added the non-default initialization.  I'd like to understand better what happened here...

nicolasvasilache:
@ftynse see #504 (it should show in the CI)

nicolasvasilache:
@ftynse reworded the PR

PR #453, opened 2018-05-28
Genetic Tuning Improvements

ttheodor:
this PR replaces #160

PR #501, opened 2018-06-11
Autotuning improvements

nicolasvasilache:
This PR isolates a small subset from @ttheodor's #453.
I took the commits that looked independent, fix clear bugs or implement simple new behavior and do not touch autotuner-inl.h (which underwent some changes related to early exit recently and for which I need to look deeper).


tc/autotuner/genetic_search.cc
```diff
@@ -166,6 +195,7 @@ GeneticSearch::GeneticSearch(
       lastBestConf(confs[0]),
       numGenerations(numGenerations),
       maxPopulationSize(populationSize),
+      matingPoolSize(populationSize * 3),
```
ttheodor: In one of the subsequent commits I introduced a separate flag for the mating pool size (instead of hardcoding `populationSize * 3`).

PR #506, opened 2018-06-12
Drop implicit tuner cache interactions

### LINKED ISSUE #473: Do not store mapping options proto after tuning unless explicitly asked to

nicolasvasilache:
closes #473


tc/autotuner/autotuner-inl.h
```diff
@@ -438,28 +398,14 @@ Autotuner<Backend, SearchStrategy>::tune(
...
-  if (restoredCandidates.size() > 0) {
-    startingPoints.reserve(1 + restoredCandidates.size());
-    std::move(
-        restoredCandidates.begin(),
-        restoredCandidates.end(),
-        std::back_inserter(startingPoints));
-  }
-
   // Create initial configs based on options + model configuration
+  const std::vector<typename Backend::MappingOptionsType> strategies{
```
ftynse: I don't get why a vector of "MappingOptionsType" is called `strategies`?

tc/autotuner/autotuner.h
```diff
@@ -164,12 +164,13 @@ class Autotuner {
...
       std::unordered_map<size_t, std::vector<const DLTensor*>>& outputs,
-      const MappingOptionsType& baseMapping,
-      const std::string& cacheFileName = "",
+      const std::vector<MappingOptionsType>& baseMapping,
       const TuningParameterFixer& fixedParams = TuningParameterFixer());
 
- private:
-  std::shared_ptr<OptionsCache<Backend>> optionsCache_;
+ public:
+  /// This is accessed by multiple threads in the tuning harness, be sure to
```
ftynse: Shouldn't it be guarded by some mutex then?

tc/autotuner/options_cache-inl.h
```diff
@@ -317,9 +318,41 @@ void OptionsCache<Backend>::fromProtobuf(
...
 
-} // namespace autotune
+template <typename Backend>
+std::vector<typename Backend::MappingOptionsType> loadTopKFromCache(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::string& cacheFileName,
+    const std::vector<const DLConstTensor*>& inputs,
+    const size_t numCandidates) {
+  LOG(INFO) << "Loading proto from: " << cacheFileName << std::endl;
```
ftynse: Do we still need these logs?

tc/autotuner/options_cache-inl.h
```diff
@@ -317,9 +318,41 @@ void OptionsCache<Backend>::fromProtobuf(
...
+  if (cacheFileName.empty()) {
```
ftynse: Now that this is a separate function, let's get rid of this special-casing from the tuner that was necessary to avoid saving.  Just let it go further and throw on attempting to open a non-existent file.

tc/autotuner/options_cache-inl.h
```diff
@@ -317,9 +318,41 @@ void OptionsCache<Backend>::fromProtobuf(
...
+}
 
-inline std::string makeOptionsFilename(const std::string& fn) {
-  return fn + ".options";
+template <typename Backend>
+void storeTopKToCache(
+    const std::shared_ptr<OptionsCache<Backend>>& cache,
+    const std::string& cacheFilename,
+    uint32_t count) {
+  if (cacheFilename.empty()) {
```
ftynse: Let's drop this special casing

tc/autotuner/options_cache-inl.h
```diff
@@ -317,9 +318,41 @@ void OptionsCache<Backend>::fromProtobuf(
...
+    LOG(INFO) << "No filepath provided, not saving cache" << std::endl;
+  } else {
+    LOG(INFO) << "Dumping cache to " << cacheFilename << std::endl;
+    cache->pruneKeepTopK(count);
```
ftynse: Guessing from the function name, I'd expect it to store top-k options to the given file but keep the cache object itself intact.  Here it looks like it prunes _from the cache_.  Please address this before landing.

tc/autotuner/options_cache.h
```diff
@@ -163,9 +163,22 @@ struct OptionsCache {
...
+template <typename Backend>
+std::vector<typename Backend::MappingOptionsType> loadTopKFromCache(
+    const std::string& tc,
+    const std::string& entryPoint,
+    const std::string& cacheFileName,
+    const std::vector<const DLConstTensor*>& inputs,
+    const size_t numCandidates);
+
+template <typename Backend>
+void storeTopKToCache(
```
ftynse: some documentation?  the behavior of the function is not obvious from its name (see above)

tc/examples/blockdiagperm.cc
```diff
@@ -72,14 +79,23 @@ def blockdiagperm2dfissioned_2(float(B, N) I, int32(N) Idx) -> (O) {
...
   at::Tensor I = at::CUDA(at::kFloat).rand({128, 10, 50});
   at::Tensor W = at::CUDA(at::kFloat).rand({10, 50, 50});
-  auto options = tc::CudaMappingOptions::makeNaiveMappingOptions();
+  auto options = tc::autotune::loadTopKFromCache<tc::CudaBackend>(
+      tc,
+      "blockdiagperm2dfissioned_1",
+      FLAGS_options_cache,
+      extractRawPtrs(tc::aten::makeDLConstTensors({I, W})),
+      1);
+  if (options.size() == 0) {
```
ftynse: I'd rather not even load if the name is empty/wrong.

tc/examples/blockdiagperm.cc
```diff
@@ -116,12 +142,19 @@ def blockdiagperm2dfissioned_2(float(B, N) I, int32(N) Idx) -> (O) {
     tc::aten::uncheckedRun(*pExecutor, {I, W}, outputs);
     tc::aten::uncheckedRun(*pExecutor2, {O, Idx}, outputs2);
   }
+
+  // 5. Explicitly store the best found options in cache so we can reinforce
+  // later if required
+  tc::autotune::storeTopKToCache(
```
ftynse: I'd rather not call "store" if the path is empty explicitly, then have an implicit behavior that does not store (and logs) given an empty path

README.md
```diff
@@ -41,50 +41,48 @@ def tensordot(float(N, C1, C2, H, W) I0,
...
+            << "size I1: " << I1.sizes()
+            << " ran in: " << timings.kernelRuntime.toMicroSeconds() << "us\n";
 }
 ```
 
 Note that we only need to **autotune a TC once** to obtain reasonable mapping options
 that can translate to other problem sizes for a given TC as the following snippet
 illustrates:
 ```cpp
-// 5. Reuse bestOptions from autotuning on another kernel
```
ftynse: spurious change

tc/benchmarks/README.md
```diff
@@ -0,0 +1,40 @@
...
+If you have been following the instructions given [here](https://facebookresearch.github.io/TensorComprehensions/installation.html), you can use the command:
+
+```
+BUILD_TYPE=Release WITH_CAFFE2=ON CLANG_PREFIX=$(${CONDA_PREFIX}/bin/llvm-config --prefix) ./build.sh
+```
+
+# Running the autotuner manually
+By default a full evolutionary search is run with 25 generations and 100 candidates per generation. This will take some time for some of the kernels. This setting be changed by using the proper gflags options: ```--tuner_gen_generations``` and ```--tuner_gen_pop_size```.
+
+For instance, a more reasonable search could iterate as such:
```
ftynse: This is not a "more reasonable search", but a "shorter tuning session".  Let's not send a message that our default options are unreasonable

tc/autotuner/autotuner.h
nicolasvasilache: it is of course guarded, I'll update the comment

PR #508, opened 2018-06-12
[DO-NOT-MERGE] extractDomainToThread: differentiate between different branches

### LINKED ISSUE #505: extractDomainToThread: take into account filters

skimo-openhub:
Within a given schedule tree (different) elements of the same statements
may get mapped to threads in different branch (and therefore possibly
also using a different mapping).  This happens in particular
in case of full/partial tile separation for reduction detection.
The domains of these separate mappings need to be made disjoint
to prevent them from interacting (in particular, getting summed
together).

Closes #505

Note that this conflicts with #489, which extracts out an
`extractMappingToIds` from `extractDomainToThread`
for some reason.


skimo-openhub:
It looks like #489 will solve this, hopefully in a better way.

PR #510, opened 2018-06-12
makeIslAffFromExpr: only accept parametric expressions

skimo-openhub:
The function makeIslAffBoundsFromExpr currently relies on the names
of the set dimensions in a space.  Even though the dependence
is very localized and it is unlikely that those names would
get dropped, it is better to not rely on them at all.

First change the users of makeIslAffFromExpr to only create expressions
in terms of parameters and then change makeIslAffFromExpr
to only construct those.

Start by collecting the constraints on the outer loop iterators
in terms of parameters and only convert them into set dimensions
at the point where the iteration domain is constructed.

Treating the outer loop iterators as parameters at first
makes sense because they have a fixed value at the point
where the statement is executed.
It is only at the point where the iteration domain is constructed
that the loop iterators should no longer be considered as parameters.

### DIFF
--- a/tc/core/halide2isl.cc
+++ b/tc/core/halide2isl.cc
@@ -251,31 +251,25 @@ isl::set makeParamContext(isl::ctx ctx, const ParameterVector& params) {
 namespace {
 
 isl::map extractAccess(
-    isl::set domain,
+    const IterationDomain& domain,
     const IRNode* op,
     const std::string& tensor,
     const std::vector<Expr>& args,
     AccessMap* accesses) {
   // Make an isl::map representing this access. It maps from the iteration space
   // to the tensor's storage space, using the coordinates accessed.
+  // First construct a set describing the accessed element
+  // in terms of the parameters (including those corresponding
+  // to the outer loop iterators) and then convert this set
+  // into a map in terms of the iteration domain.
 
-  isl::space domainSpace = domain.get_space();
-  isl::space paramSpace = domainSpace.params();
+  isl::space paramSpace = domain.paramSpace;
   isl::id tensorID(paramSpace.get_ctx(), tensor);
-  auto rangeSpace = paramSpace.named_set_from_params_id(tensorID, args.size());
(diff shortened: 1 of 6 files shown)


tc/core/halide2isl.cc
```diff
@@ -285,19 +279,34 @@ isl::map extractAccess(
 
     // The coordinate written to in the range ...
     auto rangePoint =
-        isl::pw_aff(isl::local_space(rangeSpace), isl::dim_type::set, i);
-    // ... equals the coordinate accessed as a function of the domain.
-    auto domainPoint = halide2isl::makeIslAffFromExpr(domainSpace, args[i]);
+        isl::pw_aff(isl::local_space(tensorSpace), isl::dim_type::set, i);
```
ftynse: Can't we construct `rangePoint` in the parameter space and unbind it as well, to avoid using `dim_type::set`.  For example, using dummy parameters.  Alternatively, create an identity `multi_pw_aff` and take individual pw_affs from it?

skimo-openhub:
On Wed, Jun 27, 2018 at 05:57:55AM -0700, ftynse wrote:
> > @@ -285,19 +279,34 @@ isl::map extractAccess(
>  
>      // The coordinate written to in the range ...
>      auto rangePoint =
> -        isl::pw_aff(isl::local_space(rangeSpace), isl::dim_type::set, i);
> -    // ... equals the coordinate accessed as a function of the domain.
> -    auto domainPoint = halide2isl::makeIslAffFromExpr(domainSpace, args[i]);
> +        isl::pw_aff(isl::local_space(tensorSpace), isl::dim_type::set, i);
> 
> Can't we construct `rangePoint` in the parameter space and unbind it as well, to avoid using `dim_type::set`.  For example, using dummy parameters.  Alternatively, create an identity `multi_pw_aff` and take individual pw_affs from it?

The latter is what I did in #479 (which you already approved, but which
depends on this one).

skimo

Issue #511, opened 2018-06-12
Indexing with byte arrays.


mdouze:
- OS: Linux
- How you installed TC (docker, conda, source): conda
- Python version: 3.6.5
- CUDA/cuDNN version: N/A
- Conda version (if using conda): 4.5.0

In addition, including the following information will also be very helpful for us to diagnose the problem:
- A script to reproduce the issue (highly recommended if its a build issue)

```
import tensor_comprehensions as tc
import numpy as np
import torch

N = 10 ** 7
M = 32

codes = np.random.randint(1<<32, size=(N, M // 4)).astype('uint32')
codes = codes.view('uint8')

codes_t = torch.from_numpy(codes)

luts = np.random.randn(M, 256).astype('float32')
luts_t = torch.from_numpy(luts)

lang = """
def mindis(float(M, 256) L, uint8(N, M) codes) -> (v, s) {
    s(i) +=! L(j, codes(i, j))
    v min=! s(i)
}
"""

mindis = tc.define(lang, name="mindis")

vmin = mindis(luts_t, codes_t)
```

- Error messages and/or stack traces of the issue (create a gist)

```
[WARNING]: No mapping options passed, 'naive' type mapping options will be used and will likely have bad performance. See help(your_layer.__call__) for setting mapping options.
[ERROR]: Caught Exception: Internal error at /opt/conda/conda-bld/halide_1520457560724/work/src/IR.cpp:560
Condition failed: args[i].type() == Int(32)
Args to load from image must be type Int(32)
```

- Context around what you are trying to do

This code is to find the min value of a distance estimation with a large number of Product Quantization codes. 
The problems in the code above are: 
- the error above
- when adding a cast `int32(codes(i, j))`, the code crashes with https://gist.github.com/mdouze/472248f7d4205dc5a96464d0851a1a7e
- we are not interested in s, it is a temp variable
- we are interested in the argmin, but this seems to be on the roadmap.

ftynse:
Seems to come from Halide pass, related to #98

abadams:
Needs the changes in #262 at the least. In trunk the declarations for types smaller than int32 don't even exist at the cuda level.

ftynse:
Why does it need #262?  There is no indirection on the _left_ hand-side in this code and I think we cherry-picked typedefs

nicolasvasilache:
yeah #262 is unrelated, should be an easy one, I'll take it

abadams:
[...]

PR #512, opened 2018-06-12
Types support and min_distance.py function example

nicolasvasilache:
Addresses #511, naive options perf of the 5 kernel variant seems reasonable for a start.
A lot more work is needed to get the single kernel and the 1-staged impl to good performance so the 2-staged version will probably be a nice low hanging fruit.

### DIFF
--- a/tc/core/libraries.h
+++ b/tc/core/libraries.h
@@ -31,9 +31,20 @@ namespace code {
 namespace c {
 
 constexpr auto types = R"C(
+// Can't include system dependencies with NVRTC
+// Can't include cuda_fp16.h with NVRTC due to transitive system dependencies
+// #include <cuda_fp16.h>
+
 // Halide type handling
+typedef char int8;
+typedef short int16;
 typedef int int32;
 typedef long int64;
+typedef unsigned char uint8;
+typedef unsigned short uint16;
+typedef unsigned int uint32;
+typedef unsigned long uint64;
+// typedef half float16;
 typedef float float32;
 typedef double float64;
 )C";
--- a/tc/core/tc2halide.cc
+++ b/tc/core/tc2halide.cc
@@ -53,10 +53,17 @@ Type translateScalarType(int tcType) {
       return Int(32);
     case lang::TK_INT64:
       return Int(64);
+    case lang::TK_FLOAT16:
+      return Float(16);
+    case lang::TK_FLOAT32:
+      return Float(32);
+    case lang::TK_FLOAT64:
+      return Float(64);
     case lang::TK_FLOAT:
       return Float(32);
     case lang::TK_DOUBLE:
       return Float(64);
+
     default:
       LOG(FATAL) << "Unhandled TC scalar type: " << tcType << '\n';
       return Type();
--- a/tc/lang/lexer.h
+++ b/tc/lang/lexer.h
@@ -41,8 +41,6 @@ namespace lang {
   _(TK_MIN, "min", "min")                        \
   _(TK_MAX, "max", "max")                        \
   _(TK_WHERE, "where", "where")                  \
-  _(TK_FLOAT, "float", "float")                  \
-  _(TK_DOUBLE, "double", "double")               \
   _(TK_DEF, "def", "def")                        \
   _(TK_ARROW, "arrow", "->")                     \
   _(TK_EQUIVALENT, "equivalent", "<=>")          \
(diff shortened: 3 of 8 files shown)


tc/core/tc2halide.cc
```diff
@@ -53,10 +45,83 @@ Type translateScalarType(int tcType) {
...
+      return UInt(16, 4);
+    case lang::TK_VECTOR4_UINT32:
+      return UInt(32, 4);
+    case lang::TK_VECTOR4_UINT64:
+      return UInt(64, 4);
+    case lang::TK_VECTOR4_FLOAT16:
+      return Float(16, 4);
+    case lang::TK_VECTOR4_FLOAT32:
+      return Float(32, 4);
+    case lang::TK_VECTOR4_FLOAT64:
```
abadams: Are these used in this PR? Explicit vector types in the front-end seems like a bad idea to me. I thought that Halide would reject them anyway at Func definition time - it views vectorization as a property of a loop, and thus target-dependent and implicit, not as a property of the data type.

nicolasvasilache:
cc @Ravenwater

nicolasvasilache:
Sounds, good I'll put them in a WIP PR

@abadams they are currently used in `test/cuda/test_compile_and_run.cc`, I didn't have time to implement checks before leaving last night so it may well be incorrect code.

Re. types it is one of the goals of TC to help easily implement and experiment with various types (quantized, posits, vector, sparse etc). Is it envisionable to have the pieces of Halide we use currently view ADTs as first class citizens? If not, at least let us propagate more exotic types to the TC mapper.
Then we can easily implement low-level operations as libraries in the generated code (like the strawman I did for the vector types in library.h).

abadams:
In Halide, mathematical vectors, complex numbers, etc are represented as Tuples. The vector types are for simd, which is just a way of executing a loop, not a data type. E.g. simd on the next version of arm doesn't necessarily have a statically known width, so that's not a useful data type. If we want vector front-end types for the purpose of representing vectors, e.g. in R^3, we should lower them to Halide Tuples.

tc/core/libraries.h
```diff
@@ -31,9 +31,20 @@ namespace code {
...
 // Halide type handling
+typedef char int8;
+typedef short int16;
 typedef int int32;
 typedef long int64;
+typedef unsigned char uint8;
+typedef unsigned short uint16;
+typedef unsigned int uint32;
+typedef unsigned long uint64;
+// typedef half float16;
```
ftynse: This essentially means that the compilation will fail with some, obscure for the user, message from nvrtc when attempting to use `float16`.  I'd rather ditch it, or add a TC compilation error saying "float16 not supported because nvrtc does not want it".

tc/lang/lexer.h
```diff
@@ -67,15 +65,21 @@ namespace lang {
...
   _(TK_UINT8, "uint8", "uint8")                  \
   _(TK_UINT16, "uint16", "uint16")               \
   _(TK_UINT32, "uint32", "uint32")               \
   _(TK_UINT64, "uint64", "uint64")               \
-  _(TK_BOOL, "bool", "bool")                     \
+  _(TK_INT8, "int8", "int8")                     \
+  _(TK_INT16, "int16", "int16")                  \
+  _(TK_INT32, "int32", "int32")                  \
+  _(TK_INT64, "int64", "int64")                  \
+  _(TK_FLOAT16, "float16", "float16")            \
```
ftynse: Do we want to have _both_ `float32` and `float`?  Do we want to differentiate?

tc/lang/sema.h
```diff
@@ -32,22 +32,26 @@ struct TypeInfo {
   TypeInfo(Code code_, uint8_t bits_) : code_(code_), bits_(bits_) {}
   TypeInfo(TreeRef scalar_type) {
     switch (scalar_type->kind()) {
-#define TYPE_INFO_OPTION(tok, c, b) \
-  case tok:                         \
-    code_ = c;                      \
-    bits_ = b;                      \
+#define TYPE_INFO_OPTION(tok, c, b, l) \
```
ftynse: what is `l` user for?

tc/lang/lexer.h
nicolasvasilache: Given the semantics for the other types it is consistent.
Removing float will break codes.
I don't see a big issue in the duplication here.

PR #514, opened 2018-06-14
[WIP] User-facing python and PyTorch integration

nicolasvasilache:
WIP, for information purposes only atm


test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+    #
+    # This test implements single kernel forward and 2 kernel backward with
+    # tc.define_with_autograd. The performance of the backward is expected to be
+    # significantly better because the loops types in the single kernel may not
+    # be currently fused profitably (need to investigate the fused case deeper).
+    # The API is more bloated than hoped but it is unclear we can do better,
+    # suggestions most welcome.
+    # Maybe consider making cache_filename and tuner_config global parameters.
+    #
+    def test_conv_backward_2kernels(self):
```
nicolasvasilache: cc @ jopamer @ftynse @apaszke comments most welcome here

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+        mean, var, out = group_norm(I, gamma, beta)
+        # TODO: test results!!!
+
+    #
+    # This test implements single kernel forward and backward with
+    # tc.define_with_autograd. This is the "light touch" mode and the function call
+    # may already be too verbose.
+    # Maybe consider making cache_filename and tuner_config global parameters.
+    #
+    def test_conv_backward_fused(self):
```
nicolasvasilache: cc @ jopamer @ftynse @apaszke comments most welcome here

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+    # This test implements group normalization as 2 TC kernels using the
+    # tc.define_with_autograd abstraction. In particular it is ** NOT **
+    # possible to insert reshapes between TCs moments operate on a 5-D tensor
+    # (Note: loop collapsing modulo conforming strides is an easy TC-level
+    #  transformation with nice expected benefits).
+    # The tc.define_with_autograd API is more bloated than hoped but it is
+    # unclear we can do better, suggestions most welcome.
+    # Maybe consider making cache_filename and tuner_config global parameters.
+    #
+    def test_group_norm_2kernels_autograd(self):
```
nicolasvasilache: cc @ jopamer @ftynse @apaszke comments most welcome here

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+
+    #
+    # This test implements group normalization as 2 TC kernels using the
+    # tc.define abstraction. In particular it is possible to insert reshapes
+    # between TCs and have moments operate on a 2-D view which is mapped
+    # more efficiently to CUDA at this time
+    # (Note: loop collapsing modulo conforming strides is an easy TC-level
+    #  transformation with nice expected benefits).
+    #
+    def test_group_norm_2kernels(self):
```
nicolasvasilache: cc @jopamer @ftynse @apaszke comments most welcome here

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+        add(A, B, outputs=[out], unchecked=True)
+        # TODO: test results!!!
 
-from tensor_comprehensions import TcCompilationUnit
-from common import TestCase, run_tests
+    #
+    # This test implements group normalization as a single TC kernel.
+    # Performance is not expected to be as good as when using 2 kernels.
+    #
+    def test_group_norm_fused(self):
```
nicolasvasilache: cc @jopamer @ftynse @apaszke comments most welcome here

test_python/test_tc.py
nicolasvasilache: @jopamer

test_python/test_tc.py
nicolasvasilache: @jopamer

test_python/test_tc.py
nicolasvasilache: @jopamer

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+                d_O(  n, r_m, h - r_kh, w - r_kw) * W1(r_m, c, r_kh, r_kw)
+            d_W1(m, c, kh, kw) +=!
+                d_O(r_n,   m, r_h - kh, r_w - kw) *  I(r_n, c,  r_h,  r_w)
+        }
+        """
 
-if not os.path.exists(PATH_PREFIX):
-    os.makedirs(PATH_PREFIX)
+        N, C, H, W, O, kH, kW = 32, 4, 56, 56, 16, 1, 1
+        cache_filename = "/tmp/" + str(uuid.uuid4())
```
apaszke: It's generally better to do
```python
with tmpfile.NamedTemporaryFile() as f:
    ...
```
because it guarantees that it will get cleaned up even when an error is raised

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+        convolution = tc.define_with_autograd(
+            conv,
+            forward_entry_points=("convolution", ),
+            backward_entry_points=("convolution_igrad", "convolution_wgrad"),
+            cache_filename=cache_filename,
+            # First: (I, W1, d_O) -> (W1, d_O) => (1, 2)
+            # Second: (I, W1, d_O, d_I) -> (I, d_O) => (0, 2)
+            backward_input_indices=((1, 2), (0, 2), ),
+            tuner_config=tuner_config)
+        I = Variable(torch.randn(N, C, H, W).cuda(), requires_grad=True)
```
apaszke: Variables are deprecated now. Do `tensor.randn(N, C, H, W, device='cuda', requires_grad=True)` (this is much faster too, because it uses a CUDA PRNG)

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+        W = Parameter(torch.randn(O, C, kH, kW).cuda())
+        # First occurrence on a new cache filename triggers tuning
+        out, = convolution(I, W)
+        out.sum().backward()
+        # Subsequent occurrences do not
+        out, = convolution(I, W)
+        out.sum().backward()
+        # TODO: test results!!!
```
apaszke: You might be interested in `torch.autograd.gradcheck` (see e.g. right above [this paragraph](https://pytorch.org/docs/stable/notes/extending.html?highlight=gradcheck#extending-torch-nn)).

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+            conv,
+            forward_entry_points=("convolution", ),
+            backward_entry_points=("convolution_igrad", "convolution_wgrad"),
+            cache_filename=cache_filename,
+            # First: (I, W1, d_O) -> (W1, d_O) => (1, 2)
+            # Second: (I, W1, d_O, d_I) -> (I, d_O) => (0, 2)
+            backward_input_indices=((1, 2), (0, 2), ),
+            tuner_config=tuner_config)
+        I = Variable(torch.randn(N, C, H, W).cuda(), requires_grad=True)
+        W = Parameter(torch.randn(O, C, kH, kW).cuda())
```
apaszke: `Parameter`s are unnecessary too in here (they are functionally equivalent to regular tensors). Also, would be good to test autodiff wrt. weights as well.

test_python/test_tc.py
apaszke: There's also `gradgradcheck` if you care about higher order derivatives

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+            cache_filename=cache_filename,
+            tuner_config=tuner_config)
+        I = Variable(torch.randn(N, G, D, H, W).cuda())
+        gamma = Variable(torch.randn(G, D).cuda())
+        beta = Variable(torch.randn(G, D).cuda())
+        # First occurrence on a new cache filename triggers tuning
+        mean, var, out = group_norm(I, gamma, beta)
+        # Subsequent occurrences do not
+        mean, var, out = group_norm(I, gamma, beta)
+        # TODO: test results!!!
```
apaszke: It would be good to test that the tuning did actually happen somehow. We've got bitten by that when we were writing the JIT tests (after some changes, parts that were supposed to run some compiled versions weren't doing that)

test_python/test_tc.py
```diff
@@ -12,89 +12,499 @@
...
+        print("\nTune group_norm from naive options, don't store best options")
+        group_norm.tune(
+            I,
+            gamma,
+            beta,
+            mean.view((N, G)),
+            var.view((N, G)),
+            starting_options=tc.MappingOptions('naive'),
+            tuner_config=tuner_config)
+        print("\nRun tuned group_norm with best options")
```
apaszke: I'd generally advise against using `print`s in tests. Their whole point is to be automatic, so they won't help you too much.

test_python/test_tc.py
apaszke: Also, since the `TODO: test results` appears in so many places, it's probably a good idea to have a common testing procedure with a signature similar to `gradcheck` (take the function and example inputs, do some checks). This way you write it once and any subsequent improvements apply to all functions you check.

test_python/test_tc.py
nicolasvasilache: yay RAII, thanks!

test_python/test_tc.py
nicolasvasilache: inherited from older impl, killed, thanks!

test_python/test_tc.py
nicolasvasilache: inherited from older impl, killed, thanks!

test_python/test_tc.py
nicolasvasilache: mostly inherited from older impl, killed, thanks!

nicolasvasilache:
This has served its purpose, thanks for your comments, I'll close as is so we can keep that for later reference. The "configure a function composition via index lists" is similar to the caffe2 impl but after chatting with @apaszke it is clear that this is not worth the trouble in PyTorch. For the language features I wanted, well I can just use Python.. For a pure C++ solution we punt for now.

docs/source/framework/pytorch_integration/writing_layers.rst
```diff
@@ -1,137 +1,190 @@
...
-generated. Therefore, it is important to use good :code:`Options` for running a
-kernel. You can read more about mapping options here - :ref:`tc_mapping_options`.
+        # Setup code for mm and tensors as above
+        cache = tc.MappingOptionsCache(cache_filename)
+        best_options, = cache.load(mm, entry_point, (A, B), 1)
+        matmul = tc.define(mm, entry_point="matmul", fallback=best_options)
+        C, = matmul(A, B)
 
-There are two ways to set the :code:`Options`:
+One may also create a TC without fallback options and call compile explicitly
```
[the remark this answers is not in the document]
nicolasvasilache: So for this I unfortunately do not have a good answer:
1. yes there is a state, the latest compiled executor for a TC def + input sizes is the one that is memoized. I don't want to start putting protobuf + hashing on the critical path;
2. implicit compilation is made loud (instead of generating failures) and will get into tuning mode if no options are passed;
3. I don't really see a good, usable, alternative to the above for people who "just want to get a PyTorch layer working". Implicit compilation, tuning, memoization is pleasant to use but can of course come with surprises.

Let's chat more on the next PR, @apaszke your opinion would also be valuable here.

docs/source/framework/pytorch_integration/writing_layers.rst
```diff
@@ -1,137 +1,190 @@
...
+----------
 
-An example for how to pass options:
+Tuning can be achieved by constructing a TC and calling :code:`tune` on it.
+If the optional parameter :code:`cache_filename` is provided, the best options
+will be loaded from file via a :code:`MappingOptionsCache`
+object and will be used as a starting point. If additionally, the optional
+parameter :code:`store_to_cache` is set to True, tuning will append the best
+options to the cache. In the absence of a cache filename, tuning will start
+from :code:`tc.MappingOptions('naive')`.
```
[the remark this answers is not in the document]
nicolasvasilache: updated that

python/examples/min_distance.py
```diff
@@ -111,25 +112,30 @@ def argmin_1d(float(N) S, int32(D) MinIdx) -> (min_idx) {
...
+    lang, entry_point="min_1d", fallback=tc.MappingOptions('naive'))
+argmin_2d = tc.define(
+    lang, entry_point="argmin_2d", fallback=tc.MappingOptions('naive'))
+argmin_1d = tc.define(
+    lang, entry_point="argmin_1d", fallback=tc.MappingOptions('naive'))
+
+S, = reduce_codes(luts_t, codes_t)
+V, = min_2d(S.view((D, N / D)))
+v, = min_1d(V)
+MinIdx, = argmin_2d(S.view((D, N / D)), v)
```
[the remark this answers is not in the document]
nicolasvasilache: I have used lower case v because it is a value, single capital letter for parameters, CamelCase for tensor names and snake_case is for tc names as python objects.

I see there is a potential collision for V which looks like a parameter but is a tensor.
How about v, V -> minValue, MinTensor

python/examples/min_distance.py
```diff
@@ -169,22 +171,28 @@ def min_1d(float(K, D) V2) -> (V) {
     V(k) min=! V2(k, r_d)
 }
 def argmin_2d(float(K, D, NBYD) S, float(K) V) -> (MinIdx2) {
-    MinIdx2(k, d) min=! (S(k, d, r_nbyd) == V(k)) ? d * NBYD + r_nbyd : N
+    MinIdx2(k, d) min=!
+        (S(k, d, r_nbyd) == V(k)) ? d * NBYD + r_nbyd : NBYD * D
```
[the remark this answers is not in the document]
nicolasvasilache: not strictly but the more I write them the more I seem to

python/examples/tc_pybind_example.py
```diff
@@ -30,17 +28,13 @@
 from tensor_comprehensions.tclib import MappingOptions
 
 # Define a timing function to print some results
-def time_tc(iters, prepend, runFun, tc_name, inputs):
-    global dump_backward_overhead
-
+def time_tc(iters, prepend, runFun, tc_entry_point, inputs):
```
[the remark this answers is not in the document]
nicolasvasilache: fixed

PR #518, opened 2018-06-14
no longer consider ScheduleTreeElemMappingFilter to be a subclass of filter

skimo-openhub:
This type was introduced as filter because it is used as a filter
during AST generation.  However, AST generation is performed on
an isl schedule tree and a ScheduleTreeElemMappingFilter can easily
be converted to an isl filter independently of whether
ScheduleTreeElemMappingFilter itself is a filter.
Having ScheduleTreeElemMappingFilter as a subclass is causing
more problems than it is worth because every time a reference
is made to a filter, a careful consideration has to be made
whether or not to include the mapping filters.
Simply stop considering ScheduleTreeElemMappingFilter objects as filters
and deal with them specifically when needed.

This PR also contains some collateral clean-ups.

### DIFF
--- a/tc/core/polyhedral/cuda/tighten_launch_bounds.cc
+++ b/tc/core/polyhedral/cuda/tighten_launch_bounds.cc
@@ -62,45 +62,27 @@ std::pair<size_t, size_t> rangeOfMappingParameter(
       static_cast<size_t>(max.get_num_si()));
 }
 
-// Look for nodes with no children.
-inline std::vector<const detail::ScheduleTree*> leaves(
-    const detail::ScheduleTree* tree) {
-  return functional::Filter(
-      [](const detail::ScheduleTree* st) { return st->numChildren() == 0; },
-      detail::ScheduleTree::collect(tree));
-}
-
 /*
  * Compute the maximal value attained by the mapping parameter "id".
  * Return std::numeric_limits<size_t>::max() if this value cannot
  * be determined.
  */
-size_t maxValue(const Scop& scop, const mapping::MappingId& id) {
+template <typename MappingIdType>
+size_t maxValue(const Scop& scop, const MappingIdType& id) {
+  using namespace polyhedral::detail;
+
   auto root = scop.scheduleRoot();
   auto params = scop.context();
   size_t sizetMax = std::numeric_limits<size_t>::max();
   size_t max = 0;
   size_t min = sizetMax;
-  auto nonSyncLeaves = functional::Filter(
-      [root, params](const detail::ScheduleTree* node) {
-        auto f = node->elemAsBase<detail::ScheduleTreeElemFilter>();
-        if (!f) {
-          return true;
-        }
-        if (f->filter_.n_set() != 1) {
-          std::stringstream ss;
-          ss << "In tree:\n"
-             << *root << "\nnot a single set in filter: " << f->filter_;
-          throw tightening::TighteningException(ss.str());
-        }
-        auto single = isl::set::from_union_set(f->filter_);
-        auto single_id = single.get_tuple_id();
-        return !Scop::isSyncId(single_id) && !Scop::isWarpSyncId(single_id);
-      },
-      leaves(root));
-  for (auto p : nonSyncLeaves) {
-    auto active = activeDomainPoints(root, p).intersect_params(params).params();
-    auto range = rangeOfMappingParameter(active, id);
+  auto filters = root->collect(root, ScheduleTreeType::MappingFilter);
+  filters = functional::Filter(isMappingTo<MappingIdType>, filters);
+  for (auto p : filters) {
+    auto mappingNode = p->elemAs<ScheduleTreeElemMappingFilter>();
+    auto active = activeDomainPoints(root, p).intersect_params(params);
+    active = active.intersect(mappingNode->filter_);
+    auto range = rangeOfMappingParameter(active.params(), id);
     min = std::min(min, range.first);
     max = std::max(max, range.second);
   }
--- a/tc/core/polyhedral/cuda/tighten_launch_bounds.cc
+++ b/tc/core/polyhedral/cuda/tighten_launch_bounds.cc
@@ -75,6 +75,50 @@ inline std::vector<const detail::ScheduleTree*> leaves(
       [](const detail::ScheduleTree* st) { return st->numChildren() == 0; },
       detail::ScheduleTree::collect(tree));
 }
+
+/*
+ * Compute the maximal value attained by the mapping parameter "id".
+ * Return std::numeric_limits<size_t>::max() if this value cannot
+ * be determined.
+ */
+size_t maxValue(const Scop& scop, const mapping::MappingId& id) {
+  auto root = scop.scheduleRoot();
+  auto params = scop.context();
+  size_t sizetMax = std::numeric_limits<size_t>::max();
+  size_t max = 0;
+  size_t min = sizetMax;
+  auto nonSyncLeaves = functional::Filter(
+      [root, params](const detail::ScheduleTree* node) {
+        auto f = node->elemAsBase<detail::ScheduleTreeElemFilter>();
+        if (!f) {
+          return true;
+        }
+        if (f->filter_.n_set() != 1) {
+          std::stringstream ss;
+          ss << "In tree:\n"
+             << *root << "\nnot a single set in filter: " << f->filter_;
+          throw tightening::TighteningException(ss.str());
+        }
+        auto single = isl::set::from_union_set(f->filter_);
+        auto single_id = single.get_tuple_id();
+        return !Scop::isSyncId(single_id) && !Scop::isWarpSyncId(single_id);
+      },
+      leaves(root));
+  for (auto p : nonSyncLeaves) {
+    auto range = rangeOfMappingParameter(root, p, params, id);
+    min = std::min(min, range.first);
+    max = std::max(max, range.second);
+  }
+  // Ignore min for now but there is a future possibility for shifting
+  LOG_IF(WARNING, min > 0)
+      << "Opportunity for tightening launch bounds with shifting -> min:"
+      << min;
+  // Inclusive range needs + 1 to translate to sizes
+  if (max < sizetMax) { // avoid overflow
+    return max + 1;
+  }
+  return sizetMax;
+}
 } // namespace
 
 // Takes grid/block launch bounds that have been passed to mapping and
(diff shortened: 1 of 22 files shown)


tc/core/polyhedral/mapping_types.h
```diff
@@ -28,6 +28,17 @@ struct MappingId : public isl::id {
  public:
   MappingId(const MappingId& id) : isl::id(id), dim(id.dim) {}
 
+  // Is "id" a mapping of the type provided as template argument?
+  template <typename MappingType>
+  bool isOfType() const {
```
ftynse: I would go with just `is`, which is sort of consistent with `as` that we use for conversions. 

We could also have an `as` here and return a null id if the type is wrong (given that we have implicit conversion to bool), but I don't like relying on implicit conversions.

tc/core/polyhedral/schedule_tree.h
```diff
@@ -344,6 +344,20 @@ struct ScheduleTree {
         detail::ScheduleTreeType::Sequence, std::forward<Args>(args)...);
   }
 
+  // Is "tree" a mapping filter that maps identifiers of the type provided as
+  // template argument?
+  template <typename MappingType>
+  static bool isMappingTo(const detail::ScheduleTree* tree) {
```
ftynse: Does it have to be a _member_ function?  I would just put it into `schedule_transforms.h` together with `activeDomainPoints` and schedule collectors.

tc/core/polyhedral/cuda/tighten_launch_bounds.cc
```diff
@@ -75,6 +75,50 @@ inline std::vector<const detail::ScheduleTree*> leaves(
       [](const detail::ScheduleTree* st) { return st->numChildren() == 0; },
       detail::ScheduleTree::collect(tree));
 }
+
+/*
+ * Compute the maximal value attained by the mapping parameter "id".
+ * Return std::numeric_limits<size_t>::max() if this value cannot
+ * be determined.
+ */
+size_t max(const Scop& scop, const mapping::MappingId& id) {
```
ftynse: Can we have a longer name here?   It's easy to confuse with `std::max` given that we are often `using namespace std`.

ftynse:
Took me some time to understand that 086800aa8850316927e20402fc9fbe64eb6c7d20 hoists the computation of mappingContext outside `rangeOfMappingParameter`.  A comment in the commit message is welcome.

ftynse:
30d75d618f8d2cad27583e0e2fd915a7b08308fb typo: "apply to proper filters are also to mapping filters."

skimo-openhub:
On Fri, Jun 15, 2018 at 03:49:34AM -0700, ftynse wrote:
> 
> cceea1b0401153432934d75c6922926bad1dbad5: do we always have all mapped identifiers in a mapping node? (I suppose so because the check does not fail)

Yes.

skimo

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -260,9 +260,9 @@ def fun(float(N, M) A, float(N, M) B) -> (C) {
     EXPECT_EQ(groups.size(), 3u);
 
     USING_MAPPING_SHORT_NAMES(BX, BY, BZ, TX, TY, TZ);
```
ftynse: this becomes unnecessary

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -260,9 +260,9 @@ def fun(float(N, M) A, float(N, M) B) -> (C) {
...
-    isl::space blockSpace = isl::space(ctx, 0);
-    isl::set blockZero =
-        isl::makeSpecializationSet<int>(blockSpace, {{BX, 0}, {BY, 0}});
+    isl::space tileSpace = isl::space(ctx, 0).unnamed_set_from_params(2);
+    auto tileZero =
+        isl::map(isl::multi_aff::zero(tileSpace.from_range())).range();
```
ftynse: This looks like a very intricate way of constructing a 2-d set where both dimensions are equal to 0.  A comment would be appreciated.

I would also prefer the syntax we use for individual affs (something along the lines of `aff_set(isl::multi_aff::identity(tileSpace)) == isl::multi_val::zero(tileSpace)`), but we don't have those operator overloads for `multi_*` yet.  Not for this PR.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -101,24 +101,44 @@ isl::union_map partialSchedule(
 }
 
 namespace {
+/*
+ * If "domain" is any filter, then intersect "domain" with that filter.
```
ftynse: If "node" is any filter

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -101,24 +101,44 @@ isl::union_map partialSchedule(
...
+ */
+isl::union_set anyFilter(const ScheduleTree* node, isl::union_set domain) {
+  if (auto filterElem = node->elemAsBase<ScheduleTreeElemFilter>()) {
+    return domain.intersect(filterElem->filter_);
+  }
+  return domain;
+}
+
+/*
+ * If "domain" is a mapping filter, then intersect "domain" with that filter.
```
ftynse: If "node" is a mapping filter

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -101,24 +101,44 @@ isl::union_map partialSchedule(
 }
 
 namespace {
+/*
+ * If "domain" is any filter, then intersect "domain" with that filter.
+ */
+isl::union_set anyFilter(const ScheduleTree* node, isl::union_set domain) {
```
ftynse: How about calling this `applyAnyFilter` and swapping the order of arguments?  Makes the operation clear.  Maybe mark as `inline` as well.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -101,24 +101,44 @@ isl::union_map partialSchedule(
...
+  if (auto filterElem = node->elemAsBase<ScheduleTreeElemFilter>()) {
+    return domain.intersect(filterElem->filter_);
+  }
+  return domain;
+}
+
+/*
+ * If "domain" is a mapping filter, then intersect "domain" with that filter.
+ */
+isl::union_set mappingFilter(const ScheduleTree* node, isl::union_set domain) {
```
ftynse: Same as above for the naming, otherwise there is a risk of confusing this with the matcher function.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -101,24 +101,44 @@ isl::union_map partialSchedule(
...
+}
+
 // Get the set of domain elements that are active below
-// the given branch of nodes.
+// the given branch of nodes, filtered using "filter".
 //
-// Domain elements are introduced by the root domain node.  Filter nodes
-// disable the points that do not intersect with the filter.  Extension nodes
+// Domain elements are introduced by the root domain node.  Some nodes
+// refine these elements based on "filter".  Extension nodes
```
ftynse: Not sure "refine" makes it clear that there is intersection taking place.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -101,24 +101,44 @@ isl::union_map partialSchedule(
...
 // are considered to introduce additional domain points.
-isl::union_set activeDomainPointsHelper(
+isl::union_set collectDomain(
     const ScheduleTree* root,
-    const vector<const ScheduleTree*>& nodes) {
+    const vector<const ScheduleTree*>& nodes,
+    isl::union_set (*filter)(const ScheduleTree* node, isl::union_set domain)) {
```
ftynse: `std::function<isl::union_set (const ScheduleTree*, isl::union_set)> filter`

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -128,8 +148,23 @@ isl::union_set activeDomainPointsHelper(
   }
   return domain;
 }
+
+// Get the set of domain elements that are active below
+// the given branch of nodes.
+isl::union_set activeDomainPointsHelper(
```
ftynse: Just change `activeDomainPoints` instead of introducing an extra one-line function?  Especially because it looks like you renamed the original `activeDomainPointsHelper`

ftynse:
> - scoped accesses are relative to the outer schedule dimensions,
>   so they do not need to specifically take into account
>   any outer mapping

Unless the promotion scope is above the respective mapping, but that can be taken into account separately through domain intersection.

tc/core/polyhedral/schedule_transforms.cc
```diff
@@ -102,10 +102,10 @@ isl::union_map partialSchedule(
 
 namespace {
 /*
- * If "domain" is any filter, then intersect "domain" with that filter.
+ * If "domain" is a filter, then intersect "domain" with that filter.
  */
-isl::union_set anyFilter(const ScheduleTree* node, isl::union_set domain) {
-  if (auto filterElem = node->elemAsBase<ScheduleTreeElemFilter>()) {
+isl::union_set filter(const ScheduleTree* node, isl::union_set domain) {
```
ftynse: Same naming comment as in some previous commit (`applyFilter`) to avoid confusion with both the tree matcher and  `functional::filter`

skimo-openhub:
On Fri, Jun 15, 2018 at 04:11:27AM -0700, ftynse wrote:
> 
> 19504509d8c861d345687b604f4da99b82f7903d:
> I would not call the modified function `approximateFootprint` then because I expect the footprint to be a set of accessed elements.  I suppose things like `joinOverlappingWrites` that rely on footprint intersection still work because we only need to join groups within one iteration of the scope loops.

ScopedFootprint is also called a "footprint" and also lives in a map space.
I don't see why this is any different.

>      USING_MAPPING_SHORT_NAMES(BX, BY, BZ, TX, TY, TZ);
> -    isl::space blockSpace = isl::space(ctx, 0);
> -    isl::set blockZero =
> -        isl::makeSpecializationSet<int>(blockSpace, {{BX, 0}, {BY, 0}});
> +    isl::space tileSpace = isl::space(ctx, 0).unnamed_set_from_params(2);
> +    auto tileZero =
> +        isl::map(isl::multi_aff::zero(tileSpace.from_range())).range();
> 
> This looks like a very intricate way of constructing a 2-d set where both dimensions are equal to 0.  A comment would be appreciated.

// Work around missing isl_set_from_multi_aff.

skimo

skimo-openhub:
On Fri, Jun 15, 2018 at 04:25:20AM -0700, ftynse wrote:
> +// Get the set of domain elements that are active below
> +// the given branch of nodes.
> +isl::union_set activeDomainPointsHelper(
> 
> Just change `activeDomainPoints` instead of introducing an extra one-line function?  Especially because it looks like you renamed the original `activeDomainPointsHelper`

activeDomainPointsHelper has two callers.
I split up the commit to clarify what is going on.

skimo

skimo-openhub:
On Fri, Jun 15, 2018 at 04:28:31AM -0700, ftynse wrote:
> > - scoped accesses are relative to the outer schedule dimensions,
> >   so they do not need to specifically take into account
> >   any outer mapping
> 
> Unless the promotion scope is above the respective mapping, but that can be taken into account separately through domain intersection.

In that case, the prefix schedule wouldn't include those mappings either,
so its not affected by the change in this commit.

skimo

skimo-openhub:
On Fri, Jun 15, 2018 at 04:36:42AM -0700, ftynse wrote:
> 
> Looks good in general, a couple of stylistic comments.  Nice clean-up of the ScheduleTree concepts.
> 
> I am pre-approving this directly.  Feel free to address comments and merge (if configuration was changed correctly, you don't need a second approval after pushing into an approved PR).

Are you sure?  After you approved #509, I pushed one more
commit on top of it and then you had to approve it again.
In this case I even have to rebase the whole thing,
so there is even less chance that the approval will stick.

skimo

ftynse:
On 15/06/2018 14:23, skimo-openhub wrote:
> Are you sure? After you approved #509, I pushed one more
> commit on top of it and then you had to approve it again.
> In this case I even have to rebase the whole thing,
> so there is even less chance that the approval will stick.

It may have been before the sticky approval was actually enabled (some 
time Wednesday).

PR #520, opened 2018-06-15
use polyhedral reduction description to detect reductions

### LINKED ISSUE #493: polyhedral representation of reductions
The current representation is based on a list of reduction dimensions, which is

- more verbose
- less generic
- incorrect (in theory)

The last two points _probably_ don't come into play in TC because of its strict constraints on the schedule, but a polyhedral representation will be easier to reason about.

This will be done on top of the clean-up that requires the changes in
https://github.com/nicolasvasilache/isl/pull/141

skimo-openhub:
The exact notion of what constitutes a "reduction" was
not spelled out very clearly, but in the use case
of mapping the detected reduction to CUB, each thread
needs to write to a single tensor element.
Take this property as the definition of a "reduction".

A necessary and sufficient condition is that
the schedule dimensions outside the band members
that are considered to form the reduction,
determine a unique tensor element that is updated.

The condition used in the original detection mechanism
(that the reduction band members only involve a specific
reduction dimension of the reduction update statement)
is neither sufficient nor necessary.
It is not sufficient because it doesn't force
the schedule dimensions not involving the reduction dimensions
to appear outside of those band members.
It is not clear if such a case could ever arise in practice
in TC, because the coincident band members get placed outermost.
However, this situation could change in the future if other
criteria such as consecutivity get taken into account.

It is not necessary because it is perfectly fine
for such a band member to reference multiple
reduction dimensions or even to involve non-reduction
dimensions, as long as those non-reduction dimensions
also appear in outer schedule band members.
Since no skewing is currently allowed, this
is currently also unlikely to happen.

In summary, what the reduction band members look like
is largely irrelevant.  What is important is that
the outer band members make sure that only a single
tensor element gets accessed.  It just so happens
that the reduction band members are very likely
to have a particular form in that case.

Change the reduction detection mechanism to use
the polyhedral reduction representation.
Note that if a tensor element is involved in multiple
separate reductions (again, unlikely in the current
code base), then the detection will not only ensure
that a single tensor element is involved, but more
specifically, a single reduction on that tensor element.

This new mechanism is also much simpler.

Closes #493


tc/core/halide2isl.h
```diff
@@ -82,10 +83,8 @@ struct ScheduleTreeAndAccesses {
   /// for each leaf node is captured below.
   tc::polyhedral::ScheduleTreeUPtr tree;
 
-  /// Union maps describing the reads and writes done. Uses the ids in
-  /// the schedule tree to denote the containing Stmt, and tags each
-  /// access with a unique reference id of the form __tc_ref_N.
-  isl::union_map reads, writes;
+  /// Information extracted from the bodies of the statements.
```
ftynse: it makes sense to rename the enclosing struct from `ScheduleTreeAndAccesses` to something else, it has outgrown that name

tc/core/polyhedral/body.h
```diff
@@ -0,0 +1,47 @@
...
+#include <iostream>
+
+#include "tc/external/isl.h"
+
+namespace tc {
+namespace polyhedral {
+
+// Information about the bodies of the polyhedral statements.
+struct Body {
+  Body() = default;
```
ftynse: I'm not entirely sure we want this class to be default-constructible, but it is apparently necessary in the code

tc/core/halide2isl.cc
```diff
@@ -453,8 +453,8 @@ isl::schedule makeScheduleTreeHelper(
     std::tie(newReads, newWrites) =
         extractAccesses(iterationDomain, op, accesses);
 
-    *reads = reads->unite(newReads);
-    *writes = writes->unite(newWrites);
+    body->reads = body->reads.unite(newReads);
```
ftynse: maybe we want a function that takes a union on instances of "Body"?

tc/core/polyhedral/body.h
```diff
@@ -26,19 +26,31 @@ namespace polyhedral {
...
   }
 
   // Union maps describing the reads and writes done. Uses the ids in
   // the schedule tree to denote the containing Stmt, and tags each
   // access with a unique reference id of the form __tc_ref_N.
   isl::union_map reads, writes;
+
+  // A function on reduction update statement instances that partitions them
+  // into individual reductions.
+  // Each reduction involves a single a single tensor element,
```
ftynse: typo "a single a single"

tc/core/polyhedral/body.h
```diff
@@ -26,19 +26,31 @@ namespace polyhedral {
...
+  // meaning that the partition of statement instances based
+  // on the reductions forms a refinement of the partition based
+  // on the element modified by the statement.
```
ftynse: This comment confused me more than helped.  Can we have the spaces spelled out, maybe with an example?

tc/core/polyhedral/schedule_tree_elem.cc
```diff
@@ -175,6 +175,17 @@ void ScheduleTreeElemBand::drop(size_t pos, size_t n) {
   TC_CHECK_EQ(nBegin - n, nMember());
 }
 
+isl::multi_union_pw_aff ScheduleTreeElemBand::memberRange(
+    size_t first,
+    size_t n) const {
+  auto list = mupa_.get_union_pw_aff_list();
+  auto space = addRange(mupa_.get_space().domain(), n);
+  auto end = first + n;
```
ftynse: Can we have an out-of-bounds check here?  I understand isl will complain, but `nMember() - end` is going to be some large value, inconvenient to debug

tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -267,11 +267,11 @@ bool MappedScop::detectReductions(detail::ScheduleTree* tree) {
...
   });
-  // The reduction member needs to appear right underneath
-  // the coincident members.
-  auto reductionDim = nCoincident;
-  auto member = band->mupa_.get_union_pw_aff(reductionDim);
-  if (!isReductionMember(member, updates, scop())) {
+  // The coincident members, together with the prefix schedule,
+  // need to determine a single reduction.
+  auto prefix = prefixScheduleMupa(schedule(), tree);
+  prefix = prefix.range_product(band->memberRange(0, nCoincident));
```
ftynse: This still assumes coincident band members appear before the reduction member, so this sort of contradicts the claim in the commit message of 90820bf66eaa90820a51f22c0f0b635938ef39d1 "It is not sufficient because it doesn't force the schedule dimensions not involving the reduction dimensions to appear outside of those band members."  This does not seem to force or check that property either.

tc/core/polyhedral/schedule_tree_matcher.h
```diff
@@ -29,11 +29,12 @@ namespace polyhedral {
 // that appear in "domain".
 isl::union_set reductionUpdates(isl::union_set domain, const Scop& scop);
 
-// Does the band member with the given partial schedule correspond
-// to a reduction on all statements with a domain in "domain"?
-bool isReductionMember(
-    isl::union_pw_aff member,
+// Does "prefix" partition "domain" into individual reductions?
+// In particular, do the elements of "domain" access a single tensor
```
ftynse: but the code does not actually check something on access relations...

skimo-openhub:
On Wed, Jun 27, 2018 at 09:58:44PM +0000, ftynse wrote:
> 
> This makes sense in general.
> 
> It took me some time to understand what exactly is the representation.  I would appreciate some documentation about it, at least in commit messages.

I updated the documentation on the reductions field.

> The comments often refer to a single tensor element being accessed, but the code does not really operate on access relations, which I find confusing.

This is a consequence of the extraction from Halide:

    The obvious choice for such a reduction space would be
    the space of the tensor involved in the reduction.
    However, there may in theory be multiple reductions
    on the same tensor element separated by some other statement,
    even though this is currently impossible inside TC.
    Still, since the reductions are extracted from Halide,
    it's best to only use information available in the Halide
    representation and to not assume an identity matching
    between reductions and tensor elements.

> There is a relation between the reduction representation and the write access relation, but it is unclear if it always holds.

I updated the comment on extractReduction:

      * The space of the reduction is of the form R_<op->name>_<index>.
      * Each reduction is indexed by the outer loop variables
      * that are not marked as reduction variables.
    + * Since the loop variables that iterate over output tensor elements
    + * are never marked as reduction variables, this means in particular
    + * that all statement instances that belong to the same reduction
    + * write to the same tensor element.
      */
     isl::union_map extractReduction(
	 const IterationDomain& iterationDomain,

> For example, reduction loop can be scheduled the last in some band, but there is a following band whose members appear in the index expressions and thus in the access relations.

It doesn't matter what happens _below_ the potential reduction loop,.
only what happens _above_.  If the outer band members uniquely identify
the accessed tensor elements, then it doesn't matter that inner band
members may also involve loop iterators that index the tensor element
being written.  Now, it's _unlikely_ that these inner band members
would have such a reference, but we shouldn't rely on that.

>  The terms "individual reduction" and "reduction space partitioning" are also not entirely clear in this context.  "Individual reduction" apparently refers to the computation of an individual tensor element of the result, but it would be nice to define this term explicitly.

I updated the documentation on the reductions field.

skimo

tc/core/halide2isl.h
skimo-openhub: On Wed, Jun 27, 2018 at 09:55:01AM -0700, ftynse wrote:
> it makes sense to rename the enclosing struct from `ScheduleTreeAndAccesses` t
o something else, it has outgrown that name

To what?  "halide2isl::Scop"?

skimo

tc/core/halide2isl.cc
skimo-openhub: > maybe we want a function that takes a union on instances of "Body"?

Then we would have to construct a Body from the relations first.
Since C++ doesn't have designated initializers yet, I think that
would be more error-prone because the relations all have the same type,
so you could easily specify them in the wrong order.

I could change that code to

        body->add_reads(newReads);

if you prefer.  Let me know.

skimo

tc/core/polyhedral/body.h
skimo-openhub: > This comment confused me more than helped.  Can we have the spaces spelled out
, maybe with an example?

Updated.

skimo

tc/core/polyhedral/schedule_tree_elem.cc
skimo-openhub: > Can we have an out-of-bounds check here?  I understand isl will complain, but
`nMember() - end` is going to be some large value, inconvenient to debug

Added.

skimo

tc/core/polyhedral/cuda/mapped_scop.cc
skimo-openhub: > This still assumes coincident band members appear before the reduction member,

It only assumes that within a band the coincident members appear
before the non-coincident members.  This is a property of
individual bands in the output of the isl scheduler and
is not directly related to reductions.

> so this sort of contradicts the claim in the commit message of 90820bf66eaa908
20a51f22c0f0b635938ef39d1 "It is not sufficient because it doesn't force the sch
edule dimensions not involving the reduction dimensions to appear outside of tho
se band members."  This does not seem to force or check that property either.

The check happens on the next line:
`if (!isSingleReductionWithin(updates, prefix, scop()))`

tc/core/polyhedral/schedule_tree_matcher.h
skimo-openhub: > but the code does not actually check something on access relations...

This follows from the properties of the reductions, as explained
on the "reductions" field of Body.

skimo

tc/core/polyhedral/body.h
nicolasvasilache: It's prob because `result.body = body;` above.
We should be able to drop the default constructor and just:
 `return ScheduleTreeAndAccesses{body, fromIslSchedule(schedule)};`

skimo-openhub: On Fri, Jun 29, 2018 at 08:52:39AM -0700, Nicolas Vasilache wrote:
> It's prob because `result.body = body;` above.

No, it's the default constructor.  It's needed   
because both `ScheduleTreeAndAccesses` and `Scop`
have default constructors.

> We should be able to drop the default constructor and just:
>  `return ScheduleTreeAndAccesses{body, fromIslSchedule(schedule)};`

`ScheduleTreeAndAccesses` has more than two fields,
so this wouldn't work.

tc/core/polyhedral/schedule_tree_elem.cc
```diff
@@ -175,6 +175,18 @@ void ScheduleTreeElemBand::drop(size_t pos, size_t n) {
...
+  TC_CHECK_LE(end, nMember());
+  list = list.drop(end, nMember() - end);
+  list = list.drop(0, first);
+  return isl::multi_union_pw_aff(space, list);
```
nicolasvasilache: iterator based syntax would be much nicer IMO: 
`return isl::multi_union_pw_aff(space, list.begin() + first, list.begin() + first + n);`

tc/core/halide2isl.cc
```diff
@@ -342,6 +342,64 @@ bool isReductionUpdate(const Provide* op) {
...
+/*
+ * If "op" performs a reduction, then return a mapping from
+ * the statement instances to the individual reductions.
+ * Otherwise, return an empty isl::union_map.
+ *
+ * "op" is considered to be a reduction if it has been marked
+ * as performing a reduction and if more than one statement instance
+ * is involved in the individual reductions.
+ *
+ * The space of the reduction is of the form R_<op->name>_<index>.
```
nicolasvasilache: `is of the form` ->`has a name of the form` ?

tc/core/halide2isl.cc
```diff
@@ -342,6 +342,64 @@ bool isReductionUpdate(const Provide* op) {
...
+  if (finder.reductionVars.size() == 0) {
+    return isl::union_map::empty(iterationDomain.tuple.get_space().params());
+  }
+  auto ctx = iterationDomain.tuple.get_ctx();
+  auto list = isl::aff_list(ctx, 0);
+  for (auto id : iterationDomain.tuple.get_id_list()) {
+    if (finder.reductionVars.count(id.get_name()) == 1) {
+      continue;
+    }
+    auto aff = isl::aff::param_on_domain_space(iterationDomain.paramSpace, id);
```
nicolasvasilache: can we plz get some comments in isl set notation that illustrate what this transformation does?

tc/core/halide2isl.cc
```diff
@@ -342,6 +342,64 @@ bool isReductionUpdate(const Provide* op) {
...
+    aff = aff.unbind_params_insert_domain(iterationDomain.tuple);
```
nicolasvasilache: can we plz get some comments in isl set notation that illustrate what this transformation does?

tc/core/halide2isl.cc
```diff
@@ -342,6 +342,64 @@ bool isReductionUpdate(const Provide* op) {
...
+    list = list.add(aff);
+  }
+  auto domainSpace = iterationDomain.tuple.get_space();
+  isl::id id(ctx, kReductionLabel + op->name + "_" + std::to_string(index));
+  auto space = domainSpace.params().named_set_from_params_id(id, list.size());
+  space = domainSpace.product(space).unwrap();
+  auto reduction = isl::multi_aff(space, list);
+  return isl::union_map(isl::map(reduction));
```
nicolasvasilache: FMI, why immediately convert to union_map, or maybe better, why do we need to go through isl::map?

tc/core/polyhedral/body.h
nicolasvasilache: thanks, isl notation helps a lot

nicolasvasilache:
very nice, thanks!

tc/core/halide2isl.cc
skimo-openhub: On Fri, Jun 29, 2018 at 10:57:17AM -0700, Nicolas Vasilache wrote:
> can we plz get some comments in isl set notation that illustrate what this tra
nsformation does?

Which transformation?

tc/core/halide2isl.cc
skimo-openhub: On Fri, Jun 29, 2018 at 10:57:22AM -0700, Nicolas Vasilache wrote:
> can we plz get some comments in isl set notation that illustrate what this tra
nsformation does?

Are you talking about `aff.unbind_params_insert_domain` here?
I've added some more documentation, but I don't think
it would be a good idea to illustrate every call to isl and
I don't see why this call is any different from any other call.

skimo

tc/core/halide2isl.cc
skimo-openhub: On Fri, Jun 29, 2018 at 11:03:15AM -0700, Nicolas Vasilache wrote:
> FMI, why immediately convert to union_map, or maybe better, why do we need to
go through isl::map?

That's because there is no direct conversion function in isl and
because the C++ bindings do not consider an isl::map
to be a subclass of isl::union_map (the Python bindings do).

skimo

PR #521, opened 2018-06-15
Tunable register promotion depth

### LINKED ISSUE #491: Register promotion: tunable promotion depth
After #490 is implemented, we may use schedule depth to control where the promotion happens independently of the schedule tree structure.

ftynse:
Introduce a tunable parameter for the privateDepth option introduced by
the previous commit.  Choose initial depth within the range [0, 6] since
we rarely have deeper loops.

Closes #491 
Closes #468

### DIFF
--- a/tc/core/cuda/cuda_mapping_options.cc
+++ b/tc/core/cuda/cuda_mapping_options.cc
@@ -289,6 +289,11 @@ CudaMappingOptions& CudaMappingOptions::useReadOnlyCache(bool b) {
   return *this;
 }
 
+CudaMappingOptions& CudaMappingOptions::privateDepth(uint32_t depth) {
+  ownedProto_.set_private_depth(depth);
+  return *this;
+}
+
 CudaMappingOptions& CudaMappingOptions::mapToThreads(
     const std::string& commaSeparatedSizes) {
   auto sizes = parseCommaSeparatedIntegers<uint64_t>(commaSeparatedSizes);
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -602,6 +602,15 @@ void promoteToSharedGreedy(
     scop.insertSyncsAroundCopies(bandNode);
   }
 }
+
+/*
+ * Check if "tree" is a band node mapped to threads.  In particular, check that
+ * "tree" is a band and a thread-specific node appears as its only child.
+ */
+inline bool isThreadMappedBand(const detail::ScheduleTree* tree) {
+  return matchOne(band(threadSpecific(any())), tree) ||
+      matchOne(band(threadSpecific()), tree);
+}
 } // namespace
 
 void promoteGreedilyAtDepth(
@@ -717,6 +726,50 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* scope) {
   auto ancestors = scope->ancestors(root);
   if (functional::Filter(isMappingTo<mapping::ThreadId>, ancestors).empty()) {
     scop.insertSyncsAroundSeqChildren(scope->child({0, 0}));
+  }
+}
+
+/*
+ * Promote to registers below "depth" schedule dimensions.  Split bands if
+ * necessary to create promotion scopes.  Do not promote if it would require
+ * splitting the band mapped to threads as we assume only one band can be
+ * mapped.
+ */
+void promoteToRegistersAtDepth(MappedScop& mscop, size_t depth) {
+  using namespace detail;
+
+  auto root = mscop.scop().scheduleRoot();
+
+  // 1. Collect all bands with a member located at the given depth in the
+  // overall schedule.  Make sure this is the last member of the band by
+  // splitting off the subsequent members into a different band.  Ignore bands
+  // mapped to threads if splitting is required as it would break the invariant
+  // of a single band being mapped to threads in a subtree.
+  // TODO: allow splitting the thread-mapped bands; for example, tile them
+  // explicitly with block size, use the point loops for thread mapping
+  // but ignore them in depth computation.
+  auto bands = bandsContainingScheduleDepth(root, depth);
+  bands = functional::Filter(
+      [root, depth](ScheduleTree* tree) {
+        auto band = tree->elemAs<ScheduleTreeElemBand>();
+        return !isThreadMappedBand(tree) ||
+            tree->scheduleDepth(root) + band->nMember() == depth;
+      },
+      bands);
+  bands = bandsSplitAfterDepth(bands, root, depth);
+
+  // 2. We don't want copies inserted between thread-mapped bands and the
+  // thread-specific marker, but rather below that marker.  If any of the bands
+  // are mapped to threads, take their first children as promotion scope
+  // instead of the band itself.
+  std::function<ScheduleTree*(ScheduleTree*)> findScope =
+      [](ScheduleTree* tree) {
+        return isThreadMappedBand(tree) ? tree->child({0}) : tree;
+      };
+  auto scopes = functional::Map(findScope, bands);
+
+  for (auto scope : scopes) {
+    promoteToRegistersBelow(mscop, scope);
   }
 }
 
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -437,13 +437,19 @@ bool isPromotableToRegistersBelow(
 }
 
 /*
- * Starting from the root, find bands where depth is reached.  Using
+ * Starting from the root, find bands where depth is reached.  If zero depth is
+ * requested, insert a zero-dimensional band node below the root (or the
+ * context node if present) and return it.  Otherwise, use
  * DFSPreorder to make sure order is specified and consistent for tests.
  */
 std::vector<detail::ScheduleTree*> bandsContainingScheduleDepth(
     detail::ScheduleTree* root,
     size_t depth) {
   using namespace tc::polyhedral::detail;
+
+  if (depth == 0) {
+    return {insertTopLevelEmptyBand(root)};
+  }
 
   auto bands =
       ScheduleTree::collectDFSPreorder(root, detail::ScheduleTreeType::Band);
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -779,16 +779,5 @@ void promoteToRegistersAtDepth(MappedScop& mscop, size_t depth) {
   }
 }
 
-// Promote at the positions of the thread specific markers.
-void promoteToRegistersBelowThreads(MappedScop& mscop, size_t nRegisters) {
-  auto& scop = mscop.scop();
-  auto root = scop.scheduleRoot();
-  auto markers = findThreadSpecificMarkers(root);
-
-  for (auto marker : markers) {
-    promoteToRegistersBelow(mscop, marker);
-  }
-}
-
 } // namespace polyhedral
 } // namespace tc
--- a/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
+++ b/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
@@ -698,6 +698,26 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* scope) {
           partialSched);
     }
   }
+
+  // Return immediately if nothing was promoted.
+  if (scope->numChildren() == 0 ||
+      !matchOne(extension(sequence(any())), scope->child({0}))) {
+    return;
+  }
+
+  // If promoting above thread mapping, insert synchronizations.
+  // It is possible that promoted array elements are accessed by different
+  // threads outside the current scope (either in different iterations of the
+  // scope loops, or in sibling subtrees).  For now, always insert
+  // synchronizations, similarly to copies to shared memory.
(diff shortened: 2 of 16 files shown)


ftynse:
@nicolasvasilache could you please run a large-scale tuning on this one?

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -608,22 +612,21 @@ TEST_F(MatMulBias, RegistersBelowFirstBand) {
   // Disable automatic promotion to registers because we are going to call it
   // manually.
   auto mappingOptions = CudaMappingOptions::makeNaiveMappingOptions()
+                            .unroll(512)
```
skimo-openhub: You forgot updating the comment (to explain why unrolling is needed).

skimo-openhub:
f73c231b4a0be962d5f85528e37753e947397f4e ok

tc/core/polyhedral/functional.h
```diff
@@ -119,9 +130,24 @@ std::vector<T> Filter(Func f, const std::vector<T>& input) {
   static_assert(
       function_traits<Func>::n_args == 1,
       "Filtering function must take one argument");
+
+  // Allow combinations:
```
skimo-openhub: Specify which column is which.

tc/core/polyhedral/functional.h
```diff
@@ -119,9 +130,24 @@ std::vector<T> Filter(Func f, const std::vector<T>& input) {
...
+  //       T  and T
+  // const T  and T
+  // const T  and const T
+  //       T* and T*
+  // const T* and T*
+  // const T* and const T*
+  using arg0 = typename function_traits<Func>::template arg<0>::type;
+  constexpr bool sameType = std::is_same<arg0, T>::value;
+  constexpr bool sameTypeRightMbConst =
```
skimo-openhub: What does "Mb" mean?

tc/core/polyhedral/functional.h
```diff
@@ -119,9 +130,24 @@ std::vector<T> Filter(Func f, const std::vector<T>& input) {
...
+      is_same_allow_rhs_non_const<arg0, T>::value;
+  constexpr bool ptrToSameTypeRightMbConst = std::is_pointer<arg0>::value &&
+      std::is_pointer<T>::value &&
+      is_same_allow_rhs_non_const<typename std::remove_pointer<arg0>::type,
+                                  typename std::remove_pointer<T>::type>::value;
   static_assert(
-      std::is_same<typename function_traits<Func>::template arg<0>::type, T>::
-          value,
+      sameType || sameTypeRightMbConst || ptrToSameTypeRightMbConst,
```
skimo-openhub: Update the comment on Filter to explicitly mention that this is allowed.
(It currently only mentions the inverse case.)

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -601,6 +601,15 @@ void promoteToSharedGreedy(
     scop.insertSyncsAroundCopies(bandNode);
   }
 }
+
+/*
+ * Check if "tree" is a band node mapped to threads.  In particular, check that
+ * "tree" is a band and a thread-specific nodes appears as its only child.
```
skimo-openhub: node

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -719,6 +728,50 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* scope) {
...
+ */
+void promoteToRegistersAtDepth(MappedScop& mscop, size_t depth) {
+  using namespace detail;
+
+  auto root = mscop.scop().scheduleRoot();
+
+  // 1. Collect all bands with a member located at the given depth in the
+  // overall schedule.  Make sure this is the last member of the band by
+  // splitting off the subsequent members into a different band.  Ignore bands
+  // mapped to threads if splitting is required as it would break invariant of
```
skimo-openhub: _the_ invariant

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -719,6 +728,50 @@ void promoteToRegistersBelow(MappedScop& mscop, detail::ScheduleTree* scope) {
...
+  // a single band being mapped to threads in a subtree.
```
skimo-openhub: Do we still need this invariant after the mapping has taken place?

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
```diff
@@ -436,14 +436,25 @@ bool isPromotableToRegistersBelow(
...
     size_t depth) {
   using namespace tc::polyhedral::detail;
 
+  if (depth == 0) {
+    auto node = root;
+    if (root->numChildren() > 0 &&
+        root->child({0})->elemAs<detail::ScheduleTreeElemContext>()) {
+      node = root->child({0});
+    }
+    return {insertNodeBelow(node, ScheduleTree::makeEmptyBand(root))};
```
skimo-openhub: This is a bit distracting.  Can we extract some or all of this?
Maybe a `skipContext` or a `insertTopLevelEmptyBand`

tc/core/cuda/cuda_mapping_options.cc
```diff
@@ -320,6 +325,7 @@ CudaMappingOptions CudaMappingOptions::makeUnmappedMappingOptions() {
       .useSharedMemory(false)
       .usePrivateMemory(false)
       .unrollCopyShared(false)
+      .privateDepth(3)
```
skimo-openhub: Why 3?

test/test_cuda_mapper_memory_promotion.cc
```diff
@@ -544,6 +544,7 @@ def fun(float(N,K) A, float(K,M) B, float(N,M) C) -> (O) {
 TEST_F(MatMulBias, RegisterPromotion) {
   auto mappingOptions = CudaMappingOptions::makeNaiveMappingOptions()
                             .tile(32, 32, 32)
+                            .privateDepth(5)
```
skimo-openhub: Why 5?

skimo-openhub:
On Tue, Jun 19, 2018 at 02:11:37AM -0700, skimo-openhub wrote:
> This is a bit distracting.  Can we extract some or all of this?
> Maybe a `skipContext` or a `insertTopLevelEmptyBand`

Here's one I used in an earlier iteration of #525:

diff --git a/tc/core/polyhedral/schedule_transforms.cc b/tc/core/polyhedral/schedule_transforms.cc
index 4490e994..1a351d90 100644
--- a/tc/core/polyhedral/schedule_transforms.cc
+++ b/tc/core/polyhedral/schedule_transforms.cc
@@ -511,6 +511,15 @@ isl::multi_union_pw_aff partialScheduleMupa(
   return band ? prefix.flat_range_product(band->mupa_) : prefix;
 }
 
+ScheduleTree* insertTopLevelEmptyBand(ScheduleTree* root) {
+  auto node = root;
+  if (node->numChildren() > 0 &&
+      node->child({0})->elemAs<detail::ScheduleTreeElemContext>()) {
+    node = node->child({0});
+  }
+  return insertNodeBelow(node, ScheduleTree::makeEmptyBand(root));
+}
+
 void updateTopLevelContext(detail::ScheduleTree* root, isl::set context) {
   if (!matchOne(tc::polyhedral::domain(tc::polyhedral::context(any())), root)) {
     root->appendChild(ScheduleTree::makeContext(
diff --git a/tc/core/polyhedral/schedule_transforms.h b/tc/core/polyhedral/schedule_transforms.h
index 0c264165..213496e7 100644
--- a/tc/core/polyhedral/schedule_transforms.h
+++ b/tc/core/polyhedral/schedule_transforms.h
@@ -114,6 +114,8 @@ detail::ScheduleTree* bandScale(
     detail::ScheduleTree* tree,
     const std::vector<size_t>& scales);
 
+detail::ScheduleTree* insertTopLevelEmptyBand(detail::ScheduleTree* root);
+
 // Update the top-level context node by intersecting it with "context".  The
 // top-level context node must be located directly under the root of the tree.
 // If there is no such node, insert one with universe context first.

tc/core/polyhedral/functional.h
ftynse: "maybe"

ftynse:
> Are you talking about the core computation here?
> It's a bit confusing because the paragraph starts with register promotion.

I don't think we use the "core computation" term in TC.  We found it confusing at some point.  Here, register promotion may be happening in, e.g., one child of a sequence node that has its own thread mapping.  In other children, which we did not analyze because they are out of current scope, promotion to registers could be impossible because of different thread mappings.  Hence the need for synchronization.  The paragraph starts with a register promotion _scope_, and we are still talking about scopes.

> Is that the same as "the mapping strategy" above?
> If so, I'd add the "OuterBlockInnerThread" to the first occurrence
> rather than the second.

Ok, I don't think we have any other strategy now.

> Does TC always add synchronization after a write to an element
> that may get read later? If so, you only need synchronization
> after the write-back and not before reading in.
> This is what PPCG does. (See Section 4.6 of
> "pencil support in pet and PPCG" for an up-to-date description
> of the synchronization in PPCG.)

It does not look like it does, especially with "synchronization strength reduction" that could replace some block synchronizations by warp synchronizations.

tc/core/polyhedral/functional.h
skimo-openhub: On Wed, Jun 20, 2018 at 02:34:14AM -0700, ftynse wrote:
> "maybe"

Hmm... it's probably not that important, but I find it difficult
to read that.  Isn't the point here that the first argument
is the const version of the second one?
How about picking a name that reflects that?
Maybe "sameTypeLeftConst".

skimo

tc/core/polyhedral/cuda/memory_promotion_heuristic.cc
ftynse: Some promotion heuristics used to need the depth of thread x, which has been replaced by finding the thread-specific marker and taking the last member of the band immediately above that.  Shared memory promotion in particular.  I am considering to swap the order of promotions and do registers before shared memory, at which point the invariant must be preserved.  I left a to do to analyze this issue separately from everything else.

skimo-openhub: On Wed, Jun 20, 2018 at 02:57:49AM -0700, ftynse wrote:
> Some promotion heuristics used to need the depth of thread x, which has been replaced by
 finding the thread-specific marker and taking the last member of the band immediately abo
ve that.

In what way does this depend on a single band being mapped to threads?

skimo

ftynse: It may not require that, indeed. I am not going to spend time and check if it does in this PR.

ftynse:
updated

skimo-openhub:
On Wed, Jun 20, 2018 at 03:26:31AM -0700, ftynse wrote:
> > That is, we used to promote to registers at the level of the kernel by default, which seems reasonable. Maybe you disagree on this one? 
> 
> We did not promote at the _kernel_ level, we were promoting below the thread mapping.  Since we map the innermost bands to threads, that promotion was mostly useless: the downstream compiler would have done it anyway.

Sorry.  I miswrote.  I meant the thread mapping level.
I seem to remember register promotion being useful in PPCG,
but maybe the compiler has improved since then.
(We did hoist the promotion to the highest possible level
based on the computed mapping.  I know PPCG maps (the point band of)
the outermost tilable band, but I think in our experiments we
usually didn't have nested tilable bands.)

> > Now we don't have a default other than "3". What does that mean?
> 
> I don't really see how this is more or less reasonable than hardcoding promotion to be below thread-mapped loops.  We expect two loops to be mapped to blocks.  Promotion to shared will happen at depth 2 then.  It makes sense to me to promote to registers deeper than that, but not very deep because it will be useless.  So I chose 3, which is deeper than 2.

Please document this.

> I did consider having promotion depth relative to thread mapping, but there is a situation I did not want to handle.  Consider a schedule tree with a sequence node after a band with 2 members.  The first child has two 2-member bands, with the innermost one mapped to threads.  The second child has only one 2-member band, which mapped to threads.  Relative depth "-4" would make me leave both children and perform promotion _both_ above and below the common ancestor band.  Does not really make sense.  Certainly, we can define a rule to handle such situations, like the deepest common depth, but we want to go in a different direction (band-local decisions in the tuner) anyway, so why bother

Please mention this in the appropriate commit message.

skimo

ftynse:
> Sorry. I miswrote. I meant the thread mapping level. I seem to remember register promotion being useful in PPCG, but maybe the compiler has improved since then. (We did hoist the promotion to the highest possible level based on the computed mapping. I know PPCG maps (the point band of) the outermost tilable band, but I think in our experiments we usually didn't have nested tilable bands.)

We don't have perfectly nested tilable bands, but we do have tilable bands as descendants of sequence nodes.  And we do map those bands to threads (3FCRelu is an example, I think).  We did run the tuner with the current implementation of register promotion, and it has 0 effect on performance in most cases.  In practice, because there was no unrolling check, it essentially promoted the cases like

```c++
A[i][j] = 0;
for (k ...)
  A[i][j] += ...;
```

> Please document this.

I removed that default completely.  Naive options have private memory promotion disabled anyway, so the depth does not matter.

ftynse:
updated again

nicolasvasilache:
Looks good to me except f798b09d489a8830fbf8bee9dd7068798cddca58 which looks like a dangling commit? There is nothing coming after to use `insertTopLevelEmptyBand`.

ftynse:
From the results I managed to get, this gives up to 25% speedups after tuning.

PR #524, opened 2018-06-18
Shut up some -Wsign-compare warnings

ftynse:
I find these annoying, especially the caffe2 stuff, there are hundreds of those.  The default flags for gcc-5.4 have `-Werror -Wno-sign-compare` so these are ignored.


ftynse:
And that makes gcc not emit warnings inside those headers

PR #526, opened 2018-06-19
[Cuda Codegen] Emit launch bounds

thetheodor:
Cuda functions can be annotated with launch bounds, that is the maximum
number of threads per block (the minimum blocks per multiprocessor can
also be specified). This information is used by nvrtc/nvcc during
register allocation (and probably other phases as well).

### DIFF
--- a/tc/core/polyhedral/cuda/codegen.cc
+++ b/tc/core/polyhedral/cuda/codegen.cc
@@ -153,9 +153,17 @@ void emitArgs(stringstream& ss, const Scop& scop) {
 void emitKernelSignature(
     stringstream& ss,
     const std::string& specializedName,
-    const Scop& scop) {
+    const Scop& scop,
+    const Block& block) {
   TC_CHECK_NE(specializedName, "") << "name not provided";
-  ss << "__global__ void " << specializedName << "(";
+  auto b0 = block.view[0];
+  b0 = b0 == 0 ? 1 : b0;
+  auto b1 = block.view[1];
+  b1 = b1 == 0 ? 1 : b1;
+  auto b2 = block.view[2];
+  b1 = b2 == 0 ? 1 : b2;
+  ss << "__global__ __launch_bounds__(" << b0 * b1 * b2 << ") void "
+     << specializedName << "(";
   emitArgs(ss, scop);
   ss << ") {" << endl;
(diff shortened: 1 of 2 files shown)


ftynse:
did you check what happens if somebody manually maps to `.mapToThreads(32,0,0)` ?

ttheodor:
Fixed.

skimo-openhub (changes requested):
I'm just putting a block on this because it fails for me with
```
[ RUN      ] TensorDot_32_512_8_2_28_28.BaseCorrect
unknown file: Failure
C++ exception with description "Error at: /home/skimo/git/c2isl/tc/core/cuda/cuda_rtc.cc:188: CUDA_ERROR_INVALID_VALUE" thrown in the test body.
[  FAILED  ] TensorDot_32_512_8_2_28_28.BaseCorrect (540 ms)
```

skimo-openhub:
On Tue, Jun 19, 2018 at 08:09:18AM -0700, ftynse wrote:
> did you check what happens if somebody manually maps to `.mapToThreads(32,0,0)` ?

Does that make any sense?
Surely, the kernel is not going to run at all in that case,
so why bother with special cases for this situation?

skimo

ttheodor:
Oh, this test is failing for me as well. However, if I dump the cuda and compile it with nvcc, then I see no error.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -153,9 +153,17 @@ void emitArgs(stringstream& ss, const Scop& scop) {
...
+    const Scop& scop,
+    const Block& block) {
   TC_CHECK_NE(specializedName, "") << "name not provided";
-  ss << "__global__ void " << specializedName << "(";
+  auto b0 = block.view[0];
+  b0 = b0 == 0 ? 1 : b0;
+  auto b1 = block.view[1];
+  b1 = b1 == 0 ? 1 : b1;
+  auto b2 = block.view[2];
+  b1 = b2 == 0 ? 1 : b2;
```
skimo-openhub: This should be `b2` instead of `b1`.
However, I would suggest you remove this special handling of 0.

nicolasvasilache (approved):

PR #527, opened 2018-06-19
ScheduleTree::bandTile: drop arbitrary special casing of 0D bands

### LINKED ISSUE #525: Missing mapping filter when the outer band has no child.
When the outer band has no child, there is no mapping filter inserted.
It can be seen in the tc
```
def f(float(N) I) -> (a)
{
    a = 0
}
```
which produce a kernel using all the threads to do the a[0] = 0 instruction.

skimo-openhub:
There appears to be no reason for special casing 0D bands.
On the contrary, the special casing causes confusion in the caller
which expects two nested bands as a result of the tiling.
Remove the special case.

Closes #525.


skimo-openhub:
On Wed, Jun 20, 2018 at 03:58:17AM -0700, ftynse wrote:
> 
> Actually, I think "drop arbitrary special casing of 0D bands" is a wrong description.  The check you drop is `if (tileSizes.size() == 0)`, the number of tile sizes provided.

You're right.  The description was wrong.  I updated the commit message.

> The band itself may or may not be 0-dimensional at this point.  Now I remember the original problem: we wanted to interpret the empty tile size vector as "do not tile" and therefore "do not insert the extra band".

What for?

> Let's check if somebody expects that behavior.

How could we check that?

skimo

ftynse:
On 20/06/2018 13:58, skimo-openhub wrote:
> On Wed, Jun 20, 2018 at 03:58:17AM -0700, ftynse wrote:
>
> > The band itself may or may not be 0-dimensional at this point. Now I 
> remember the original problem: we wanted to interpret the empty tile 
> size vector as "do not tile" and therefore "do not insert the extra band".
>
> What for?

I suppose one of the previous mapping strategies made some assumptions 
about the point loop band being inserted or not.  The one we have now 
assumes the point loop band it always inserted, which makes sense to me.

>
> > Let's check if somebody expects that behavior.
>
> How could we check that?

Put a check that the number of band members is equal to the number of 
tile sizes before returning from the function.  It will fail if somebody 
is using empty tile vector on a non-0D band whereas the function will 
simply return now.

skimo-openhub:
On Wed, Jun 20, 2018 at 12:05:53PM +0000, ftynse wrote:
> On 20/06/2018 13:58, skimo-openhub wrote:
> > On Wed, Jun 20, 2018 at 03:58:17AM -0700, ftynse wrote:
> > > Let's check if somebody expects that behavior.
> >
> > How could we check that?
> 
> Put a check that the number of band members is equal to the number of 
> tile sizes before returning from the function.  It will fail if somebody 
> is using empty tile vector on a non-0D band whereas the function will 
> simply return now.

The test will always pass because the tile size vector
is extended to match the band size.

skimo

ftynse:
On 20/06/2018 14:13, skimo-openhub wrote:
> On Wed, Jun 20, 2018 at 12:05:53PM +0000, ftynse wrote:
> > On 20/06/2018 13:58, skimo-openhub wrote:
> > > On Wed, Jun 20, 2018 at 03:58:17AM -0700, ftynse wrote:
> > > > Let's check if somebody expects that behavior.
> > >
> > > How could we check that?
> >
> > Put a check that the number of band members is equal to the number of
> > tile sizes before returning from the function.  It will fail if 
> somebody
> > is using empty tile vector on a non-0D band whereas the function will
> > simply return now.
>
> The test will always pass because the tile size vector
> is extended to match the band size.

Is it?   I don't see anything like this in `bandTile` before the "if" 
you remove (and that's where I suggest to put the check), and we only 
shrink the tile vector in tileOuterBand.

skimo-openhub:
On Wed, Jun 20, 2018 at 05:17:57AM -0700, ftynse wrote:
> On 20/06/2018 14:13, skimo-openhub wrote:
> > On Wed, Jun 20, 2018 at 12:05:53PM +0000, ftynse wrote:
> > > On 20/06/2018 13:58, skimo-openhub wrote:
> > > > On Wed, Jun 20, 2018 at 03:58:17AM -0700, ftynse wrote:
> > > > > Let's check if somebody expects that behavior.
> > > >
> > > > How could we check that?
> > >
> > > Put a check that the number of band members is equal to the number of
> > > tile sizes before returning from the function.  It will fail if 
> > somebody
> > > is using empty tile vector on a non-0D band whereas the function will
> > > simply return now.
> >
> > The test will always pass because the tile size vector
> > is extended to match the band size.
> 
> Is it?   I don't see anything like this in `bandTile` before the "if" 
> you remove (and that's where I suggest to put the check), and we only 
> shrink the tile vector in tileOuterBand.

Ah... I didn't realize you wanted to put it there.
The adjustment happens later.

Anyway, I checked the actual calls to bandTile.
None of them could possibly depend on the original behavior.
I updated the commit message.

skimo

PR #528, opened 2018-06-19
locally generate and store isl C++ interface

skimo-openhub:
The process of building the C++ interface in TC's (cmake based)
copy of isl was considered to be too fragile for mere users of TC.
It was therefore decided to put this C++ interface inside TC's isl.
Furthermore, in order not to disrupt development of TC's isl itself,
the bindings are automatically generated and stored in a separate branch.
This is only a temporary solution since the goal is
to be able to use mainline isl whenever possible and mainline isl
will not include this interface in its git repository.

Put it in TC itself.  While, in general, it's not a good idea
to include generated files, this should be no worse than
the current situation where the generated files are stored
in TC's isl.
The interface now needs to be generated by whoever updates
the isl submodule, but it has to be generated in advance
anyway to be able to test the isl submodule changes,
so this shouldn't be too much of an extra burden.

The header is generated by generate_isl_cpp_h instead of
generate_isl because generate_isl is still defined by
the isl submodule.  This support can be removed from
the isl submodule in a later step.
Similarly, the local copy of the extract_interface
executable is called extract_isl_interface.

The configuration is based on the one in isl mainline,
except that the defines are hard-coded to a specific
version of LLVM/clang.  This is the same as in
TC's copy of isl, except that the latter missed one define
causing it to miss out on the standard include paths.


ftynse:
Leaving this decision to @nicolasvasilache. 

What I'd like to know (and probably @tobig as well) is the indented distribution and use of isl bindings in mainline.

PR #529, opened 2018-06-20
Move tc/core/libraries.h to tc/core/cuda/cuda_libraries.h

ftynse:
This was left out of the commit in prehistory that moved CUDA-specific
files to cuda/ subdirectories.


ftynse:
> The duplicate "cuda" in the name looks wonky, but it seems they are all like that. I guess we'll need another rename later.

Yes, I don't like that either, just made it consistent with other files.

> I reformatted the commit for you.

Thanks!

Issue #196, opened 2018-03-23
untangle compilation/tuning/options relation in the documentation


ftynse:
Documentation, in particular the [tutorial](https://facebookresearch.github.io/TensorComprehensions/tutorials/tutorial_tensordot_with_tc.html), sort of hides the relation between compilation and tuning.  This has led to many misconceptions about how TC actually works:
- compilation != autotuning, the latter runs the former thousands of times
- compilation errors mean incorrect syntax and the TC expression should be fixed, errors in autotuning mean internal failures and should be reported (but do not necessarily break the further execution)
- mapping options passed to the autotuner are only a starting point in the search space, unlike those passed to compilation
- autotuning is non-deterministic by nature, different runs produce different results _independently_ of the starting point.

ftynse:
cc @prigoyal

prigoyal:
thanks @ftynse, great point. Would you like to take a stab at correcting the docs to make things more verbose and clear since you have been seeing these misconceptions from users? :) assigning to you tentatively.

ftynse:
tutorial is dead

PR #531, opened 2018-06-21
User-facing python and PyTorch integration

nicolasvasilache:
User-facing Python API

This PR reimplements the user-facing Python API and address numerous issues:
1. significant overhead on the python side, due in particular to caching with 
   python string keys based on input tensor sizes;
2. significant overhead on the tensor allocation side because of multiple
    transformations on the ATen to TensorInfo path + range inferencing every time;
3. implicit behaviors everywhere which gave a distorted view of how to use TC
    properly (see issue #196). This is also contrary to PEP-20;
4. 1 cache file per TC and size combination which makes it very cumbersome
    to use caching to files;
5. every python side API call being implemented in terms of
    `TCFunction.apply(self.cu, tc_info, kwargs, *inputs)` which puts the PyTorch
    python autograd abstraction on the critical path to everything. This is not
    portable or even future-proof since with PyTorch 0.5 we want to integrate TC
    at the C++ level after measuring 100+us for just getting into a torch.backward;
6. benchmark overhead of the abstraction which was quite more expensive than
    the CUDA kernels themselves in low latency mode;
7. `tc.define` abstraction only allows 1 TC function for `forward` and 1 TC
    function for `backward` and has a very clunky interface with a function to map
    tensors to the inputs of `backward`

After design discussions with @apaszke it became quickly clear that
using python language features was a significantly better way to go.
This rewrite also kills overhead everywhere and removes as much implicit
behavior as possible.

A little implicit behavior is still remaining in the form of how MappingOptions
are generated in TC class function calls on the first compilation (i.e. similar
behavior to cudnn benchmark except here we JIT-compile).
To make that behavior as explicit as possible we require the user to pass a
generating function which returns a MappingOptions object.

We provide 3 such functions:
1. always naive options for debug mode
2. load from cache or fail for using with a compilation cache file
3. autotune backed by cache with the same API as the `tc.autotune` function

In the future we can add more behaviors depending on user feedback.
For now it seems safer to let the user use whatever preferred behavior.


skimo-openhub:
On Wed, Jun 20, 2018 at 09:20:21PM -0700, Nicolas Vasilache wrote:
> @ftynse was saying:
> I don't like this "implicit compilation" thing. One, it introduces some state that is hard to track:
> 
> ```
> x = tc.define(mm, "matmul", OptionsA)
> x.compile(OptionsB, Y, Z)
> x(Y, Z)  # do we use options A or B?
> x.compile(OptionsC, Y, Z)  # will this work or fail?
> x(Y, Z)  # which options do we use here?
> x.compile(OptionsC, Z, Z)  # same options, different size
> x(Y, Z)  # did "x" adapt to new sizes while also working for old? probably not, but reasonable assumption
> ```

It's not clear to me if the above is the current proposal
or a counter-proposal without implicit compiling, but I would
think that `compile` should return a different kind of object
(that is also runnable).

So,

    x = tc.define(mm, "matmul", OptionsA)
    y = x.compile(OptionsB, Y, Z)
    x(Y, Z)  # compile and use A
    y(Y, Z)  # use B
    z = x.compile(OptionsC, Y, Z)  # this works
    y.compile(OptionsC, Y, Z)  # this fails

skimo

nicolasvasilache:
@skimo-openhub I can make easily compile for an object of the TC class return the underlying executor directly (or a thin wrapper around that so the __call__ has the same interface to both).

python/tests/test_tc.py
```diff
@@ -0,0 +1,420 @@
...
+    #
+    # Simple TC example with explicit 'naive' compilation
+    #
+    def test_tc(self):
+        add = tc.define(
+            "def add(float(N) A, float(N) B) -> (C) { C(i) = A(i) + B(i) }",
+            "add")
+        A, B = torch.randn(100, device='cuda'), torch.randn(100, device='cuda')
+        add.compile(A, B, mapping_options='naive')
+        C, = add(A, B)
```
apaszke: Something we also did in PyTorch is to automatically treat single returns as values, and multi returns as tuples. This is more consistent with Python (`return x` is of `type(x)`, `return x, y` is a `tuple`).

python/tests/test_tc.py
```diff
@@ -0,0 +1,420 @@
...
+            tc.assert_almost_equal(C - torch.add(A, B), (A, B))
+
+            # Sanity check on cache contents
+            cache = tc.MappingOptionsCache(cache_file.name)
+            loaded_options, = cache.load(
+                "def add(float(N) A, float(N) B) -> (C) { C(i) = A(i) + B(i) }",
+                "add",
+                (A, B),
+                10)
+            assert best_options.__str__() == loaded_options.__str__(), (
```
apaszke: nit: `str(x)` instead of `x.__str__()`

python/tests/test_tc.py
```diff
@@ -0,0 +1,420 @@
...
+            conv,
+            "convolution_grad",
+            tuner_config=tuner_config)
+        I, W = (
+            torch.randn(N, C, H, W, device='cuda', requires_grad=True),
+            torch.randn(O, C, kH, kW, device='cuda', requires_grad=True))
+
+        convolution_function = tc.define_with_autograd(
+            lambda I, W: convolution(I, W),
+            lambda I, W, d_O: convolution_grad(I, W, d_O)
```
apaszke: nit: you can eta-reduce those to `tc.define_with_autograd(convolution, convolution_grad)`

python/tests/test_tc.py
```diff
@@ -0,0 +1,434 @@
...
+
+        # First occurrence triggers tuning
+        mean, var = T.moments(I.view((N * G, -1)))
+        out, = T.group_normalization(
+            I, gamma, beta, mean.view((N, G)), var.view((N, G)))
+
+        # Subsequent occurrences do not
+        # But we can still choose to deactivate tuning and use naive options if
+        # ever the need arises.
+        T.generate_implicit_mapping_options = TC.generate_naive_options
```
ftynse: I don't like this part.   It rummages in some state internal to `T` in an inconsistent way (`.define` uses the result of `TC.generate_naive_options()` _call_, which is presumably a function as well; here, you just assign `generate_naive_options` to something without calling it; other examples use `TC.generate_options_from_tuning(...)` that _requires_ parameters which you cannot pass here because you are not calling the function).

python/tests/test_tc.py
```diff
@@ -0,0 +1,434 @@
...
+##############################################################################
+import unittest
+
+import tempfile
+
+import torch
+import torch.cuda
+
+import tensor_comprehensions as tc
+from tensor_comprehensions import TC
```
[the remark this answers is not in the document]
nicolasvasilache: I'll probably end up lifting the generators from TC somewhere else

python/tests/test_tc.py
```diff
@@ -0,0 +1,434 @@
...
+
+    #
+    # Simple TC example with explicit 'naive' compilation
+    #
+    def test_tc(self):
+        A, B = torch.randn(100, device='cuda'), torch.randn(100, device='cuda')
+        add = tc.compile(
+            "def add(float(N) A, float(N) B) -> (C) { C(i) = A(i) + B(i) }",
+            "add",
+            'naive',
```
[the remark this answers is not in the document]
nicolasvasilache: yes you can, @apaszke suggested we want to also have a string to avoid users typing tc.MAppingOptions('naive') everywhere

python/tests/test_tc.py
```diff
@@ -0,0 +1,434 @@
...
+    #
+    def test_define(self):
+        T = tc.define(
+            """
+            def add(float(N) A, float(N) B) -> (C) { C(i) = A(i) + B(i) }
+            def sub(float(N) A, float(N) B) -> (C) { C(i) = A(i) - B(i) }
+            """,
+            TC.generate_naive_options())
+        A, B = torch.randn(100, device='cuda'), torch.randn(100, device='cuda')
+        C, = T.add(A, B)
```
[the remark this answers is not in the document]
nicolasvasilache: Compilation always happens exactly once and is loud unless silent mode is on.
Tuning does not happen here because `TC.generate_naive_options()`

python/tests/test_tc.py
nicolasvasilache: it's unnecessary I'll kill

python/tests/test_tc.py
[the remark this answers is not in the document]
ftynse: so to understand what happens here one needs to look into two other, potentially distant, places...

nicolasvasilache:
> Ultimately, we would like to separate tuning from compilation from execution to avoid
> misinterpretation of time measurements. At least that was one of the major concerns 
> for changing the bindings.

When using the lower level API there is no such risk since you call compile and tune explicitly. 
When using the higher-level tc.define construction there is such risk however forcing to pass an explicit generator for mapping options makes it clear what the behavior is. If your generator calls tuning then you tune loudly, otherwise you compile loudly with whatever options the generator returns.

Note that it is generally common to have the first iteration of each run be quite slower (caching allocator not yet set up, cudnn.benchmark not yet run on kernels etc) so this shouldn't be confusing as long as it is explicit (it now is) and it can be turned off (I'll add a `set_mapping_options_generator` method to TC). @apaszke how does this sound

> Without thinking of practicality, we can achieve this behavior by somehow collecting the sizes 
> relevant to a network without running it, in a sort of "dry run" mode. Unless sizes depend on some > intermediate results of computation, this should be feasible, but probably requires work far out of 
> TC's scope.

I am not sure this is generally a reasonable assumption. However as the primary user for now it is significantly more cumbersome to manually compile and tune

python/tests/test_tc.py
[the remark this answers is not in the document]
nicolasvasilache: Or 1 single place: the documentation that I will add next.
And we'll live with that until someone proposes actual code that significantly improves things.

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+        self.tc = tc
+        self.mapping_options_factory = mapping_options_factory
+        self.compilation_cache = CompilationCache(self.tc)
+        # Make each TC def in the tc str a method of the TC object so we can:
+        #     T = tc.define("def add() ...")
+        #     T.add()
+        #
+        def make_closure(obj: TC, tc_def_name: str):
+            def fun(*inputs: torch.Tensor,
+                    outputs: Optional[Tuple[torch.Tensor]] = (),
```
ftynse: isn't `None` a safer default choice here?

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+                    unchecked: Optional[bool] = False) -> List[torch.Tensor] :
+                return obj.__call__(
```
ftynse: just use `obj(...)` instead of `__call__`

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+        def __implicit_compile__(tc_obj: TC,
+                                 entry_point: str,
+                                 *inputs: torch.Tensor):
+            already_compiled = tc_obj.compilation_cache.is_compiled(
+                entry_point, inputs)
+
+            if already_compiled:
+                return
+
+            if not SILENT:
```
ftynse: you probably want `global SILENT` above this line

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+
+        Returns
+            List[torch.Tensor]: the tensors returned by the TC. If outputs are
+                not specified then new allocations take place. Otherwise the TC
+                writes in-place and returns the same tensors it received as
+                outputs.
+        """
+
+        # Locally scoped implicit compilation
+        def __implicit_compile__(tc_obj: TC,
```
ftynse: you don't need the `__name__` syntax here, no name mangling happens at the function scope

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+            #     (entry point, inputs)
+            start = time.clock()
+            tc_obj.compilation_cache.compile(
+                entry_point, inputs, mapping_options)
+            if not SILENT:
+                print(
+                    "Done compiling TC \"{}\" (compile time: {}ms)".format(
+                        entry_point, int((time.clock() - start) * 10 ** 3)))
+
+        __implicit_compile__(self, entry_point, *inputs)
```
ftynse: Why define a function in a function scope and only call it once? I'd just paste the code and add the comments

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+                        i.size().__str__()) for i in inputs)) +
+                    "....Generate implicit MappingOptions")
+
+            mapping_options = tc_obj.mapping_options_factory(
+                tc_obj.tc, entry_point, *inputs)
+
+            assert mapping_options is not None, (
+                "No options found for TC {} ".format(entry_point) +
+                "with inputs of sizes:\n  {}\n".format(
+                    "".join("{} ".format(i.size().__str__()) for i in inputs)))
```
ftynse: `str(i.size())`;  you almost never want to _call_ the magic functions (with `__syntax__`)

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+        __implicit_compile__(self, entry_point, *inputs)
+
+        inputs = tuple(inputs) if len(inputs) > 1 else tuple(inputs, )
+        if unchecked:
+            return self.compilation_cache.unchecked_run(
+                entry_point, inputs, outputs)
+
+        return self.compilation_cache.run(entry_point, inputs, outputs)
+
+    def set_mapping_options_factory(self, mapping_options_factory):
```
ftynse: Is this setter necessary?  The field is not name-mangled anyway, and there is no additional type hinting here.  I remember commenting on the field assignment in the example code, but I was questioning the idea of changing the factory on an already-constructed object.  You can end up in a state where you have compiled some kernels with naive options (because originally using naive factory), have changed the factory to tuning, but no tuning happens for already compiled kernels.  This sounds too complex to explain, so it's probably a bad idea

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+
+class TCWithAutograd(object):
+    def __init__(self, forward_fun, backward_fun):
+        self.forward_fun = forward_fun
+        self.backward_fun = backward_fun
+
+    def __call__(self, *inputs):
+        return TcFunction.apply(self.forward_fun, self.backward_fun, *inputs)
+
+def define_with_autograd(forward_fun, backward_fun):
```
ftynse: the naming implies there is some relation with `define`, but there is very little here

tensor_comprehensions/__init__.py
```diff
@@ -12,20 +12,372 @@
...
+
+    base_options = (
+        MappingOptions(starting_options)
+        if isinstance(starting_options, str) else starting_options)
+    if base_options is None and load_from_cache:
+        cache = MappingOptionsCache(cache_filename)
+        loaded = cache.load(tc, entry_point, inputs, 1)
+        if len(loaded) > 0:
+            base_options = loaded[0]
+    if base_options is None:
```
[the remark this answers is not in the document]
nicolasvasilache: ok failling hard, we'll see when users start complaining

tensor_comprehensions/__init__.py
nicolasvasilache: makes it as explicit and structured as possible that this is an implicit behavior

tensor_comprehensions/__init__.py
nicolasvasilache: fine let users complain and decide then, I can live without it

nicolasvasilache:
> 579dcc4 So with define, we parse twice now...

memoizing this is premature optimization

nicolasvasilache:
> 921e1ca why do we need to rename those functions?

shorter names are better, consistency with the C++ options (we don't do `--set_dump_cuda=true`), `set_xxx` also leaks that there is something to "set" in the underlying impl.

nicolasvasilache:
> I don't see why do you insist on having all arguments in the constructor. It's possible to have the 
> same fluent-style interface (aka chained calls) as we do for mapping options. I would say it is 
> actually preferable to have a consistent API.

sure that's an easy change

> TunerConfig is never used as a python context manager, the __enter__ and __exit__ functions are 
> not even exported to python (and the argument list for __exit__ are wrong). So let's not pretend like 
> it is and have a proper RAII in C++ (not TunerConfig because it's constructed way too early).

You can't have both RAII and avoid list of parameters. If you construct an RAII context object on the C++ side, you need to pass the arguments to some function which makes your point just above moot.
It seems whatever the solution we go for we need an extra set of variables that mirror the flags. You may choose to not group them together in an object and/or use these new variables everywhere and keep the flags readonly but that's just displacing the problem. There is no fundamental difference except renaming things. Feel free to pick this up in another PR if you see a much better way of doing this. I'll just change to chaining to keep consistency.

ftynse:
> You can't have both RAII and avoid list of parameters. If you construct an RAII context object on the C++ side, you need to pass the arguments to some function which makes your point just above moot.

That's why I said "a proper RAII in C++ (_not_ TunerConfig <...>)"  `TunerConfig` is essentially a value type.  If you want a RAII on top, introduce `WithTunerConfig` constructible from `TunerConfig`.  A clean solution would instead get rid of the global variables and pass the config to the tuning call instead.  This removes the need for RAII.  Not in this PR indeed.

However, let's not use python-like names for unexported functions to avoid confusion and drop the reference to context manager.

concretevitamin:
+1 looking forward to this.  I'm at a commit from a few weeks ago and benchmarking from Python side after autotuning seems to trigger recompilation every single time.  This would be a fantastic change to really enable PyTorch users to benefit from TC.

nicolasvasilache:
@concretevitamin using the pybind API directly (like I do [here](https://github.com/facebookresearch/TensorComprehensions/blob/master/python/examples/tc_pybind_example.py#L89)) allows you to bypass the issue until #531 lands.

nicolasvasilache:
@caffe2bot retest this please

nicolasvasilache:
CI was green before rebase taking a small leap of faith

PR #535, opened 2018-06-22
check that all accesses to registers have constant index expressions

### LINKED ISSUE #517: Check that tensors promoted to registers are only accessed with constant index expressions
This could be done by checking the AST inside `emitCudaKernel`.

skimo-openhub:
The index expressions of accesses to arrays of registers should
be constant since registers are not addressable.
Failing to use constant index expressions would result
in accesses to "local" memory, which is much slower than registers.
Some MatMulBias tests checked for particular patterns that should
not appear when the index expressions are constant, but they did
not check that they are actually constants.
Replace these checks by generic checks during emitCudaKernel
that explicitly check for constant index expressions and
that get applied to all accesses to registers.

Closes #517

### DIFF
--- a/tc/core/polyhedral/cuda/codegen.cc
+++ b/tc/core/polyhedral/cuda/codegen.cc
@@ -410,10 +410,33 @@ void emitAccess(AFF access, const CodegenStatementContext& context) {
   emitAccess(buildAccess(access, context), context);
 }
 
+// Check that the given expression is an access with constant index expressions
+void checkConstantAccess(isl::ast_expr expr) {
+  auto op = expr.as<isl::ast_expr_op>();
+  auto access = op.as<isl::ast_op_access>();
+  TC_CHECK(access);
+  for (int i = 1; i < access.get_n_arg(); ++i) {
+    auto arg = access.get_arg(i);
+    TC_CHECK(arg.as<isl::ast_expr_int>())
+        << "expected constant subscript, got " << arg.to_C_str();
+  }
+}
+
+// Print an access to a(n array of) register(s), checking that
+// the index expressions are constant.
+void emitRegisterAccess(
+    isl::pw_multi_aff access,
+    const CodegenStatementContext& context) {
+  auto expr = buildAccess(access, context);
+  checkConstantAccess(expr);
+  emitAccess(expr, context);
+}
+
 // Print an access to global memory, wrapping the access in an "__ldg()"
 // call if the accessed tensor is known to be read-only.
-template <typename AFF>
-void emitGlobalAccess(AFF access, const CodegenStatementContext& context) {
+void emitGlobalAccess(
+    isl::multi_pw_aff access,
+    const CodegenStatementContext& context) {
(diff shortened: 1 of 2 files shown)


tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -396,10 +396,46 @@ struct LdgWrapper {
...
+}
+
+// Check that the given expression is an access with constant index expressions
+void checkConstantAccess(isl::ast_expr expr) {
+  auto op = expr.as<isl::ast_expr_op>();
+  auto access = op.as<isl::ast_op_access>();
+  TC_CHECK(access);
+  for (int i = 1; i < access.get_n_arg(); ++i) {
+    auto arg = access.get_arg(i);
+    TC_CHECK(arg.as<isl::ast_expr_int>());
```
ftynse: Let's have a description here saying something like "expected constant subscripts, got <..>".  Not sure if AST nodes can be simply converted to strings, no problem if not.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -396,10 +396,46 @@ struct LdgWrapper {
...
+  }
+}
+
+// Print an access to a(n array of) register(s), checking that
+// the index expressions are constant.
+template <typename AFF>
```
ftynse: Why do we need a template here?  As far as I can see, there is only one caller of this function so there is no need to support different types.

tc/core/polyhedral/cuda/codegen.cc
```diff
@@ -396,10 +396,46 @@ struct LdgWrapper {
...
+void emitRegisterAccess(AFF access, const CodegenStatementContext& context) {
+  auto expr = buildAccess(access, context);
+  checkConstantAccess(expr);
+  emitAccess(expr, context);
+}
+
+// Print an access to global memory, wrapping the access in an "__ldg()"
+// call if the accessed tensor is known to be read-only.
+template <typename AFF>
+void emitGlobalAccess(AFF access, const CodegenStatementContext& context) {
```
ftynse: Same question for the template: there are two call sites using the same argument type.

skimo-openhub:
On Thu, Jun 28, 2018 at 01:49:26AM -0700, ftynse wrote:
> +// Check that the given expression is an access with constant index expressions
> +void checkConstantAccess(isl::ast_expr expr) {
> +  auto op = expr.as<isl::ast_expr_op>();
> +  auto access = op.as<isl::ast_op_access>();
> +  TC_CHECK(access);
> +  for (int i = 1; i < access.get_n_arg(); ++i) {
> +    auto arg = access.get_arg(i);
> +    TC_CHECK(arg.as<isl::ast_expr_int>());
> 
> Let's have a description here saying something like "expected constant subscripts, got <..>".  Not sure if AST nodes can be simply converted to strings, no problem if not.

Added.

> +// Print an access to a(n array of) register(s), checking that
> +// the index expressions are constant.
> +template <typename AFF>
> 
> Why do we need a template here?  As far as I can see, there is only one caller of this function so there is no need to support different types.

*shrug*
I removed them.

skimo

PR #541, opened 2018-06-29
Suppress warnings

### LINKED ISSUE #456: Add a flag to supress warnings
Currently, TC will emit warnings about preconditions not being checked and non-initialized reductions.  When developing TC kernels, these are useful.  But when autotuning, these are only annoying as they appear in each compilation.  Add a `-w` flag that suppresses warning output.

ftynse:
Currently, warnings are always emitted during TC parsing (including
bounds inference) and semantic analysis.  This leads to megabytes of
identical warning spew in the autotuning, among other problems.
Introduce a boolean flag, "--warn", to control whether warnings should
be printed.  Set it to true by default.

This required to introduce the gflags support into the tc_lang library:
previously, flags were concentrated in the tc_core library, but tc_core
links to tc_lang making it impossible to put lang-specific flags in
core.  As a consequence, tc_lang now links gflags statically and
tc_core does not.  This is not a problem since tc_core links tc_lang
anyway.

Introducing even more gflags is not the best solution, since they are essentially global variables, but it is a pragmatic one.  We want to consider introducing some `CompilerOptions` structure that would include `MappingOptions` and other configurations (for example, kernel timeout), but I expect the discussion to take long.

Closes #456

### DIFF
--- a/tc/autotuner/autotuner-inl.h
+++ b/tc/autotuner/autotuner-inl.h
@@ -394,6 +394,9 @@ Autotuner<Backend, SearchStrategy>::tune(
   TC_CHECK_EQ(tcEntryPointMap.count(tcEntryPoint), 1u)
       << "Error looking up " << tcEntryPoint;
 
+  // Do not emit lang warnings during the tuning run
+  auto enableWarnings = lang::EnableWarnings(false);
+
   // Initialize a model configuration
   TuningConfiguration modelConfiguration;
   TC_CHECK_GE(inputs.size(), 1u);
--- a/tc/autotuner/utils.cc
+++ b/tc/autotuner/utils.cc
@@ -19,6 +19,7 @@
 #include <cmath>
 #include <sstream>
 
+#include <glog/logging.h>
 #include <glog/stl_logging.h>
 
 #include "tc/core/flags.h"
--- a/tc/core/CMakeLists.txt
+++ b/tc/core/CMakeLists.txt
@@ -39,7 +39,6 @@ target_include_directories(tc_core PUBLIC ${LLVM_INCLUDE_DIRS})
 target_link_libraries(
   tc_core
 
-  gflags
   glog
 
   ${HALIDE_LIBRARIES}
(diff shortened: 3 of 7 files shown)


nicolasvasilache:
Looks good, not immediately clear to me why the linking error pops up though

ftynse:
Neither is it to me.  There is no linking problem on my local machine.  I also don't see any dynamic linking of gflags.  It is true that `tc_core` links `tc_lang` dynamically, so `tc_core` transitively links gflags dynamically.

Tests link gflags directly.  I tried removing that on the CI machine, but it did not help.  Maybe some build artifact.

nicolasvasilache:
Another option in a world where everyone is linking more and more things statically (pytorch 1.0 even links cuda libraries dynamically), is to ourselves go the static linking route as I had previously suggested. Are we more open to this now? @skimo-openhub ?

ftynse:
Wasn't there a technical problem with static linking?

skimo-openhub:
On Tue, Jul 03, 2018 at 11:57:56AM -0700, Nicolas Vasilache wrote:
> Another option in a world where everyone is linking more and more things statically (pytorch 1.0 even links cuda libraries dynamically), is to ourselves go the static linking route as I had previously suggested. Are we more open to this now? @skimo-openhub ?

I don't know why everyone is switching to static libraries.
I do think we shouldn't link in static libraries in a dynamic library.
I fixed the gflags issue in this PR.

skimo

ftynse:
> I fixed the gflags issue in this PR.

Thanks!

skimo-openhub:
On Wed, Jul 04, 2018 at 04:52:56AM -0700, skimo-openhub wrote:
> On Tue, Jul 03, 2018 at 11:57:56AM -0700, Nicolas Vasilache wrote:
> > Another option in a world where everyone is linking more and more things statically (pytorch 1.0 even links cuda libraries dynamically), is to ourselves go the static linking route as I had previously suggested. Are we more open to this now? @skimo-openhub ?
> 
> I don't know why everyone is switching to static libraries.
> I do think we shouldn't link in static libraries in a dynamic library.
> I fixed the gflags issue in this PR.

Actually, I made the error go away, but I'm not sure I fixed
the issue completely.
libgflags is still getting linked in several times,
because anything that depends on libtc_lang (including other
shared libraries) now links in libgflags.

skimo

PR #542, opened 2018-06-29
Separate out schedule_utils.{cc,h} from schedule_transforms.{cc,h}

ftynse:
Many of the functions in schedule_transforms.{cc,h} are not
transformations but some higher-level functions that operate on (parts
of) schedule trees rather than on individual nodes.  These functions
were introduced into schedule_transforms.{cc,h} for historical reasons.
The file has grown large.  Separate the non-transformation functions
into separate files, schedule_utils.{cc,h}.

This commit only _moves_ the existing code and updates includes.


skimo-openhub:
On Fri, Jun 29, 2018 at 06:45:34AM -0700, ftynse wrote:
> Many of the functions in schedule_transforms.{cc,h} are not
> transformations but some higher-level functions that operate on (parts
> of) schedule trees rather than on individual nodes.  These functions
> were introduced into schedule_transforms.{cc,h} for historical reasons.
> The file has grown large.  Separate the non-transformation functions
> into separate files, schedule_utils.{cc,h}.
> 
> This commit only _moves_ the existing code and updates includes.
> You can view, comment on, or merge this pull request online at:

Are you sure you only moved code written in 2018?

(github doesn't show the diff properly, so I can't comment
on the lines that introduce 2018)

skimo

tc/core/polyhedral/schedule_utils.cc
```diff
@@ -0,0 +1,318 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
```
skimo-openhub: Are you sure you only moved code written in 2018?

tc/core/polyhedral/schedule_utils.h
```diff
@@ -0,0 +1,141 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
```
skimo-openhub: Are you sure you only moved code written in 2018?

ftynse:
@skimo-openhub you are right, there is some pre-historic code

Issue #300, opened 2018-04-12
Properly evaluate the scratchpad size necessary for CUB reductions


ftynse:
For parallel reductions, CUB requires a block of shared memory, which is wrapped under an opaque object.  Shared memory promotion mechanism needs to know the amount of available shared memory.  Currently, we bluntly subtract 4096 bytes from the amount of shared memory on GPU for each reduction we use.  Some reductions apparently require more space than that, which leads to our kernels requiring more shared memory than physically available.  This manifests as either "invalid ptx" or "launch out of resources" errors.  #299 allows the autotuner to survive these errors, but they may be avoided completely if the size of the reduction scratchpad is known.

The logic for shared memory allocation inside CUB is incredibly complicated due to using different algorithms depending on the available architectural features.  It is implemented as templates making it impossible to extract the size computation into runtime without reimplementing it.  However, maintaining an eventual reimplementation on par with CUB changes seems hard.  One possible solution could be to emit the code (or only the reduction part) without shared memory promotion, compile it into PTX, and parse PTX to collect the required amount of shared memory.  This may create a notable overhead in compilation time though.

nicolasvasilache:
cc vinodgro

PR #543, opened 2018-06-29
ScheduleTree subclasses

ftynse:
```
The existing implementation of ScheduleXYZ inherited from the prehistory
chose to have an abstract tree node type (ScheduleTree) with a unique
pointer to a specific node element (ScheduleTreeElem*) as payload.  It
is unclear now why this structure was chosen, possible reasons include:
the duality between isl_schedule_tree and isl_schedule_node in isl
internals; gradual movement from raw pointers, to shared, to unique
pointers; an attempt to separate tree memory management (unique pointers
to children in ScheduleTree) from data (fields in ScheduleTreeElem*).
However, the resulting structure ended up being unnecessarily complex.
Conceptually, a specific node type _is-a_ tree node, so it makes sense
to have specific nodes inherit from a generic one.  isl C++ API already
provides this structure and it proved to be more convenient.  In
particular, it enables functions to specify that they expect a certain
type of node statically as a part of their signature rather than
dynamically through CHECK macros.

Make ScheduleTreeElem* classes inherit from ScheduleTree instead of
ScheduleTreeElemBase and remove the "elem_" field from ScheduleTree (the
tree class itself can now be casted to a subclass). 
```

This is essentially a cleaup of ScheduleTree that we have been wanting for some time.  There are more things to be done, but let's discuss the first part since it may impact the polyhedral parts.  In particular, the previous implementation was using macro-based static dispatch.  One reason for having it was the decoupling between `ScheduleTree` that stored `type_` and the `ScheduleTreeElemBase` that was dependent on it.  At the same time `ScheduleTreeElemBase` was (mostly) relying on virtual methods for polymorphic behavior.  Now we ended up with both approaches in the same inheritance tree.  I don't like macro dispatch in modern C++.  We should go either for either virtual methods or CRTP.


tc/core/polyhedral/schedule_isl_conversion.cc
```diff
@@ -237,6 +237,60 @@ isl::schedule toIslSchedule(const ScheduleTree* root) {
...
+    return std::unique_ptr<ScheduleTreeElemFilter>(
+        new ScheduleTreeElemFilter(f));
+  } else if (auto guard = node.as<isl::schedule_node_guard>()) {
+    LOG(FATAL) << "guard nodes not supported";
+    return nullptr;
+  } else if (auto mark = node.as<isl::schedule_node_mark>()) {
+    LOG(FATAL) << "mark nodes not supported";
+    return nullptr;
+  } else if (node.isa<isl::schedule_node_leaf>()) {
+    LOG(FATAL) << "ScheduleTreeElemBase::make called on explicit leaf";
```
skimo-openhub: update error message

tc/core/polyhedral/schedule_tree.h
```diff
@@ -453,6 +453,10 @@ struct ScheduleTree {
     return type_;
   }
 
+  virtual std::ostream& write(std::ostream& os) const {
```
skimo-openhub: Can you make it abstract?  (not at this point, but later in the series)
I'd also pick a name that clarifies that this only prints the contents of a node and
node the entire (sub)tree.  At least at a comment to that effect.

tc/core/polyhedral/schedule_tree.h
```diff
@@ -276,8 +276,31 @@ struct ScheduleTree {
...
+        std::is_base_of<mapping::MappingId, MappingIdType>::value,
+        "can only map to ids that are subclasses of MappingId");
+    std::vector<mapping::MappingId> ids;
+    ids.reserve(mappedIds.size());
+    for (const auto& id : mappedIds) {
+      ids.push_back(id);
+    }
+    // The trailing argument is necessary to avoid overload resolution and
+    // potential infinite recursion here.
+    return makeMapping(ids, mappedAffs, std::move(children), 0);
```
skimo-openhub: Ugh.  Why not use a different name instead of adding a dummy argument?

tc/core/polyhedral/schedule_tree.cc
```diff
@@ -279,6 +279,30 @@ std::unique_ptr<ScheduleTree> ScheduleTree::makeFilter(
...
+  TC_CHECK_GE(mapping.size(), 1u) << "empty mapping";
+  TC_CHECK_EQ(mappedIds.size(), mapping.size())
+      << "some id is used more than once in the mapping";
+  auto ctx = mappedIds[0].get_ctx();
+  ScheduleTreeUPtr res(new ScheduleTree(
+      ctx,
+      std::move(children),
+      ScheduleTreeType::Mapping,
+      std::unique_ptr<ScheduleTreeElemMapping>(
+          new ScheduleTreeElemMapping(mapping))));
```
skimo-openhub: This last statement isn't just moved by also rewritten.  Why?
In particular, why do you change it do call a constructor with a children
argument if you later in this series remove this children argument
from the constructor?
Actually, it's also not clear to me why you remove this argument later.

tc/core/polyhedral/schedule_tree.h
```diff
@@ -124,6 +124,12 @@ struct ScheduleTree {
       : ctx_(ctx), type_(type), elem_(std::move(elem)) {
     appendChildren(std::move(children));
   }
+
+  // Copy constructor for internal use only.
+  // Note that this does not copy the underlying elem_.
+  // All callers should use makeScheduleTree(const ScheduleTree&) instead,
+  // which copies the underlying elem_ as well as maintains the memory
+  // management scheme consistent.
```
skimo-openhub: This sounds weird (possibly French?).
Possible alternatives:
"keeps the memory management scheme consistent"
"maintains the memory management scheme"
"maintains a consistent memory management scheme"

tc/core/polyhedral/schedule_tree.h
```diff
@@ -399,23 +417,18 @@ struct ScheduleTree {
...
-    auto ctx = arg->ctx_;
-    std::vector<ScheduleTreeUPtr> children =
-        vectorFromArgs(std::forward<Arg>(arg), std::forward<Args>(args)...);
-
-    auto res = ScheduleTreeUPtr(new ScheduleTree(
-        ctx,
-        std::move(children),
-        type,
-        std::unique_ptr<ScheduleTreeElemBase>(new T)));
+    auto ctx = arg->ctx_; // FIXME: globalctx
```
skimo-openhub: What needs to be fixed here?

tc/core/polyhedral/schedule_tree.cc
```diff
@@ -404,7 +404,6 @@ bool ScheduleTree::operator==(const ScheduleTree& other) const {
   }
   return true;
 }
-
```
skimo-openhub: spurious whitespace change

ftynse:
`ScheduleXYZ` was a code name for the our tree structure

skimo-openhub:
On Tue, Jul 03, 2018 at 01:35:28PM +0000, ftynse wrote:
> `ScheduleXYZ` was a code name for the our tree structure

That must have been something from prehistory.
There is no reference to this term in the history and
there is only a single reference in the code (which should
probably be removed).
Let's use a term that is more likely to be understood
instead of some prehistoric code name.

skimo

tc/core/polyhedral/schedule_tree.h
ftynse: Done

ftynse:
> Let's use a term that is more likely to be understood instead of some prehistoric code name.

I just used ScheduleTree

>  I see you didn't rename elemEquals as well as several variable names ending in "Elem". Not sure if that was on purpose.

`elemEquals` uses macro-based dispatch.  I think the equality comparison behavior is better implemented with a virtual `equals(const ScheduleTree*)` function overridden in each subclass and called from `operator==`.  This removes the need for `elemEquals` completely.  But this dispatch is a different discussion that is yet to be held.

Renaming `elemAs` to `as` was done by grep+sed, getting rid of `Elem` in names is trickier.  Let's do it progressively.

nicolasvasilache:
making a pass on this today, please wait before merging

tc/core/polyhedral/schedule_isl_conversion.cc
```diff
@@ -237,6 +237,60 @@ isl::schedule toIslSchedule(const ScheduleTree* root) {
...
+  auto res =
+      ScheduleTreeElemBand::fromMultiUnionPwAff(b.get_partial_schedule());
+  res->permutable_ = b.get_permutable();
+  for (size_t i = 0; i < b.n_member(); ++i) {
+    res->coincident_[i] = b.member_get_coincident(i);
+  }
+  return res;
+}
+
+std::unique_ptr<ScheduleTreeElemBase> elemFromIslScheduleNode(
```
nicolasvasilache: I find that the `fromXXX` nomenclature is better for static class members while I prefer `makeYYY(const XXX&)` for free functions. I think we have generally been consistent with `std::unique_ptr<YYY> makeYYY(...)` (except apparently for `fromIslScheduleNodeBand` above). 

I would use `makeElemBand` and `makeElem` instead, or something similar.

tc/core/polyhedral/schedule_tree.cc
```diff
@@ -156,6 +156,8 @@ static std::unique_ptr<ScheduleTreeElemBase> makeElem(const ScheduleTree& st) {
 ////////////////////////////////////////////////////////////////////////////////
 ScheduleTree::ScheduleTree(isl::ctx ctx) : ctx_(ctx) {}
 
+ScheduleTree::~ScheduleTree() {}
```
nicolasvasilache: What is the point of having 3 commits for introducing virtual methods?
It creates an intermediate state in which methods have virtual methods but no virtual destructors.
In any case I think the splitting of commits goes a bit too far here, I'd squash the 3 commits in 1.

tc/core/polyhedral/schedule_tree.cc
```diff
@@ -286,7 +286,7 @@ std::unique_ptr<ScheduleTree> ScheduleTree::makeMappingUnsafe(
       << "some id is used more than once in the mapping";
   auto ctx = mappedIds[0].get_ctx();
   auto res =
-      std::unique_ptr<ScheduleTree>(new ScheduleTreeElemMapping(mapping));
+      std::unique_ptr<ScheduleTree>(new ScheduleTreeElemMapping(ctx, mapping));
```
nicolasvasilache: do we need 4 different commits for the same change applied to different types, I'd squash

tc/core/polyhedral/schedule_tree.h
```diff
@@ -126,6 +126,7 @@ namespace detail {
 
 std::ostream& operator<<(std::ostream& os, const ScheduleTree& tree);
 
+template <typename T>
```
nicolasvasilache: while you're modifying this function, shouldn't we make it return the flattened tree for API consistency

nicolasvasilache:
LGTM, thanks for the cleanups

tc/core/polyhedral/schedule_tree.cc
ftynse: Not sure.  The changes are conceptually similar (except for `ThreadSpecificMapping`, which has a different motivation), but completely independent from each other.  What do you think @skimo-openhub ?

tc/core/polyhedral/schedule_tree.h
ftynse: This change is irrelevant to the goal of this PR.  Lately, we have been enforcing the policy of not having irrelevant commits in the PR.  I'm okay with either way, but let's be consistent.

nicolasvasilache: sure let's keep PR consistency first

tc/core/polyhedral/schedule_tree.cc
skimo-openhub: On Tue, Jul 03, 2018 at 11:58:25AM -0700, ftynse wrote:
> Not sure.  The changes are conceptually similar (except for `ThreadSpecificMapping`, which
has a different motivation), but completely independent from each other.  What do you think @
skimo-openhub ?

I think the current split-up is fine.

However, if you really want to squash things, then I'd squash
all four of them into
"ScheduleTreeElem*: inherit from ScheduleTree instead of ScheduleTreeElemBase"
because I don't think it's good idea to "break" things (at least prompting
you to introduce FIXMEs) and then fixing them up later in the same series.

On the other hand, I'm not sure why we are storing the isl::ctx in schedule
trees in the first place.  Other parts of the code are apparently simply
using the global ctx.  I don't know what the motivation was for introducing
this global ctx (it was introduced by a commit saying only
"Trying to rebase #300 properly after github craziness"), so I can't tell 
what makes schedule trees so special that they shouldn't use the global ctx.

skimo

ftynse: > However, if you really want to squash things, then I'd squash
all four of them into
"ScheduleTreeElem*: inherit from ScheduleTree instead of ScheduleTreeElemBase"
because I don't think it's good idea to "break" things (at least prompting
you to introduce FIXMEs) and then fixing them up later in the same series.

I wouldn't say things are broken, they are just inconsistent.

> On the other hand, I'm not sure why we are storing the isl::ctx in schedule
trees in the first place. Other parts of the code are apparently simply
using the global ctx.

There are very few parts that use the global ctx: compiler API, mapping id constructors and some tests.  Pretty much everywhere else we extract the ctx from an available isl object, or pass it from somewhere.  At this point, I'd say we keep doing that for the sake of consistency.

> I don't know what the motivation was for introducing
this global ctx

AFAIR, in the old "executor" API based on C2 operators, there was a problem with copying objects between threads.  Anyway, a `thread_local` ctx is a reasonable thing to have until there is support for moving objects between threads/ctxs.

tc/core/polyhedral/schedule_isl_conversion.cc
ftynse: Except that we use `make*` for static member functions in this codebase.
`makeElem` already exists, but ok, I'll rename this one.

skimo-openhub:
On Wed, Jul 04, 2018 at 01:37:01AM -0700, ftynse wrote:
> > I don't know what the motivation was for introducing
> this global ctx
> 
> AFAIR, in the old "executor" API based on C2 operators, there was a problem with copying objects between threads.

How is that solved by introducing a thread-local "global" ctx?

> Anyway, a `thread_local` ctx is a reasonable thing to have until there is support for moving objects between threads/ctxs.

I'll be glad to add them if there is any need for them.

skimo

tc/core/polyhedral/schedule_isl_conversion.cc
ftynse: After looking in more detail, I think I keep the "from*" notation.  These are functions _local_ to a file, and almost every other function is that file is called "fromIsl*" (except for a handful of "toIsl*" that explains the choice of "from"/"to").

tc/core/polyhedral/schedule_tree.cc
ftynse: That can be fixed by swapping the order of commits

ftynse:
> How is that solved by introducing a thread-local "global" ctx?

I don't think that particular change did, but it helped uncover the actual problem.

ftynse:
Jenkins was passing before the force-push, which only swapped the order of commits.  Merging.

PR #547, opened 2018-07-02
Adding functions in Python binder

Artix18:
- max_shared_memory
- profile (returns execution time)
- profile_stats (returns mean execution time over a sample)

### DIFF
--- a/tensor_comprehensions/pybinds/tclib.cc
+++ b/tensor_comprehensions/pybinds/tclib.cc
@@ -30,6 +30,7 @@
 #include "tc/aten/aten_autotuner.h"
 #include "tc/autotuner/genetic_search.h"
 #include "tc/autotuner/options_cache.h"
+#include "tc/core/cuda/cuda.h"
 #include "tc/core/cuda/cuda_backend.h"
 #include "tc/core/cuda/cuda_tc_executor.h"
 #include "tc/core/flags.h"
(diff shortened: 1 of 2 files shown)


tensor_comprehensions/pybinds/tclib.cc
```diff
@@ -447,6 +473,11 @@ PYBIND11_MODULE(tclib, m) {
     return res;
   });
 
+  // Get GPU maximum shared memory
+  m.def("max_shared_memory", []() {
```
nicolasvasilache: This is confusing because `max_shared_memory` is a mapping options and here you want to expose properties of the GPUs. I would call it `shared_memory_size`.
Also please split the addition of this function in a separate commit.

tensor_comprehensions/pybinds/tclib.cc
```diff
@@ -459,7 +490,19 @@ PYBIND11_MODULE(tclib, m) {
           "unchecked_run",
           &TcExecutor::uncheckedRun,
           py::arg("inputs"),
-          py::arg("outputs") = py::tuple());
+          py::arg("outputs") = py::tuple())
+      .def(
+          "profile",
```
nicolasvasilache: I would call this `profile_kernel` to avoid surprising behavior in userland.

tensor_comprehensions/pybinds/tclib.cc
```diff
@@ -459,7 +490,19 @@ PYBIND11_MODULE(tclib, m) {
...
+          &TcExecutor::profile,
+          py::arg("inputs"),
+          py::arg("outputs") = py::tuple())
+      .def(
```
nicolasvasilache: I don't think increasing the API complexity in this case is worth it: in userland one can just call profile repeatedly and build one's own p0, p25, p50, p90

PR #549, opened 2018-07-02
Adding functions in Python binder

Artix18:
Adding two functions:
- tc.shared_memory_size()
- TcExecutor.profile_kernel(input, (output))

### DIFF
--- a/tensor_comprehensions/__init__.py
+++ b/tensor_comprehensions/__init__.py
@@ -35,6 +35,7 @@ from tensor_comprehensions.tclib import MappingOptionsCache
 from tensor_comprehensions.tclib import TcExecutor
 from tensor_comprehensions.tclib import Tuner
 from tensor_comprehensions.tclib import TunerConfig
+from tensor_comprehensions.tclib import shared_memory_size
 
 import tensor_comprehensions.tclib as tclib
 
@@ -608,6 +609,7 @@ __all__ = [
     'compile',
     'autotune',
     'autotune_and_compile',
+    'shared_memory_size',
     # Classes exposed by the tclib
     'CompilationCache',
     'MappingOptions',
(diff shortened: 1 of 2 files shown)


tensor_comprehensions/__init__.py
```diff
@@ -35,6 +35,7 @@
 from tensor_comprehensions.tclib import TcExecutor
 from tensor_comprehensions.tclib import Tuner
 from tensor_comprehensions.tclib import TunerConfig
+from tensor_comprehensions.tclib import shared_memory_size
```
ftynse: This is a quite low-level function, I would keep it in `tclib` and not import by default with `import tensor_comprehensions`.  We really want to keep the high-level API as simple as possible

tensor_comprehensions/pybinds/tclib.cc
```diff
@@ -273,6 +273,23 @@ struct TcExecutor {
...
+
+  size_t profile(const py::tuple& inputs, const py::tuple& outputs) {
+    if (outputs.size() > 0) {
+      auto atOutputs = getATenTensors(outputs);
+      auto atInputs = getATenTensors(inputs);
+      tc::ProfilingInfo profinfo =
+          tc::aten::profile(*executor, atInputs, atOutputs);
+      return profinfo.kernelRuntime.toMicroSeconds();
+    } else {
+      auto atInputs = getATenTensors(inputs);
```
ftynse: Most of these lines are duplicates of the "if" block and can be hoisted.  You should be able to just use a ternary operator when assigning `atOutputs` and remove the branch completely.

tensor_comprehensions/pybinds/tclib.cc
```diff
@@ -465,7 +482,13 @@ PYBIND11_MODULE(tclib, m) {
           "unchecked_run",
           &TcExecutor::uncheckedRun,
           py::arg("inputs"),
+          py::arg("outputs") = py::tuple())
+      .def(
+          "profile_kernel",
```
ftynse: Let's keep the names consistent with C++ API.  It should be called `profile`.

nicolasvasilache: profile returns a ProfileInfo structure that we don't want to expose, `profile_kernel` returns the kernel time in us.

PR #550, opened 2018-07-03
tightenLaunchBounds: use mapping schedule instead of mapping filters

### LINKED ISSUE #515: tightenLaunchBounds: extract bounds from mapping
The bounds are currently extracted from a context space.
This doesn't give the wrong results, but it would be more natural
to extract them from the actual block/thread mapping.

skimo-openhub:
The mapping filters are derived from the mapping schedule.
It is more natural, simpler and less error-prone
to derive the tightened launch bounds directly from
the mapping schedule.

Closes #515

### DIFF
--- a/tc/core/polyhedral/cuda/mapped_scop.cc
+++ b/tc/core/polyhedral/cuda/mapped_scop.cc
@@ -48,6 +48,26 @@ namespace polyhedral {
 
 namespace {
 
+/*
+ * Check if the minimum of any affine function in "list" is strictly
+ * positive on "domain" and, if so, print a warning.
+ * The functions in "list" are used for mapping to block/thread identifiers.
+ * If the minimum is strictly positive, then this means some
+ * identifiers do nothing in the corresponding subtree.
+ * Note that tightenLaunchBounds (if extended to consider
+ * the minimal value of the mapping) may not help much because
+ * those identifiers may get used in other parts of the tree.
+ * Furthermore, at that point it is too late to change the mapping.
+ */
+static void checkMinimum(isl::union_set domain, isl::union_pw_aff_list list) {
+  for (auto upa : list) {
+    upa = upa.intersect_domain(domain);
+    auto min = upa.min_val();
+    LOG_IF(WARNING, min.is_pos())
+        << "Opportunity for shifting mapping -> min:" << min;
+  }
+}
+
 template <typename ExceptionType>
 inline void throwIfHasPattern(
     ScheduleTreeMatcher matcher,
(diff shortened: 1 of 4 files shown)


tc/core/polyhedral/cuda/mapped_scop.cc
```diff
@@ -531,24 +543,25 @@ isl::multi_aff constructThreadToWarp(
 }
 } // namespace
 
-isl::multi_union_pw_aff MappedScop::threadMappingSchedule(
+template <typename MappingTypeId>
+isl::multi_union_pw_aff MappedScop::mappingSchedule(
```
nicolasvasilache: This should be a free function with static scop that takes the parameters it needs.
Exposing its declaration is useless (since it's not supposed to be used), error-prone (if used with some non-explicitly instantiated type in the future -> linking errors) and increases the API.

Alternatively it could go to a -inl.h and we could drop both `threadMappingSchedule` and `blockMappingSchedule`; whichever is fine with me.

nicolasvasilache:
looks good except the `mappingSchedule` in the API part

tc/core/polyhedral/cuda/mapped_scop.cc
skimo-openhub: On Tue, Jul 03, 2018 at 11:43:23AM -0700, Nicolas Vasilache wrote:
> This should be a free function with static scop that takes the parameters it needs.

What exactly do you propose?
(I don't see why `MappedScop::mappingSchedule` should have a different
prototype from `MappedScop::threadMappingSchedule`.)

> Exposing its declaration is useless (since it's not supposed to be used), error-prone (if u
sed with some non-explicitly instantiated type in the future -> linking errors) and increases
 the API.

The declaration is not useless.  It is used in the last commit.
As explained in the commit message, there are only two possible instantiations.

> Alternatively it could go to a -inl.h and we could drop both `threadMappingSchedule` and `b
lockMappingSchedule`; whichever is fine with me.

I don't see the point.  I'll just move it into `tighten_launch_bounds.cc`.

skimo

skimo-openhub:
@caffe2bot retest this please

tc/core/polyhedral/cuda/mapped_scop.cc
ftynse: > if used with some non-explicitly instantiated type in the future -> linking errors

`static_assert` to the rescue!

I think @nicolasvasilache looks for something like `collectMappingsTo` that we have in memory promotion heuristic code https://github.com/facebookresearch/TensorComprehensions/blob/7f4ed011b9df9695cf7b8b366d66d90ab2685185/tc/core/polyhedral/cuda/memory_promotion_heuristic.cc#L303

`threadMappingSchedule` comes from my commit, I believe.  The idea was to have `mappingSchedule<mapping::Block>()` and `mappingFilter<mapping::Block>()` together at some point, they only differ by the return type.

skimo-openhub:
@caffe2bot retest this please

PR #551, opened 2018-07-03
More pybind functions

Artix18:
Add two functions to the python binder: 

tclib.shared_memory_size(), TcExecutor.profile_kernel(input, (outputs))
- first one for accessing the size of GPU shared memory (useful for one option setting)
- second one for getting the same execution time as in C++

### DIFF
--- a/tensor_comprehensions/pybinds/tclib.cc
+++ b/tensor_comprehensions/pybinds/tclib.cc
@@ -273,6 +273,17 @@ struct TcExecutor {
       return tupleOrTensor(convertToPyObjects(atOutputs));
     }
   }
+
+  size_t profile(const py::tuple& inputs, const py::tuple& outputs) {
+    auto atInputs = getATenTensors(inputs);
+    auto atOutputs = (outputs.size() > 0)
+        ? getATenTensors(outputs)
+        : tc::aten::prepareOutputs(tc, entryPoint, atInputs);
+    tc::ProfilingInfo profinfo =
+        tc::aten::profile(*executor, atInputs, atOutputs);
+    return profinfo.kernelRuntime.toMicroSeconds();
+  }
+
   std::string tc;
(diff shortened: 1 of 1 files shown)


tensor_comprehensions/pybinds/tclib.cc
```diff
@@ -459,7 +476,13 @@ PYBIND11_MODULE(tclib, m) {
           "unchecked_run",
           &TcExecutor::uncheckedRun,
           py::arg("inputs"),
+          py::arg("outputs") = py::tuple())
+      .def(
+          "profile_kernel",
```
ftynse: Please call this "profile" too.

Artix18: Nicolas Vasilache prefers "profile_kernel" as it doesnt bind the whole ProfileInfo structure.

ftynse: > as it doesnt bind the whole ProfileInfo structure

why not (mention that in the commit message)?  

Anyway, there are two points:
- the C++ function should have the same name as the python function, otherwise we will quickly loose grip of what corresponds to what and risk creating name conflicts;
- I strongly prefer python API to be as close to C++ as possible;  we went the other way, you can ask @nicolasvasilache how much time did he spend reworking that to remove overhead and inconsistencies.

Ok, the function only returns the kernel time.  It is sort of indicated in the name, but not immediately clear (one can interpret it as "profile one kernel" as opposed to "profile the entire network" or even "profile the compilation and the kernel").  I don't see another name for now, but this is something that should be mentioned in the commit message.

ftynse:
> (And for the comment above, as I am working on finding the best set of options, i need to be able to set an option to all its different good possible values)

What are these good possible values?  Are they anyhow related to the provided shared memory size?  If so, how exactly?

PR #557, opened 2018-07-09
do not rely on CMakeLists.txt in isl submodule

skimo-openhub:
Mainline isl uses autoconf/automake as a build system.
The isl submodule used by TC is a heavily modified version of isl and
also supports cmake.
Reducing the difference and keeping up with changes in mainline isl
occasionally requires changes to the build system configuration files.
Having to perform the changes for both systems at the same time
while merging is a needless additional complication.
Support for cmake will therefore be removed from the modified version and
TC can no longer rely on this support.

There is at least two ways of handling the removal of cmake support
from the isl submodule.  The isl submodule could be configured
as a cmake "external project" or the isl files can simply be
compiled in directly.  This commits opts for the second option
based on the assumption that an external project will be more
difficult to handle on Windows systems.
Note that this commit also hard-codes some options that will
not work on Windows, but setting these options correctly
should not be too difficult for someone familiar with cmake.

Note that if any source files get added to isl, then
the local CMakeLists.txt still needs to get updated, but this
can be done while updating the submodule and after the merge
has been completed.


skimo-openhub:
@caffe2bot retest this please

skimo-openhub:
On Wed, Jul 11, 2018 at 12:32:35AM -0700, ftynse wrote:
> 
> We already have `tc/external/isl`, why put these files in `tc/core`? 

The files in tc/external/ (there is no tc/external/isl/) are about
the interface between isl and TC, while the newly added files are
about compiling isl itself.  I don't think we should mix the two.
As to why I put them in tc/core/, I thought that made sense
because the isl code is part of the tc_core library, but I'm open
to suggestions.

skimo

ftynse:
> The files in tc/external/ (there is no tc/external/isl/)

Then my proposition is to create it.

>  are about the interface between isl and TC, while the newly added files are about compiling isl itself. I don't think we should mix the two. As to why I put them in tc/core/, I thought that made sense because the isl code is part of the tc_core library, but I'm open to suggestions.

I'm not sure isl _code_ is part of `tc_core`.  The isl _library_ is used in `tc_core`, but I think we should maintain clean separation between the two.

skimo-openhub:
On Wed, Jul 11, 2018 at 02:20:30AM -0700, ftynse wrote:
> > The files in tc/external/ (there is no tc/external/isl/)
> 
> Then my proposition is to create it.

As I explained below, I think it would be a bad idea to mix the two.
In particular, TC code should include "tc/external/isl.h",
but it should then *NOT* include any "tc/external/isl/*.h".
I think that's too confusing and error-prone.

> >  are about the interface between isl and TC, while the newly added files are about compiling isl itself. I don't think we should mix the two.
> As to why I put them in tc/core/, I thought that made sense because the isl code is part of the tc_core library, but I'm open to suggestions.
> 
> I'm not sure isl _code_ is part of `tc_core`.  The isl _library_ is used in `tc_core`, but I think we should maintain clean separation between the two.

Well, in my view, if we start compiling isl ourselves, it becomes
part of tc_core.  But anyway, I can put it somewhere else.
I just don't think it's a good idea to put it right next
to the interface.

skimo

ftynse:
> Well, in my view, if we start compiling isl ourselves, it becomes part of tc_core.

It's still an external library, with code elsewhere and different ownership/commit procedure/licensing.  Putting build scripts for it (but not the code) in `tc/core` is also confusing.  Polly has a similar issue, but it has a local copy of isl's _code_, and puts isl in `lib/External/isl`.  I'd go for something like that.  Maybe we should not have the interface under `external/isl` in the first place.  Interface is _not_ a part of isl atm.

skimo-openhub:
On Fri, Jul 13, 2018 at 04:42:22AM -0700, ftynse wrote:
> > Well, in my view, if we start compiling isl ourselves, it becomes part of tc_core.
> 
> It's still an external library, with code elsewhere and different ownership/commit procedure/licensing.  Putting build scripts for it (but not the code) in `tc/core` is also confusing.  Polly has a similar issue, but it has a local copy of isl's _code_, and puts isl in `lib/External/isl`.  I'd go for something like that.

How about I put it in "external" (_not_ "tc/external") for now?

> Maybe we should not have the interface under `external/isl` in the first place.  Interface is _not_ a part of isl atm.

I don't know why it was put there.
It seems to contain stuff that is not going to be included
in isl any time soon.

skimo

ftynse:
>> It's still an external library, with code elsewhere and different ownership/commit procedure/licensing.  Putting build scripts for it (but not the code) in `tc/core` is also confusing.  Polly has a similar issue, but it has a local copy of isl's _code_, and puts isl in `lib/External/isl`.  I'd go for something like that.

> How about I put it in "external" (_not_ "tc/external") for now?

Sounds good!

> I don't know why it was put there. It seems to contain stuff that is not going to be included in isl any time soon.

To remove it from `tc/core`...  I guess we should rename `tc/external/isl.h` to something like `tc/util/isl_util.h`

PR #558, opened 2018-07-09
ScheduleTree*: hide constructors

ftynse:
To enforce the memory model, schedule tree nodes should not be constructed on stack but only wrapped in unique pointers.  Make the (non-copy) constructors of the ScheduleTree* classes private and provide ScheduleTree*::make functions that return unique pointers.

@skimo-openhub @nicolasvasilache do you prefer separate commits for each class or one commit with all classes?  Changes are _mostly_ identical.

### DIFF
--- a/tc/core/polyhedral/schedule_tree_elem.h
+++ b/tc/core/polyhedral/schedule_tree_elem.h
@@ -297,11 +297,15 @@ struct ScheduleTreeBand : public ScheduleTree {
  * underneath the innermost band member mapped to threads.
  */
 struct ScheduleTreeThreadSpecificMarker : public ScheduleTree {
+ public:
   static constexpr detail::ScheduleTreeType NodeType =
       detail::ScheduleTreeType::ThreadSpecificMarker;
 
+ private:
   explicit ScheduleTreeThreadSpecificMarker(isl::ctx ctx)
       : ScheduleTree(ctx, {}, NodeType) {}
+
+ public:
   virtual ~ScheduleTreeThreadSpecificMarker() override {}
 
   bool operator==(const ScheduleTreeThreadSpecificMarker& other) const {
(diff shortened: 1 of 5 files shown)


skimo-openhub:
On Mon, Jul 09, 2018 at 06:57:52AM -0700, ftynse wrote:
> @skimo-openhub @nicolasvasilache do you prefer separate commits for each class or one commit with all classes?  Changes are _mostly_ identical.

I prefer separate commits, especially if you already have them.

skimo

ftynse:

tc/core/polyhedral/schedule_tree_elem.h
```diff
@@ -31,16 +31,24 @@ namespace polyhedral {
 namespace detail {
 
 struct ScheduleTreeContext : public ScheduleTree {
+ public:
```
skimo-openhub: Hmm... seems to belong to the "reformat", but ok.

tc/core/polyhedral/schedule_tree_elem.h
```diff
@@ -31,16 +31,24 @@ namespace polyhedral {
...
+ private:
   ScheduleTreeContext() = delete;
-  ScheduleTreeContext(const ScheduleTreeContext& eb)
-      : ScheduleTree(eb), context_(eb.context_) {}
   explicit ScheduleTreeContext(isl::set s)
       : ScheduleTree(s.get_ctx(), {}, NodeType), context_(s) {}
+
+ public:
+  ScheduleTreeContext(const ScheduleTreeContext& eb)
+      : ScheduleTree(eb), context_(eb.context_) {}
```
skimo-openhub: If an outside user is not supposed to create objects directly,
then why is the copy constructor still exposed?

Issue #562, opened 2018-07-10
Autotuner invalid value for tiling


Artix18:
Built from source (master), commit b77b5835ca1165f63434e09ca915eea81df07726

input sizes: N, C, H, W, O, kH, kW = 32, 4, 56, 56, 16, 1, 1
Therefore, if i am not mistaken, the union of "divisors" (and {0}) should be:
[1, 2, 3, 4, 5, 6, 7, 8, 10, 11, 12, 14, 16, 19, 28, 56, 32, 64, 128, 256, 0]

```python
import tensor_comprehensions as tc
import torch

tc_name = "convolution"
tc_code = """
  def convolution(float(N,C,H,W) I, float(M,C,KH,KW) W1) -> (O) {
  O(n, m, h, w) +=! I(n, r_c, h + r_kh, w + r_kw) * W1(m, r_c, r_kh, r_kw)
  }
"""

N, C, H, W, O, kH, kW = 32, 4, 56, 56, 16, 1, 1
I, W1 = torch.randn(N, C, H, W, device='cuda'), torch.randn(O, C, kH, kW, device='cuda')
inp = (I, W1)

opts = tc.MappingOptions("naive")
opts.tile([5])

tc.autotune(tc_code, tc_name, *inp, starting_options=opts)
```

```
Traceback (most recent call last):
  File "bug.py", line 18, in <module>
    tc.autotune(tc_code, tc_name, *inp, starting_options=opts)
  File "/private/home/artix18/anaconda3/envs/tc_build/lib/python3.6/site-packages/tensor_comprehensions/__init__.py", line 214, in autotune
    tuner_config)
ValueError: RangeParameter::selectFromValue: value 5 not in range. The valid values are: 1 2 4 7 8 14 16 28 32 56 64 128 256 0
```

In my opinion, 5=ceil(32/7) and should then be a valid value.
I didn't have this problem previously.

nicolasvasilache:
Hmm, see [here](https://github.com/facebookresearch/TensorComprehensions/blob/master/tc/autotuner/utils.cc#L44).
We actually set the ceil by powers2 only, you can use that instead of all ceils.
But the impl seems bugged (cc @ttheodor) because we insert in place and reuse iteratively.

@Artix18 this is a good opportunity to commit an easy code fix to C++-land :)
Keeping 'ceil by powers2 only' behavior sounds reasonable to me to cut the size of the search space.

ftynse:
[...]

PR #564, opened 2018-07-11
[autotuning]User defined tile,block, grid sizes

ttheodor:
Previously if the user inlcuded a tile, block, or grid size that
was not in the set of values generated by the autotuner's heuristic,
it would result in an error.

### DIFF
--- a/tc/autotuner/autotuner-inl.h
+++ b/tc/autotuner/autotuner-inl.h
@@ -356,10 +356,12 @@ size_t largestDim(const std::vector<const DLConstTensor*>& inputs) {
   return (*maxElement)->ndim;
 }
 
-// Creates well-chosen parameter sizes to match the input shapes.
-void setupTuningParameters(
+// Creates well-chosen generic parameter sizes to match the input shapes.
+template <typename MappingOptionsType>
+inline std::pair<TuningConfiguration, std::vector<size_t>>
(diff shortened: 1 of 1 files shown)


ttheodor:
@Artix18 this should fix #562

tc/autotuner/autotuner-inl.h
```diff
@@ -357,24 +357,62 @@ size_t largestDim(const std::vector<const DLConstTensor*>& inputs) {
...
+  configuration.blockParams.setRange(blockRange, "b");
+  configuration.gridParams.setRange(gridRange, "g");
   configuration.unrollFactor =
       RangeParameter(powers2(FLAGS_tuner_max_unroll_size), "unroll");
   configuration.privateDepth =
       RangeParameter({0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10}, "pdepth");
 }
+
+// Creates well-chosen parameter sizes to match the input shapes.
+template <>
```
ftynse: Why did you chose to have the CPU version as template specialization and GPU version as default?  I would suggest having both as specializations and leaving the default template without body (link error if somebody tries it), or even doing some `enable_if` magic to get a compilation error.

You can also factor out the tiling setup to a separate function, called from both the CPU and GPU functions.

ttheodor:
@ftynse I addressed your comment.

PR #571, opened 2018-07-16
Add ReLU + masked convolution

lvdmaaten:
At training time, CondenseNets use a convolution in which certain input-output connections (that is, the entire kernel) are masked out. This example implements a combination of that masked convolution with a ReLU activations.

The goal is to further work out this example into increasingly large chunks of the CondenseNet.

Tested this internally: tests pass.

### DIFF
--- a/python/examples/masked_conv.py
+++ b/python/examples/masked_conv.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2017-present, Facebook, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


python/examples/masked_conv.py
```diff
@@ -0,0 +1,168 @@
...
+tuner_config = (
+    tc.TunerConfig()
+    .threads(args.tuner_threads)
+    .generations(args.tuner_generations)
+    .pop_size(args.tuner_pop_size)
+    .number_elites(args.tuner_number_elites)
+    .devices(args.tuner_devices))
+
+# This function is used for reinforcing tuning
+# 1. make_idx is small and does not get tuned or saved, just using naive
```
nicolasvasilache: These comments are actually from tum.py

python/examples/masked_conv.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+###############################################################################
+# Check
+###############################################################################
+tc.assert_almost_equal(
+    output.cpu(),
+    Output.cpu(),
+    input_data.cpu(), conv.weight.cpu(), mask.cpu(),
+    precision=1e-5)
```
nicolasvasilache: So this is going to be `1e-5` relative precision.
Maybe it is better to write:
```
tc.assert_almost_equal(
    output.cpu(),
    Output.cpu(),
    input_data.cpu(), conv.weight.cpu(), mask.cpu(),
    operations= C * K * K,
    precision=1e-7)
```

PR #572, opened 2018-07-16
[WIP] Exp/rnn

nicolasvasilache:
WIP cc @apaszke


nicolasvasilache:
Had a nice chat with @abadams and he'll be taking the baton on the Halide -> ISL portion as pieces of the code can be significantly simplified.

PR #578, opened 2018-07-24
prepare for templated isl types

skimo-openhub:
A bunch of typing issues discovered while working on templated isl types.


skimo-openhub:
@caffe2bot retest this please

PR #579, opened 2018-07-24
Compiler options

### LINKED ISSUE #456: Add a flag to supress warnings

ftynse:
This class intends to collect TC flow options in a single object that
    can be passed around, instead of (re)setting flags.  The latter are
    essentially global variables that may or may not be initialized from
    command-line arguments.  Start populating the CompilerOptions class with
    the new "emitWarnings" option that would otherwise become a flag.  In
    future, the use of flags must be restricted to files that produce
    binaries, e.g., those containing the main() function and having explicit
    access to command-line arguments.

    CompilerOptions is located in the new "tc/utils" directory as it needs
    to be accessed by both tc_lang and tc_core libraries.  Because it's
    header-only, there is not need to build it into a separate library and
    link.  In future, other TC-generic headers, for example tc/core/check.h,
    should move to tc/utils as well.

Closes #541 
Closes #456

### DIFF
--- a/tc/aten/aten_compiler-inl.h
+++ b/tc/aten/aten_compiler-inl.h
@@ -32,10 +32,11 @@ std::unique_ptr<typename Backend::ExecutorType> compile(
     const std::string& tc,
     const std::string& entryPoint,
     const std::vector<at::Tensor>& inputs,
-    const typename Backend::MappingOptionsType& options) {
+    const typename Backend::MappingOptionsType& options,
+    const CompilerOptions& compilerOptions) {
   auto inputDLTensors = makeDLConstTensors(inputs);
   return tc::compile<Backend>(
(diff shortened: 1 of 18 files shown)


nicolasvasilache:
beb49a81269abf1c12f82c88cf757b80ba875225 why go the explicit `CompilerOptions` route, default options don't seem bad to me here.

nicolasvasilache:
rest looks good

ftynse:
> beb49a8 why go the explicit CompilerOptions route, default options don't seem bad to me here.

It was easier to track all relevant places that must pass the options.  Updated.

Issue #553, opened 2018-07-04
ScheduleTree API evolution


ftynse:
#543 removed the notion of schedule tree element: now specific, node types inherit directly from `ScheduleTree`, which is simpler and offers more type safety.  This refactoring uncovered several technical and conceptual issues that may be worth discussing.

- Using macros for dynamic dispatch.
I don't think using macros for type-based dispatch is justified in C++.  Virtual functions are a native mechanism to achieve exactly that behavior.  Worth, _some_ functions already rely on dynamic polymorphism through virtual functions while others do macro-based dynamic dispatch.  We can also consider true CRTP for static dispatch, but it looks like premature optimization: virtual function calls are cheap.
- The `type_` field.
If we get rid of macro-based dispatch, the `type_` field is not strictly necessary.  One can use `dynamic_cast` to check if a schedule node has in fact a specific node type.  Alternatively, the field may be replaced with a virtual function overridden by derived classes to return a constant (note that #543 killed such a function, but it was duplicate data and used in much less places than the `type_` field).  With CRTP, the subclass type is readily available in the parent class.
- Separate header containing all specific node types.
Not sure if it justified to keep classes specific node types in a separate file (`schedule_tree_elem.{h,cc}`).  Most tree utils and transformations require specific node classes anyway, for example to extract the domain space from the domain node.  The only functions that may not require the definitions of specific node types are generic traversals, which are static members of `ScheduleTree` anyway.
- Construction management
Specific node classes have public constructors, but the memory management scheme for `ScheduleTree` says that one should manipulate ScheduleTrees through (unique) pointers only.  `ScheduleTree` has a set of static member functions `make*` to construct specific nodes wrapped in a unique pointer.  Given the new inheritance scheme, one can move these static constructor functions to the respective classes, removing the need for public constructors.  Alternatively, derived classes may declare `ScheduleTree` as friend, but the constructors should be hidden anyway.

tag @nicolasvasilache @skimo-openhub for discussion

ftynse:
My preferences for these are:
- use virtual methods, kill macro-dispatch with fire
- use a virtual `type()` method
- merge `schedule_tree.h` and `schedule_tree_elem.h`
- move construction to specific classes, add a virtual `clone()` method to properly dispatch copy constuction

nicolasvasilache:
[...]

PR #583, opened 2018-07-25
ScheduleTree: replace comparison operators with named functions

ftynse:
Proceeding with ScheduleTree evolution plan (#553)

This PR replaces overloaded comparison operators with two functions, `nodeEquals` and `treeEquals`.  The main motivation for this change: after grafting ScheduleTreeElem* onto ScheduleTree in the inheritance structure we ended up with `operator==` comparing **subtrees** if called on ScheduleTree and **nodes** if called on instances of other classes.  I argue that the behavior of default comparison operator is _unintutitive_ for tree-like structure.  Therefore, introducing explicitly-named comparison functions that make the intention clear is an improvement.

As an additional benefit, regular member functions can be declared virtual, unlike operators.  With this approach, each subclass needs to override _one_ abstract function that does node-to-node comparison while subtree comparison is handed automatically.  Previously, it would have been necessary to also modify the clunky macro-based logic in one of the comparison operators.


nicolasvasilache (approved):

PR #584, opened 2018-07-25
Fixing namespacing issues

nicolasvasilache:
@ftynse reported segfaults due to improper namespacing.
Isolate the cpu MappedScop and relevant cuda classes in their own namespace.


ftynse:
ci mitigation signoff

PR #599, opened 2018-08-09
 bump isl for merge of C++ bindings

skimo-openhub:
~~Depends on https://github.com/nicolasvasilache/isl/pull/176~~


tc/core/polyhedral/scop.h
```diff
@@ -71,7 +71,9 @@ struct Scop {
     res->parameterValues = scop.parameterValues;
     res->halide = scop.halide;
     res->body = scop.body;
-    res->dependences = scop.dependences;
+    if (scop.dependences) {
```
ftynse: I'm not sure this is a good idea, see discussion in isl maiilng

